PR #4, opened 2019-03-29
add development & CI test profiles

andreeaflorescu:
Added two test profiles: one for development (devel) and one for
running the integration tests as part of the CI.

Currently the only difference between the two is running the
coverage test. Specifically, in the `ci` profile the test fails
if current_coverage > previous_coverage, while in the `devel`
profile the coverage file is updated with the new value.

Signed-off-by: Andreea Florescu <fandree@amazon.com>


tests/test_coverage.py
```diff
@@ -82,15 +83,25 @@ def _update_coverage(cov_value):
...
+            assert False, "Coverage is increased from {} to {}. " \
+                          "Please update the coverage in " \
+                          "tests/coverage.".format(
+                previous_coverage,
+                current_coverage
+            )
+        elif profile == pytest.profile_devel:
+            _update_coverage(current_coverage)
+        else:
+            # This should never happen because we pytest should only accept
```
alxiord:
```suggestion
            # This should never happen because pytest should only accept
```

tests/test_coverage.py
```diff
@@ -82,15 +83,25 @@ def _update_coverage(cov_value):
     with open(coverage_path, "w") as f:
         f.write(str(cov_value))
```
alxiord, line 85: nit: can you add 2 blank lines between all top level functions while you're at it?

README.md
```diff
@@ -32,10 +33,23 @@ tests/test_coverage.py::test_coverage PASSED                             [100%]
...
 ```
 
+### Test Profiles
+
+The integration tests support two test profiles:
+- **devel**: this is the recommended profile for running the integration tests on
+  a local development machine.
+- **ci** (default option): this is the profile used when running the integration tests as part of
+  the the Continuous Integration (CI).
+
```
sameo, line 31: Could we describe the main differences between the 2 profiles?

README.md
```diff
@@ -5,6 +5,7 @@ TODO
 ## Running the tests
 
 ### Dependencies
+
```
sameo: Nit: Unrelated cleanup.

README.md
andreeaflorescu: For now the only difference is in running the coverage test. This difference is described in the Adaptive Coverage section of the doc. For this particular crate I don't believe we will need anything else, but for the virtio-devices we might want to do performance testing and we probably cannot expect to have the same thresholds on a local machine vs on a i3.metal. We can use the same model for these crates in the future.

I have no good idea about describing the difference between the profiles. Should I just say that the coverage test behaves differently with different profiles? Do you have any better suggestions?

tests/test_coverage.py
andreeaflorescu: Done.

tests/test_coverage.py
andreeaflorescu: I'll add it in a future cleanup of the code.

README.md
andreeaflorescu: Done.

andreeaflorescu:
@aghecenco @sameo please re-review.

alxiord (approved):

sameo (approved):

PR #5, opened 2019-03-29
Correctly label x86/x86_64 specific code

dianpopa:
Test it with `cargo build --target aarch64-unknown-linux-musl`

### DIFF
--- a/src/ioctls/system.rs
+++ b/src/ioctls/system.rs
@@ -9,7 +9,9 @@ use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
 use cap::Cap;
 use ioctls::vec_with_array_field;
 use ioctls::vm::{new_vmfd, VmFd};
-use ioctls::{CpuId, Result};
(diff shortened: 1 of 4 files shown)


src/ioctls/system.rs
```diff
@@ -9,6 +9,8 @@ use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
 use cap::Cap;
 use ioctls::vec_with_array_field;
 use ioctls::vm::{new_vmfd, VmFd};
+use ioctls::Result;
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
 use ioctls::{CpuId, Result};
```
andreeaflorescu: I think this line needs to be changed to `use ioctls::CpuId;`

dianpopa: Yeah, that was a miss. Corrected it.

andreeaflorescu (approved):

sameo (approved):
LGTM

PR #6, opened 2019-03-29
implement KvmVec struct

serban300:
KvmVec provides a common abstraction for all the KVM structures that resemble arrays. For example: CpuId and MsrList


src/ioctls/kvm_vec.rs
```diff
@@ -0,0 +1,394 @@
...
+//    len: u32,
+//    some_data: u32,
+//    entries: __IncompleteArrayField<F>,
+// }
+// ```
+#[allow(clippy::len_without_is_empty)]
+pub trait KvmArray<F> {
+    /// Get the array length
+    ///
+    fn len(&self) -> u32;
```
acatangiu: Standard type used in such cases is `usize`. Have a look and see if it makes sense to switch.

src/ioctls/kvm_vec.rs
```diff
@@ -0,0 +1,394 @@
...
+
+    /// Get the array length as mut
+    ///
+    fn len_mut(&mut self) -> &mut u32;
```
acatangiu: I suggest using a setter method `set_len(len: usize/u32)` instead or returning a mutable reference.

src/ioctls/kvm_vec.rs
```diff
@@ -0,0 +1,394 @@
...
+// ```
+// #[repr(C)]
+// struct T {
+//    len: u32,
+//    some_data: u32,
+//    entries: __IncompleteArrayField<F>,
+// }
+// ```
+#[allow(clippy::len_without_is_empty)]
+pub trait KvmArray<F> {
```
acatangiu: You can remove this generic type parameter `F` like so:
```
pub trait KvmArray {
    /// Type of kvm struct used as entry
    type Entry;
    
    // the rest is just as before, except methods that were using <F>

    fn entries(&self) -> &__IncompleteArrayField<Self::Entry>;
}
```

While it doesn't make much of difference here, it will help in the definition of `struct KvmVec` (see next comments).

src/ioctls/kvm_vec.rs
```diff
@@ -0,0 +1,394 @@
...
+    fn entries(&self) -> &__IncompleteArrayField<F>;
+
+    /// Get the array entries as mut
+    ///
+    fn entries_mut(&mut self) -> &mut __IncompleteArrayField<F>;
+}
+
+/// An adapter that helps in treating a KvmArray more like an actual `Vec`.
+///
+pub struct KvmVec<T: Default + KvmArray<F>, F> {
```
acatangiu: You can get rid of type parameter `F` by using the `type` option I described above when defining `KvmArray`.

This would also get rid of the `PhantomData<F>`.

In the couple of places where `F` is used in `impl KvmVec { }`, you can swap it with `<T as KvmArray>::Entry` (where `Entry` is whatever type name you define in `KvmArray`).

I believe this will make using this structure easier/more intuitive.

src/ioctls/kvm_vec.rs
```diff
@@ -0,0 +1,394 @@
...
+    }
+
+    /// Returns a mut `Vec<F>` that contains all the elements.
+    /// It is important to call `mem::forget` after using this vector. Otherwise rust will destroy it.
+    ///
+    unsafe fn as_vec(&mut self) -> Vec<F> {
+        let entries_ptr = self.as_mut_original_struct().entries_mut().as_mut_ptr();
+        Vec::from_raw_parts(
+            entries_ptr,
+            self.as_mut_original_struct().len() as usize,
```
acatangiu: nit: no need for `mut` here
```suggestion
            self.as_original_struct().len() as usize,
```

src/ioctls/kvm_vec.rs
```diff
@@ -0,0 +1,394 @@
...
+    ///
+    pub fn as_mut_ptr(&mut self) -> *mut T {
+        self.as_mut_original_struct()
+    }
+
+    /// Returns a mut `Vec<F>` that contains all the elements.
+    /// It is important to call `mem::forget` after using this vector. Otherwise rust will destroy it.
+    ///
+    unsafe fn as_vec(&mut self) -> Vec<F> {
+        let entries_ptr = self.as_mut_original_struct().entries_mut().as_mut_ptr();
```
acatangiu: nit: any reason why you're always using the accessor methods instead of just working with the struct members directly?
```suggestion
        let entries_ptr = self.kvm_array[0].entries_mut().as_mut_ptr();
```

src/ioctls/kvm_vec.rs
```diff
@@ -0,0 +1,394 @@
...
+        let current_kvm_array_len = self.kvm_array.len();
+        let required_kvm_array_len =
+            KvmVec::<T, F>::num_elements_to_kvm_array_len(self.capacity + additional);
+        let additional_kvm_array_len = required_kvm_array_len - current_kvm_array_len;
+
+        self.capacity += additional;
+
+        self.kvm_array.reserve(additional_kvm_array_len);
+        for _ in current_kvm_array_len..required_kvm_array_len {
+            self.kvm_array.push(T::default())
```
acatangiu: Are these pushes necessary?

src/ioctls/kvm_vec.rs
```diff
@@ -0,0 +1,394 @@
...
+    pub fn retain<P>(&mut self, f: P)
+    where
+        P: FnMut(&F) -> bool,
+    {
+        let mut entries = unsafe { self.as_vec() };
+        entries.retain(f);
+        *self.as_mut_original_struct().len_mut() = entries.len() as u32;
+        self.capacity = entries.len();
+
+        self.kvm_array.shrink_to_fit();
```
acatangiu: I don't think this has any effect since the number of `T` elements in `kvm_array` has not changed (even if you've changed the underlying memory re-interpreted as a `Vec<F>` when the `entries.retain(f)` was done).

src/ioctls/mod.rs
```diff
@@ -1,144 +1,86 @@
...
+mod kvm_vec;
 /// Wrappers over KVM system ioctls.
 pub mod system;
 /// Wrappers over KVM VCPU ioctls.
 pub mod vcpu;
 /// Wrappers over KVM Virtual Machine ioctls.
 pub mod vm;
+use self::kvm_vec::*;
+
+use self::kvm_vec::*;
```
acatangiu: Duplicate `use self::kvm_vec::*`

acatangiu (commented):
Noice! 👌

src/ioctls/kvm_vec.rs
serban300: Nice. Done.

src/ioctls/kvm_vec.rs
serban300: Done

src/ioctls/kvm_vec.rs
serban300: Makes sense. Done.

src/ioctls/kvm_vec.rs
serban300: Done

src/ioctls/kvm_vec.rs
serban300: I just find it more readable this way. `kvm_array[0]` looks weird. `as_mut_original_struct` gives more information on what we're actually accessing.

src/ioctls/kvm_vec.rs
serban300: Done

src/ioctls/mod.rs
serban300: Fixed

src/ioctls/kvm_vec.rs
serban300: No. Removed.

src/ioctls/kvm_vec.rs
serban300: True. The len of the kvm_array should be changed in order for this to have any effect. Fixed.

src/ioctls/kvm_vec.rs
acatangiu: ok

src/ioctls/kvm_vec.rs
```diff
@@ -0,0 +1,405 @@
...
+    /// Reserves capacity for at least `additional` more `KvmArray::Entry` elements.
+    ///
+    fn reserve(&mut self, additional: usize) {
+        let current_kvm_array_len = self.kvm_array.len();
+        let required_kvm_array_len =
+            KvmVec::<T>::num_elements_to_kvm_array_len(self.len + additional);
+        let additional_kvm_array_len = required_kvm_array_len - current_kvm_array_len;
+
+        self.kvm_array.reserve(additional_kvm_array_len);
+        self.capacity = self.len + additional;
```
acatangiu: I might be confused, but the `reserve()` is done on logic that's based on current `len`, but here you're also increasing `capacity` with the same amount.

Isn't this broken in the case where `len` is smaller than `capacity`?

src/ioctls/kvm_vec.rs
```diff
@@ -0,0 +1,405 @@
...
+        let mut entries = unsafe { self.as_vec() };
+        entries.retain(f);
+
+        self.as_mut_original_struct().set_len(entries.len());
+        self.capacity = entries.len();
+        self.len = entries.len();
+
+        let required_kvm_array_len = KvmVec::<T>::num_elements_to_kvm_array_len(self.len);
+        unsafe {
+            self.kvm_array.set_len(required_kvm_array_len);
```
acatangiu: nit: add comments for this function explaining the logic to make it easier to follow

serban300: Done

src/ioctls/kvm_vec.rs
serban300: You're right. I missed this case. Fixed.

acatangiu (dismissed):
LGTM

src/ioctls/kvm_vec.rs
```diff
@@ -0,0 +1,426 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Portions Copyright 2017 The Chromium OS Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the THIRD-PARTY file.
+
+use kvm_bindings::__IncompleteArrayField;
+use std::marker::PhantomData;
```
acatangiu: nit: no longer used

src/ioctls/kvm_vec.rs
```diff
@@ -0,0 +1,426 @@
...
+            assert!(kvm_vec_1.push(i as u32).is_ok());
+            assert!(kvm_vec_2.push(i as u32).is_ok());
+            assert!(kvm_vec_3.push(0).is_ok());
+        }
+
+        assert!(kvm_vec_1 == kvm_vec_2);
+        assert!(kvm_vec_1 != kvm_vec_3);
+    }
+
+    //    #[test]
```
acatangiu: why is this commented out?

src/ioctls/kvm_vec.rs
```diff
@@ -0,0 +1,426 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
andreeaflorescu: The ioctls module should keep the initial structure. Please checkout this comment: https://github.com/rust-vmm/kvm-ioctls/pull/12#discussion_r271179246 for details.

But since you are also adding some helpers that aren't exactly tiny, how about creating a sub-module in kvm-ioctls/ioctls called common or helpers or something to that extent. We can then add all other helpers there. What do you say?

src/ioctls/system.rs
```diff
@@ -7,13 +7,13 @@ use std::os::raw::{c_char, c_ulong};
 use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
 
 use cap::Cap;
-use ioctls::vec_with_array_field;
 use ioctls::vm::{new_vmfd, VmFd};
-#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
-use ioctls::CpuId;
 use ioctls::Result;
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+use ioctls::{CpuId, MsrList};
```
andreeaflorescu, line 20: I believe these two should also be exported as pub in src/lib.rs. Otherwise they'll not be available to external crates.

andreeaflorescu (commented):
Please add usage examples for the public interface in doc comments. Some examples in this PR: https://github.com/rust-vmm/kvm-ioctls/pull/13

src/ioctls/kvm_vec.rs
serban300: Fixed

src/ioctls/kvm_vec.rs
serban300: Just forgot about it. I uncommented it.

src/ioctls/kvm_vec.rs
serban300: Done

src/ioctls/system.rs
serban300: Done

serban300:
Can't think of meaningfull usage examples that would compile since KvmArray and KvmVec are private
> Please add usage examples for the public interface in doc comments. Some examples in this PR: #13

andreeaflorescu:
@serban300 as discussed offline the trait needs to be public so you can access the methods associated with CpuId. If you do some more investigation and you discover that they actually don't need to be public, we can write the usage examples for either CpuId or for MSRs.

serban300:
@andreeaflorescu I added examples

src/ioctls/mod.rs
```diff
@@ -24,148 +28,74 @@ pub mod vcpu;
...
-            kvm_cpuid,
-            allocated_len: self.allocated_len,
-        }
+    fn entries_mut(&mut self) -> &mut __IncompleteArrayField<kvm_cpuid_entry2> {
+        &mut self.entries
     }
 }
 
-#[cfg(test)]
+/// Wrapper for `kvm_cpuid2`.
```
andreeaflorescu, line 64: Please add details about CpuId. Right now when you open the documentation of CpuId it doesn't contain any  relevant information about how to use the structure:

```
Type Definition kvm_ioctls::CpuId

type CpuId = KvmVec<kvm_cpuid2>;

[−]

Wrapper for kvm_cpuid2.
 ```

I would suggest two things:
1. In the documentation description add a link to KvmVec. (e.g. CpuId implements trait 1 and trait 2 with methods for getting kvm_entries bla bla. And for the traits it implements you can add them as links to internal documentation. You can see some examples [here])(https://rust-lang-nursery.github.io/api-guidelines/documentation.html#prose-contains-hyperlinks-to-relevant-things-c-link).
2. Add a minimal usage example for CpuId. You can copy one that already exist in the previous documentation. For example:
```
    /// # Example
    /// ```rust
    /// use kvm_ioctls::{CpuId, Kvm, MAX_KVM_CPUID_ENTRIES};
    /// let kvm = Kvm::new().unwrap();
    /// let mut cpuid = kvm.get_supported_cpuid(MAX_KVM_CPUID_ENTRIES).unwrap();
    /// let cpuid_entries = cpuid.mut_entries_slice();
    /// ```
```

File: src/ioctls/common/kvm_vec.rs [hunk 17]
```diff
@@ -0,0 +1,542 @@
...
+    /// Get the array entries
+    ///
+    fn entries(&self) -> &__IncompleteArrayField<Self::Entry>;
+
+    /// Get the array entries as mut
+    ///
+    fn entries_mut(&mut self) -> &mut __IncompleteArrayField<Self::Entry>;
+}
+
+/// An adapter that helps in treating a KvmArray more like an actual `Vec`.
```
andreeaflorescu: "more like an actual" these are all filler words. Please rephrase.

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,542 @@
...
+
+impl<T: Default + KvmArray> KvmVec<T> {
+    /// Returns the capacity required by kvm_array in order to hold the provided number of  `KvmArray::Entry`
+    ///
+    fn num_elements_to_kvm_array_len(num_elements: usize) -> usize {
+        let size_in_bytes = size_of::<T>() + num_elements * size_of::<T::Entry>();
+        (size_in_bytes + size_of::<T>() - 1) / size_of::<T>()
+    }
+
+    /// Constructs a new KvmVec<T> that contains `num_elements` empty elements of type `KvmArray::Entry`
```
andreeaflorescu: Please add # Arguments section. See other docs for examples.

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,542 @@
...
+
+    /// Returns a mut `Vec<KvmArray::Entry>` that contains all the elements.
+    /// It is important to call `mem::forget` after using this vector. Otherwise rust will destroy it.
+    ///
+    unsafe fn as_vec(&mut self) -> Vec<T::Entry> {
+        let entries_ptr = self.as_mut_original_struct().entries_mut().as_mut_ptr();
+        Vec::from_raw_parts(entries_ptr, self.len, self.capacity as usize)
+    }
+
+    /// Get the mutable elements slice so they can be modified before passing to the VCPU.
```
andreeaflorescu, line 254: nit: s/Get/Returns
Same for all the other ones. Throughout the docs we used Returns instead of Get.

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,542 @@
...
+        KvmVec {
+            kvm_array,
+            len: num_elements,
+            capacity: num_elements,
+        }
+    }
+
+    /// Returns a reference to the actual KVM structure instance.
+    ///
+    pub fn as_original_struct(&self) -> &T {
```
andreeaflorescu: nit: as_kvm_struct? Same for the others.

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,542 @@
...
+        let mut entries = unsafe { self.as_vec() };
+        entries.push(entry);
+        self.update_len(desired_len);
+
+        mem::forget(entries);
+
+        Ok(())
+    }
+
+    /// Retains only the elements specified by the predicate.
```
andreeaflorescu, line 382: Add # Arguments section.

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,542 @@
...
+        /// We need to set the len of the kvm_array to be the number of T elements needed to fit
+        /// an array of `len` elements of type `T::Entry`. This way, when we call
+        /// `self.kvm_array.shrink_to_fit()` only the unnecessary memory will be released.
+        let required_kvm_array_len = KvmVec::<T>::num_elements_to_kvm_array_len(len);
+        unsafe {
+            self.kvm_array.set_len(required_kvm_array_len);
+        }
+    }
+
+    /// Appends an element to the end of the collection.
```
andreeaflorescu: I think is worth here to say that push will also update the length. Add # Arguments section. Also, since this function can fail, please add # Errors section in which we should describe that it returns SizeLimitExceeded when the updated length would be higher then the maximum allowed length.

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,542 @@
...
+    ///
+    pub fn push(&mut self, entry: T::Entry) -> Result<(), Error> {
+        let desired_len = self.len + 1;
+        if desired_len > T::max_len() {
+            return Err(Error::SizeLimitExceeded);
+        }
+
+        self.reserve(1);
+
+        let mut entries = unsafe { self.as_vec() };
```
andreeaflorescu: Why is this safe? Please add comment before unsafe blocks.

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,542 @@
...
+    // the number of elements of type `KvmArray::Entry` currently in the vec
+    len: usize,
+    // the capacity of the `KvmVec` measured in elements of type `KvmArray::Entry`
+    capacity: usize,
+}
+
+impl<T: Default + KvmArray> KvmVec<T> {
+    /// Returns the capacity required by kvm_array in order to hold the provided number of  `KvmArray::Entry`
+    ///
+    fn num_elements_to_kvm_array_len(num_elements: usize) -> usize {
```
andreeaflorescu: nit: I would rename this to kvm_vec_capacity. And also add a comment that it returns the kvm vec capacity required to hold `num_elements` of `KvmArray::Entry`.

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,542 @@
...
+        if desired_capacity <= self.capacity {
+            return;
+        }
+
+        let current_kvm_array_len = self.kvm_array.len();
+        let required_kvm_array_len = KvmVec::<T>::num_elements_to_kvm_array_len(desired_capacity);
+        let additional_kvm_array_len = required_kvm_array_len - current_kvm_array_len;
+
+        self.kvm_array.reserve(additional_kvm_array_len);
+        self.capacity = desired_capacity;
```
andreeaflorescu: The capacity should be updated to self.kvm_array.capacity().

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,542 @@
...
+impl<T: Default + KvmArray> PartialEq for KvmVec<T> {
+    fn eq(&self, other: &KvmVec<T>) -> bool {
+        self.len == other.len
+            && self.capacity == other.capacity
+            && self.as_entries_slice() == other.as_entries_slice()
+    }
+}
+
+impl<T: Default + KvmArray> Clone for KvmVec<T> {
+    fn clone(&self) -> Self {
```
andreeaflorescu, line 430: nit: too many whitespaces.

src/ioctls/mod.rs
andreeaflorescu: Same for MsrList.

src/lib.rs
```diff
@@ -42,3 +43,6 @@ pub use ioctls::{KvmRunWrapper, Result};
 /// It can be used for calls to [get_supported_cpuid](struct.Kvm.html#method.get_supported_cpuid) and
 /// [get_emulated_cpuid](struct.Kvm.html#method.get_emulated_cpuid).
 pub const MAX_KVM_CPUID_ENTRIES: usize = 80;
+
+/// Maximum number of MSR entries that can be returned by a call to KVM ioctls.
```
andreeaflorescu, line 216: Please add similar details as the `MAX_KVM_CPUID_ENTRIES` have. Where did you get the value from? What can it be used for?

src/ioctls/mod.rs
serban300: Done

src/lib.rs
serban300: I took it fro mthe prefious implementation of get_msr_index_list . I don't have solid details about this one.

src/ioctls/common/kvm_vec.rs
serban300: Done

src/ioctls/common/kvm_vec.rs
serban300: Done

src/ioctls/common/kvm_vec.rs
serban300: I think `kvm_vec_capacity` would be confusing. I renamed kvm_array to 'mem_allocator'  and I renamed this method to `kvm_vec_len_to_mem_allocator_len`

src/ioctls/common/kvm_vec.rs
serban300: Done. What do you mean by "the others" ?

src/ioctls/common/kvm_vec.rs
serban300: Done

src/ioctls/common/kvm_vec.rs
serban300: Done

src/ioctls/common/kvm_vec.rs
serban300: Done

src/ioctls/common/kvm_vec.rs
serban300: Done

src/ioctls/common/kvm_vec.rs
serban300: Done

src/ioctls/common/kvm_vec.rs
serban300: I removed some

src/ioctls/common/kvm_vec.rs
andreeaflorescu: Probably I wanted to write this somewhere else, sorry for the confusion.

acatangiu (dismissed):

src/ioctls/mod.rs
```diff
@@ -26,198 +28,84 @@ pub mod vcpu;
...
+///
+/// See the documentation for [KvmVec](struct.KvmVec.html)
```
andreeaflorescu: Can you be a bit more explicit here? For example:
```
/// Wrapper over the `kvm_cpuid2` structure.
/// 
/// The `kvm_cpuid2` structure has a zero sized array. For details check the 
/// [KVM API](https://www.kernel.org/doc/Documentation/virtual/kvm/api.txt)
/// documentation on `kvm_cpuid2`. To provide safe access to
/// `kvm_cpuid_entry` elements, this type is implemented using
/// [KvmVec](struct.KvmVec.html).
```

src/ioctls/mod.rs
```diff
@@ -26,198 +28,84 @@ pub mod vcpu;
...
-    pub fn as_mut_ptr(&mut self) -> *mut kvm_cpuid2 {
-        &mut self.kvm_cpuid[0]
+    fn entries_mut(&mut self) -> &mut __IncompleteArrayField<u32> {
+        &mut self.indices
     }
 }
 
+/// Wrapper for `kvm_msr_list`.
+///
+/// See the documentation for [KvmVec](struct.KvmVec.html)
```
andreeaflorescu: Also elaborate on KvmMsrList here and add examples.

src/lib.rs
andreeaflorescu: I can't find it either. Sadness :(

andreeaflorescu (commented):
As we discussed offline I believe `from_entries` should not be deleted from CpuId.
I am not sure if this is useful for kvm_msr_list, but for kvm_cpuid2 we already know some scenarios where we can use it. Considering this I would suggest to add a CpuIdExtension (or similar) trait implemented for the type CpuId

src/ioctls/mod.rs
serban300: Thanks ! Done.

src/ioctls/mod.rs
serban300: Done.

jiangliu:
Seems it's a common issue to deal with zero-sized array. Is there any best known methods for rust to support such a case? Or could we extend a basic rust-vmm crate to export helper macros/traits to deal with zero-sized array?

serban300:
> Seems it's a common issue to deal with zero-sized array. Is there any best known methods for rust to support such a case? Or could we extend a basic rust-vmm crate to export helper macros/traits to deal with zero-sized array?

If zero-sized arrays are needed outside of the kvm-ioctls crate then implementing them in a separate crate with helpers / commons sounds good to me. Anyway since I'm not very familiar with rust-vmm I would need to get more opinions on this issue.

andreeaflorescu:
We can ask people in the rust-vmm list what they think about this. @serban300 can you please send an email?

I believe the plan right now it to move the CpuId definition to vmm-vcpu. See https://github.com/rust-vmm/vmm-vcpu/pull/3

sboeuf:
@serban300 
>If zero-sized arrays are needed outside of the kvm-ioctls crate then implementing them in a separate crate with helpers / commons sounds good to me. Anyway since I'm not very familiar with rust-vmm I would need to get more opinions on this issue.

Having a common crate for this sounds like a sane idea. You should look at the existing crates already present on [crates.io](https://crates.io/), maybe you will find something there.

What would be the name of the crate btw? `rust-common`?

>I believe the plan right now it to move the CpuId definition to vmm-vcpu. See rust-vmm/vmm-vcpu#3

That's my understanding too!

andreeaflorescu:
I am thinking that maybe we should instead place it in vmm-sys-utils instead of having a separate crate for it. I don't have a lot of experience here, but is this really a general problem? Do we think that in the future we will have to handle other 0-sized arrays besides the ones we have in KVM? I would be interested to see some other usecases for which this would need to live in a separate crate.

jennymankin:
I think it makes sense to move the traits of `kvm_vec` to a common crate (or to vmm-sys-utils if that's the better option). Then the implementations of the `kvm_vec` for CPUID and MSR vectors would ultimately go in the Vcpu crate.

As things will no longer reside in a specific kvm-ioctls crate, and will instead be intended to be used in VMM-agnostic crates, could you rename things to avoid `KVM`-specific naming? File names like `kvm_vec.rs`, trait names like `KvmVec`, etc.

jennymankin:
On a similar note, something I found while moving the old `CpuId` implementation into the `Vcpu` crate is that doc tests with KVM-specific or Unix-specific crates or APIs will cause `cargo test` to fail on Windows.  In cases where the example is instructive and can't be modified to avoid such dependencies, `cfg(unix)` will at least prevent the doc tests from failing on Windows. For example:

```
    /// ```cfg(unix)
    /// extern crate kvm_ioctls;
    /// 
    /// use kvm_ioctls::Kvm;
    /// use vmm_vcpu::x86_64::{CpuId, MAX_CPUID_ENTRIES};
    /// 
    /// # fn main() {
    ///     let kvm = Kvm::new().unwrap();
    ///     let mut cpuid = kvm.get_supported_cpuid(MAX_CPUID_ENTRIES).unwrap();
    ///     let cpuid_entries = cpuid.mut_entries_slice();
    /// # }
    /// 
    /// ```
```

jiangliu:
It’s a common case for Linux ioctl related data structures, and Vhost has similar cases.
I think it’s ok to include it into the vmm-sys-utils crate.

serban300:
Thanks everyone for the feedback !

It would be nice if there was something similar on crates.io . I couldn't find anything so far. I will keep looking, but I think the chances are very slim.

Anyway, it seems that there is a consensus around using the vmm-sys-util crate. It looks good to me also. I will submit a PR there as soon as https://github.com/rust-vmm/vmm-sys-util/pull/1 is merged. Also I will make some changes in order to avoid KVM specific naming.

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,549 @@
...
+/// ```
+#[allow(clippy::len_without_is_empty)]
+pub trait KvmArray {
+    type Entry: PartialEq + Copy;
+
+    /// Get the array length
+    ///
+    fn len(&self) -> usize;
+
+    /// Get the array length as mut
```
alexandruag, line 74: This comment looks wrong.

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,549 @@
...
+    len: usize,
+    // the capacity of the `KvmVec` measured in elements of type `KvmArray::Entry`
+    capacity: usize,
+}
+
+impl<T: Default + KvmArray> KvmVec<T> {
+    /// Get the capacity required by mem_allocator in order to hold
+    /// the provided number of  `KvmArray::Entry`
+    ///
+    fn kvm_vec_len_to_mem_allocator_len(kvm_vec_len: usize) -> usize {
```
alexandruag, line 111: I guess this could be renamed to `mem_allocator_len`, since the name of the parameter kinda suggests what the conversion is about.

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,549 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `num_elements` - The number of empty elements of type `KvmArray::Entry` in the initial `KvmVec`
+    ///
+    pub fn new(num_elements: usize) -> KvmVec<T> {
+        let required_mem_allocator_capacity =
+            KvmVec::<T>::kvm_vec_len_to_mem_allocator_len(num_elements);
+
+        let mut mem_allocator = Vec::with_capacity(required_mem_allocator_capacity);
```
alexandruag, line 158: I think lines 138-141 can be replaced with `mem_allocator = vec![T::default(); required_mem_allocator_capacity]`, if it looks any better.

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,549 @@
...
+    fn mem_allocator_len_to_kvm_vec_len(mem_allocator_len: usize) -> usize {
+        if mem_allocator_len == 0 {
+            return 0;
+        }
+
+        let array_size_in_bytes = (mem_allocator_len - 1) * size_of::<T>();
+        array_size_in_bytes / size_of::<T::Entry>()
+    }
+
+    /// Constructs a new KvmVec<T> that contains `num_elements` empty elements
```
alexandruag, line 128: It might be a bit more accurate to say "default-initialized" instead of "empty" here.

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,549 @@
...
+    /// # Arguments
+    ///
+    /// * `entries` - The vector of `KvmArray::Entry` entries.
+    ///
+    pub fn from_entries(entries: &[T::Entry]) -> KvmVec<T> {
+        let mut kvm_vec = KvmVec::<T>::new(entries.len());
+
+        {
+            let kvm_vec_entries = kvm_vec.as_mut_kvm_struct().entries_mut();
+            // this is safe because the provided length is correct
```
alexandruag, line 210: This is for all the OCD ppl:  it's nice if all the sentence-like (or longer comments) use proper capitalization and punctuation. I think there might be a couple of other places where that's not the case.

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,549 @@
...
+            let kvm_vec_entries_slice = unsafe { kvm_vec_entries.as_mut_slice(entries.len()) };
+            kvm_vec_entries_slice.copy_from_slice(entries);
+        }
+
+        kvm_vec
+    }
+
+    /// Get a reference to the actual KVM structure instance.
+    ///
+    pub fn as_kvm_struct(&self) -> &T {
```
alexandruag, line 220: I think that, if it's fine to call the method below that returns a `*const T` pointer `as_ptr`, then this one can be called `as_ref`, and the following `as_mut`.

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,549 @@
...
+    ///
+    fn as_vec(&mut self) -> Vec<T::Entry> {
+        unsafe {
+            let entries_ptr = self.as_mut_kvm_struct().entries_mut().as_mut_ptr();
+            // This is safe since self.len and self.capacity should be correct
+            Vec::from_raw_parts(entries_ptr, self.len, self.capacity as usize)
+        }
+    }
+
+    /// Get the mutable elements slice so they can be modified before passing to the VCPU.
```
alexandruag, line 254: Comment seems wrong.

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,549 @@
...
+    ///
+    pub fn as_entries_slice(&self) -> &[T::Entry] {
```
alexandruag, line 268: I think this can be called `as_slice`, since there no other useful slice-like view into this other than a slice of entries. It's also shorter, and more in line with `Vec` method naming.

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,549 @@
...
+        let len = self.as_kvm_struct().len();
+        unsafe { self.as_kvm_struct().entries().as_slice(len as usize) }
```
alexandruag, line 270: Courtesy comment regarding why the `unsafe` block is safe is nice.

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,549 @@
...
+    }
+
+    /// Get the mutable elements slice so they can be modified before passing to the VCPU.
+    ///
+    pub fn as_mut_entries_slice(&mut self) -> &mut [T::Entry] {
```
alexandruag, line 288: I think we can call this `as_mut_slice`, for similar reason to `as_slice` above.

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,549 @@
...
+        let len = self.as_kvm_struct().len();
+        unsafe {
```
alexandruag, line 290: Courtesy comment regarding why the `unsafe` block is safe is nice.

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,549 @@
...
+            self.as_mut_kvm_struct()
+                .entries_mut()
+                .as_mut_slice(len as usize)
+        }
+    }
+
+    /// Reserves capacity for at least `additional` more `KvmArray::Entry` elements.
+    /// If the capacity is already reserved, this method doesn't do anything
+    ///
+    fn reserve(&mut self, additional: usize) {
```
alexandruag, line 300: Should this be `pub`?

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,549 @@
...
+            KvmVec::<T>::kvm_vec_len_to_mem_allocator_len(desired_capacity);
+        let additional_mem_allocator_len = required_mem_allocator_len - current_mem_allocator_len;
+
+        self.mem_allocator.reserve(additional_mem_allocator_len);
+        self.capacity =
+            KvmVec::<T>::mem_allocator_len_to_kvm_vec_len(self.mem_allocator.capacity());
+    }
+
+    /// Updates the length of `self` to the specified value.
+    /// Also updates the length of the `T::Entry` structure and of `self.mem_allocator` accordingly.
```
alexandruag, line 317: Doesn't this method update the length of `T`, and not that of `T::Entry`?

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,549 @@
...
+    ///
+    fn update_len(&mut self, len: usize) {
+        self.len = len;
```
alexandruag, line 320: Should we check whether `len > T::max_len` first?

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,549 @@
...
+        self.as_mut_kvm_struct().set_len(len);
+
+        /// We need to set the len of the mem_allocator to be the number of T elements needed
+        /// to fit an array of `len` elements of type `T::Entry`. This way, when we call
+        /// `self.mem_allocator.shrink_to_fit()` only the unnecessary memory will be released.
+        let required_mem_allocator_len = KvmVec::<T>::kvm_vec_len_to_mem_allocator_len(len);
+        unsafe {
```
alexandruag, line 327: Why is this actually safe?

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,549 @@
...
+    pub fn as_mut_ptr(&mut self) -> *mut T {
+        self.as_mut_kvm_struct()
+    }
+
+    /// Get a mut `Vec<KvmArray::Entry>` that contains all the elements.
+    /// It is important to call `mem::forget` after using this vector.
+    /// Otherwise rust will destroy it.
+    ///
+    fn as_vec(&mut self) -> Vec<T::Entry> {
+        unsafe {
```
alexandruag, line 247: Why is this actually safe?

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,549 @@
...
+    ///
+    pub fn push(&mut self, entry: T::Entry) -> Result<(), Error> {
+        let desired_len = self.len + 1;
+        if desired_len > T::max_len() {
+            return Err(Error::SizeLimitExceeded);
+        }
+
+        self.reserve(1);
+
+        let mut entries = self.as_vec();
```
alexandruag, line 373: Is the `self.as_vec`/`push` combination required here? Since we've already ensured there is enough space via `reserve`, it looks like we can call `self.update_len(desired_len);`, and then update the value of the last position based on `entry` using the mut slice returned by `self.as_mut_entries_slice`.

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,549 @@
...
+    }
+
+    /// Retains only the elements specified by the predicate.
+    ///
+    /// # Arguments
+    ///
+    /// * `f` - The function used to evaluate whether an entry will be kept or not.
+    ///         When `f` returns `true` the entry is kept.
+    ///
+    pub fn retain<P>(&mut self, f: P)
```
alexandruag, line 407: If `retain` remains the sole user of `as_vec`, which is pretty weird/scary as currently implemented, I would consider reimplementing this without `as_vec` (for example, using `slice::sort_by`, and dropping the tail; or writing the iterations + swap ourselves if that's deemed too inefficient), and removing `as_vec` altogether, to get rid of the extra `unsafe`-ness and prevent future use.

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,549 @@
...
+        self.capacity =
+            KvmVec::<T>::mem_allocator_len_to_kvm_vec_len(self.mem_allocator.capacity());
+
+        mem::forget(entries);
+    }
+}
+
+impl<T: Default + KvmArray> PartialEq for KvmVec<T> {
+    fn eq(&self, other: &KvmVec<T>) -> bool {
+        self.len == other.len && self.as_entries_slice() == other.as_entries_slice()
```
alexandruag, line 425: Is `self.len == other.len` implied by `self.as_entries_slice() == other.as_entries_slice()`?

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,549 @@
...
+    }
+}
+
+impl<T: Default + KvmArray> Clone for KvmVec<T> {
+    fn clone(&self) -> Self {
+        let mut clone = KvmVec::<T>::new(self.len);
+
+        let num_bytes = self.mem_allocator.len() * size_of::<T>();
+        let src_byte_slice =
+            unsafe { std::slice::from_raw_parts(self.as_ptr() as *const u8, num_bytes) };
```
alexandruag, line 435: Courtesy comment regarding why the unsafe block is safe is nice.

src/ioctls/common/kvm_vec.rs
```diff
@@ -0,0 +1,549 @@
...
+        let dst_byte_slice =
+            unsafe { std::slice::from_raw_parts_mut(clone.as_mut_ptr() as *mut u8, num_bytes) };
```
alexandruag, line 437: Ditto.

src/ioctls/common/kvm_vec.rs
serban300: Fixed

src/ioctls/common/kvm_vec.rs
serban300: Done

src/ioctls/common/kvm_vec.rs
serban300: Done

src/ioctls/common/kvm_vec.rs
serban300: Yes, looks better. Done.

src/ioctls/common/kvm_vec.rs
serban300: Will try to fix all the comments.

src/ioctls/common/kvm_vec.rs
serban300: I find this naming misleading. I tried to improve these function names though.

src/ioctls/common/kvm_vec.rs
serban300: Fixed

src/ioctls/common/kvm_vec.rs
serban300: Should be safe because the len and capacity are correct.

src/ioctls/common/kvm_vec.rs
serban300: Done

src/ioctls/common/kvm_vec.rs
serban300: I removed this code

src/ioctls/common/kvm_vec.rs
serban300: Done

src/ioctls/common/kvm_vec.rs
serban300: I removed this code

src/ioctls/common/kvm_vec.rs
serban300: I wouldn't make it public. It should only be called internally before pushing elements to the Vec

src/ioctls/common/kvm_vec.rs
serban300: Right. Fixed.

src/ioctls/common/kvm_vec.rs
serban300: Good point. Done.

src/ioctls/common/kvm_vec.rs
serban300: This is safe since the len should be correct

I added a comment

src/ioctls/common/kvm_vec.rs
serban300: Nice. Seems to work.

src/ioctls/common/kvm_vec.rs
serban300: Good point. Done. I implemented some custom logic for this since it's quite easy.

src/ioctls/common/kvm_vec.rs
serban300: Not necessarily. If something goes wrong self.len can differ from T.len()

src/ioctls/common/kvm_vec.rs
serban300: I changed the implementation in order to avoid using unsafe:

```
clone.mem_allocator = self.mem_allocator.to_vec();
```

src/ioctls/common/kvm_vec.rs
serban300: The same comment as above.

serban300:
Closing in favor of https://github.com/rust-vmm/vmm-sys-util/pull/4

Issue #9, opened 2019-04-01
Add ARM specific coverage test


andreeaflorescu:
Our CI is checking is the coverage was decreased by looking at the value in [tests/coverage](https://github.com/rust-vmm/kvm-ioctls/blob/master/tests/coverage). Check `test_coverage` in [test_coverage.py](https://github.com/rust-vmm/kvm-ioctls/blob/master/tests/test_coverage.py).

As coverage will be different on arm and x86 because of different ioctls and code paths that can be exercised on each platform, we should also have different coverage files and update the test to use different files for different platforms.

andreeaflorescu:
I tried kcov on arm and it hangs. The cargo kcov command doesn't work at all.
Related issues:
- kcov issue: https://github.com/SimonKagstrom/kcov/issues/267
- cargo kcov issue: https://github.com/kennytm/cargo-kcov/issues/40

MrXinWang:
Hi @andreeaflorescu ! we are currently trying to contribute for Arm64 platform, and I think I will try to address this issue. Therefore I am wondering if I can try the integration test locally or must do it on buildkite?

Also I think the [README](https://github.com/rust-vmm/rust-vmm-ci/blob/master/README.md) in [rust-vmm/rust-vmm-ci](https://github.com/rust-vmm/rust-vmm-ci) needs to be updated since the folder `tests` is not used anymore (I will do that together with this issue).

The idea is: Referencing #8 , I think we have already had a [PR#62](https://github.com/rust-vmm/kvm-ioctls/pull/62) to solve the `run_code` example for Arm64 and we also have unit tests for Arm64 now, so I think this is the only issue left for arm64 support?

andreeaflorescu:
> Hi @andreeaflorescu ! we are currently trying to contribute for Arm64 platform, and I think I will try to address this issue. Therefore I am wondering if I can try the integration test locally or must do it on buildkite?

You can try the test locally. You can run the command specified in [Adaptive Coverage](https://github.com/rust-vmm/rust-vmm-ci/blob/master/README.md#adaptive-coverage). For that to work you need to update the path of the coverage test. The new path, relative to the kvm-ioctls directory is `rust-vmm-ci/integration_tests/test_coverage.py`.

> 
[...]

PR #12, opened 2019-04-01
aarch64: add necessary kvm ioctls

dianpopa:
`cargo test` is functional on aarch64 too.
Fixes https://github.com/firecracker-microvm/firecracker/issues/1040.

Signed-off-by: Diana Popa <dpopa@amazon.com>


andreeaflorescu:
Before merging this PR I would like to setup a CI for ARM as well to make sure that future PRs don't break the ARM build.

dianpopa:
> Before merging this PR I would like to setup a CI for ARM as well to make sure that future PRs don't break the ARM build.

Seems like a good idea. However, until then, any feedback is welcome.

src/ioctls/device.rs
```diff
@@ -0,0 +1,84 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
andreeaflorescu, line 1: The current structure of kvm-ioctls is to have  3 sub-modules in the `ioctls` module that correspond to the KVM ioctls: system, vm and vcpu. When helper structures are required we can do one of two things:
- if they're used by more than one sub-module (system, vm or vcpu) the definition can lie in ioctls/mod.rs
- it they're used by only one sub-module than they should be added to the sub-module that uses them.

I believe `DeviceFd` is used only by `VmFd` and if this is not going to change in the future I say we place this code there rather than creating a new file so we can keep the initial structure.

src/ioctls/device.rs
```diff
@@ -0,0 +1,84 @@
...
+use kvm_ioctls::KVM_SET_DEVICE_ATTR;
+use sys_ioctl::ioctl_with_ref;
+
+/// Wrapper over the file descriptor obtained when creating an emulated device in the kernel.
+pub struct DeviceFd {
+    fd: File,
+}
+
+impl DeviceFd {
+    /// See the documentation for `KVM_SET_DEVICE_ATTR`.
```
andreeaflorescu, line 22: Please add a short description of the ioctl: 
```
/// Sets a specified piece of device configuration and/or state.
///
/// See the documentation for `KVM_SET_DEVICE_ATTR`.
```

src/ioctls/device.rs
```diff
@@ -0,0 +1,84 @@
...
+    use super::*;
+    use ioctls::system::Kvm;
+    use std::io::{Error, ErrorKind};
+
+    #[test]
+    fn test_create_device() {
+        let kvm = Kvm::new().unwrap();
+        let vm = kvm.create_vm().unwrap();
+
+        let mut gic_device = kvm_bindings::kvm_create_device {
```
andreeaflorescu, line 63: nit: can you import the structures and constants that you need from the kvm_bindings instead of using them with kvm_bindings::KVM_CREATE_DEVICE_TEST? The code is a bit hard to read.

src/ioctls/device.rs
```diff
@@ -0,0 +1,84 @@
...
+            #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+            type_: kvm_bindings::kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V3,
+            #[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
+            type_: kvm_bindings::kvm_device_type_KVM_DEV_TYPE_FSL_MPIC_20,
+            fd: 0,
+            flags: kvm_bindings::KVM_CREATE_DEVICE_TEST,
+        };
+        // We expect this to fail. No VGIC_V3 on x86_64.
```
andreeaflorescu: This is also expected to fail on ARM. Please update the comment.

src/ioctls/vcpu.rs
```diff
@@ -293,6 +300,37 @@ impl VcpuFd {
         Ok(())
     }
 
+    /// This initializes an ARM vCPU to the specified type with the specified features
```
andreeaflorescu, line 305: Please format the doc comment with summary and description. In the description also add the KVM ioctl that the method is using (KVM_ARM_VCPU_INIT). Since this ioctl has a parameter also add the `# Arguments` section.

Same for `set_one_reg`.

src/ioctls/vcpu.rs
```diff
@@ -293,6 +300,37 @@ impl VcpuFd {
...
+    /// and resets the values of all of its registers to defaults.
+    #[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
+    pub fn vcpu_init(&self, kvi: &kvm_vcpu_init) -> Result<()> {
+        // safe becuase we allocated the struct and we know the kernel will read
```
andreeaflorescu: s/becuase/because

src/ioctls/vcpu.rs
```diff
@@ -718,4 +756,38 @@ mod tests {
...
+
+        let mut kvi: kvm_bindings::kvm_vcpu_init = kvm_bindings::kvm_vcpu_init::default();
+        vm.get_preferred_target(&mut kvi)
+            .expect("Cannot get preferred target");
+        vcpu.vcpu_init(&kvi).expect("Cannot initialize vcpu");
+        let mut data: u64 = 0;
+        let mut reg_id: u64 = 0;
+
+        assert!(vcpu.set_one_reg(reg_id, data).is_err());
+        const PSTATE_REG_ID: u64 = 0x6030_0000_0010_0042;
```
andreeaflorescu, line 807: Can you add a comment here regarding what are you trying to test and what is the `PSTATE_REG_ID`. This would be useful for people looking at the tests in the future.

src/ioctls/vcpu.rs
```diff
@@ -718,4 +756,38 @@ mod tests {
...
+        vcpu.set_one_reg(PSTATE_REG_ID, data).expect("bla");
```
andreeaflorescu: nit: We can change this expect into unwrap to get rid of the "bla".

src/ioctls/vm.rs
```diff
@@ -246,6 +255,28 @@ impl VmFd {
...
+        if ret == 0 {
+            Ok((new_device(unsafe { File::from_raw_fd(device.fd as i32) })))
+        } else {
+            return Err(io::Error::last_os_error());
+        }
+    }
+
+    /// This queries the kernel for the preferred target CPU type.
+    #[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
+    pub fn get_preferred_target(&self, kvi: &mut kvm_vcpu_init) -> Result<()> {
```
andreeaflorescu, line 270: We should try to name methods as close as possible to the KVM ioctls that they provide wrappers for. Usually the names of all methods strip the `KVM_` prefix and use lower case. For this one it would be arm_preferred_target.

src/ioctls/vm.rs
```diff
@@ -358,11 +389,11 @@ mod tests {
...
-
-        assert!(vm_fd.register_irqfd(evtfd1, 4).is_ok());
-        assert!(vm_fd.register_irqfd(evtfd2, 8).is_ok());
-        assert!(vm_fd.register_irqfd(evtfd3, 4).is_ok());
-
+        if cfg!(any(target_arch = "x86", target_arch = "x86_64")) {
+            assert!(vm_fd.register_irqfd(evtfd1, 4).is_ok());
+            assert!(vm_fd.register_irqfd(evtfd2, 8).is_ok());
+            assert!(vm_fd.register_irqfd(evtfd3, 4).is_ok());
+        }
```
andreeaflorescu, line 401: With this change the following asserts fail for different reasons on arm and x86. I think it is worth mentioning why we expect the asserts to fail on each platform.

andreeaflorescu (commented):
General comments:
- Please update the coverage to 91.3. This is x86 specific unfortunately, but I added an issue to take care of it for ARM as well:  #9 .The coverage value is stored in tests/coverage.
- All doc comments (larger than one line) should be split in summary and description. Also, for methods that wrap KVM ioctls add a comment with the name of the ioctl that they provide wrappers for. If the methods have arguments, please add the `# Arguments` section to the comment.

andreeaflorescu:
@dianpopa I added a temporary fix for running cargo build with gnu/musl, and cargo test with gnu/musl on arm as well.

We still need to work on adding a coverage test for aarch64, but other than that we should be good to merge the PR once the tests pass.

The musl unit tests seem to fail for a reason that is unrelated to your PR. Do you happen to know why is there a linking problem? You can check the logs here: https://buildkite.com/rust-vmm/kvm-ioctls-ci/builds/108#55d3065d-1563-4a50-a0c9-2edd7c3f6afa

src/ioctls/device.rs
dianpopa: DeviceFd is not just a helper structure. KVM_CREATE_DEVICE (just like KVM_CREATE_VPCU and KVM_CREATE_VM) will give back an fd on which later you can call other kvm ioctls. Take a look at 
```
4.80 /KVM_GET_DEVICE_ATTR
Type: device ioctl, vm ioctl, vcpu ioctl
```
from the api.txt documentation. So `DeviceFd` corresponds to another type of ioctl, device ioctl (apart from vcpu and vm).
"I believe DeviceFd is used only by VmFd" -> DeviceFd will be used by regs and the global interrupt controller when all arm64 functionality will be merged.
I actually considered putting the code inside VMFd, however in VmFd is a lot of code already and it made it easier to add unit tests and add documentation putting it in a separate file.
If you still think it makes more sense to put it with VmFd, let me know.

src/ioctls/vm.rs
dianpopa: I find it ugly to put a cfg above this method that says it is arm specific and then name that method with an arm prefix. 
I think it is nice to keep that rule around but we can make exceptions when we feel there is a need to (see the register_irqfd function).

src/ioctls/device.rs
andreeaflorescu: Gotcha! Sounds good as it it. I checked the KVM API again, I missed the device ioctls entry before.

src/ioctls/mod.rs
```diff
@@ -1,12 +1,22 @@
-use kvm_bindings::kvm_run;
-#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
-use kvm_bindings::{kvm_cpuid2, kvm_cpuid_entry2};
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
andreeaflorescu, line 1: nit: Adding the license is an unrelated change. Can you please add it in a new commit?

src/ioctls/device.rs
```diff
@@ -0,0 +1,97 @@
...
+use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
+
+use kvm_bindings::kvm_device_attr;
+
+use ioctls::Result;
+use kvm_ioctls::KVM_SET_DEVICE_ATTR;
+use sys_ioctl::ioctl_with_ref;
+
+/// Wrapper over the file descriptor obtained when creating an emulated device in the kernel.
+pub struct DeviceFd {
```
andreeaflorescu, line 15: The structure need to be publicly exported in src/lib.rs to be accessible outside of the crate.

dianpopa: Done, see lib.rs file.

src/ioctls/mod.rs
dianpopa: Done, take a look.

andreeaflorescu (approved):

sameo (approved):

PR #13, opened 2019-04-02
kvm system ioctls: updated public functions docs

andreeaflorescu:
## Changes
Added examples and more details for public functions.
Also renamed get_run_size() to run_size().

## Note:
`cargo test` will also run the documentation examples. Even if some examples are simple and just call the function it can help us in the future to keep the documentation updated as we make code changes (cargo test fails otherwise).

## How to test it locally
The best way to review this PR is by generating the documentation as follows:

```
# Clone Repo
git clone git@github.com:rust-vmm/kvm-ioctls.git
cd kvm-ioctls
# Fetch the code from my fork 
git fetch https://github.com/andreeaflorescu/kvm-ioctls.git pub_docs
git checkout FETCH_HEAD
# Generate and open the documentation 
cargo doc --open
```

## Still missing

High level crate design & usage. Will add this as part of fixing #21

### DIFF
--- a/src/cap.rs
+++ b/src/cap.rs
@@ -7,9 +7,13 @@
 
 use kvm_bindings::*;
 
-/// Enumeration of the extension capability list that KVM exposes.
+/// Capabilities exposed by KVM.
 ///
-/// For details check the
+/// The capabilities list can be used in conjunction with
+/// [Kvm::check_extension()](struct.Kvm.html#method.check_extension) to check if a particular
+/// capability is available.
+///
+/// The list of capabilities is based on the the KVM_CAP_* defines from the
 /// [Linux KVM header](https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/kvm.h).
 ///
 #[derive(Clone, Copy, Debug)]
--- a/src/ioctls/system.rs
+++ b/src/ioctls/system.rs
@@ -300,7 +300,7 @@ mod tests {
         let kvm = Kvm::new().unwrap();
         let vm = kvm.create_vm().unwrap();
 
-        assert_eq!(vm.get_run_size(), kvm.get_vcpu_mmap_size().unwrap());
+        assert_eq!(vm.run_size(), kvm.get_vcpu_mmap_size().unwrap());
     }
 
     #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
--- a/src/ioctls/system.rs
+++ b/src/ioctls/system.rs
@@ -22,15 +22,20 @@ use ioctls::Result;
 use kvm_ioctls::*;
 use sys_ioctl::*;
 
-/// A wrapper over the `/dev/kvm` file.
-///
-/// The handle is used to issue KVM system ioctls.
+/// Wrapper over KVM system ioctls.
 pub struct Kvm {
     kvm: File,
 }
 
 impl Kvm {
-    /// Opens `/dev/kvm/` and returns a `Kvm` object on success.
+    /// Opens `/dev/kvm` and returns a `Kvm` object on success.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use kvm_ioctls::Kvm;
+    /// let kvm = Kvm::new().unwrap();
+    /// ```
     ///
     #[allow(clippy::new_ret_no_self)]
     pub fn new() -> Result<Self> {
@@ -42,6 +47,8 @@ impl Kvm {
 
     /// Creates a new Kvm object assuming `fd` represents an existing open file descriptor
     /// associated with `/dev/kvm`.
+    ///
+    /// For usage examples check [open_with_cloexec()](struct.Kvm.html#method.open_with_cloexec).
     ///
     /// # Arguments
     ///
@@ -55,9 +62,24 @@ impl Kvm {
 
     /// Opens `/dev/kvm` and returns the fd number on success.
     ///
+    /// One usecase for this method is opening `/dev/kvm` before exec-ing into a
+    /// process with seccomp filters enabled that blacklist the `sys_open` syscall.
+    /// For this usecase `open_with_cloexec` must be called with the `close_on_exec`
+    /// parameter set to false.
+    ///
     /// # Arguments
     ///
     /// * `close_on_exec`: If true opens `/dev/kvm` using the `O_CLOEXEC` flag.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use kvm_ioctls::Kvm;
+    /// let kvm_fd = Kvm::open_with_cloexec(false).unwrap();
+    /// // The `kvm_fd` can now be passed to another process where we can use
+    /// // `new_with_fd_number` for creating a `Kvm` object:
+    /// let kvm = unsafe { Kvm::new_with_fd_number(kvm_fd) };
+    /// ```
     ///
     pub fn open_with_cloexec(close_on_exec: bool) -> Result<RawFd> {
         let open_flags = O_RDWR | if close_on_exec { O_CLOEXEC } else { 0 };
@@ -73,6 +95,15 @@ impl Kvm {
     /// Returns the KVM API version.
     ///
     /// See the documentation for `KVM_GET_API_VERSION`.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use kvm_ioctls::Kvm;
+    /// let kvm = Kvm::new().unwrap();
+    /// assert_eq!(kvm.get_api_version(), 12);
+    /// ```
+    ///
     pub fn get_api_version(&self) -> i32 {
         // Safe because we know that our file is a KVM fd and that the request is one of the ones
         // defined by kernel.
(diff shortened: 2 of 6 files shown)


src/ioctls/system.rs
```diff
@@ -81,12 +114,23 @@ impl Kvm {
...
     /// # Arguments
     ///
-    /// * `c` - KVM capability.
+    /// * `c` - KVM capability to check.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use kvm_ioctls::Kvm;
+    /// use kvm_ioctls::Cap;
```
alxiord, line 135: Why does `use Kvm` need `#` while `use Cap` doesn't?

src/ioctls/system.rs
```diff
@@ -96,6 +140,14 @@ impl Kvm {
...
     ///
+    /// # Example
+    ///
+    /// ```
+    /// # use kvm_ioctls::Kvm;
+    /// let kvm = Kvm::new().unwrap();
+    /// assert!(kvm.get_vcpu_mmap_size().unwrap() > 0);
+    /// ```
+    ///
     pub fn get_vcpu_mmap_size(&self) -> Result<usize> {
```
alxiord, line 158: Since you got rid of the `get_` prefix from `get_run_size`, shouldn't the getter naming convention be consistent?

src/ioctls/system.rs
```diff
@@ -187,6 +279,19 @@ impl Kvm {
...
+    /// # Example
+    /// ```
+    /// #[cfg(any(target_arch = "x86", target_arch = "x86_64"))] {
+    ///     use kvm_ioctls::{Kvm, MAX_KVM_CPUID_ENTRIES};
+    ///
+    ///     let kvm = Kvm::new().unwrap();
+    ///     let mut cpuid = kvm.get_supported_cpuid(MAX_KVM_CPUID_ENTRIES).unwrap();
+    ///     let cpuid_entries = cpuid.mut_entries_slice();
+    ///     assert!(cpuid_entries.len() <= MAX_KVM_CPUID_ENTRIES);
+    ///
```
alxiord, line 291: nit: remove blank line

src/ioctls/system.rs
andreeaflorescu: use Cap is firstly used in the comments here. Future references to Cap will not include this line. Same happens fro use Kvm. But I can also comment both.

src/ioctls/system.rs
andreeaflorescu: get_run_size is named that way because of the name of the ioctl: KVM_GET_VCPU_MMAP_SIZE. All ioctl methods are named by removing the KVM_ prefix and converting it to lower case.

alxiord (dismissed):

src/ioctls/system.rs
andreeaflorescu: Will do, waiting for more comments before doing a push.

src/ioctls/vm.rs
```diff
@@ -247,7 +247,7 @@ impl VmFd {
     }
 
     /// Get the `kvm_run` size.
-    pub fn get_run_size(&self) -> usize {
+    pub fn run_size(&self) -> usize {
```
sameo, line 550: This change should be a separate commit. Maybe even a separate PR.

sameo (commented):
Please split the `run_size()` change into at least its own commit.
Also, one of the commit is a merge point, and we should avoid that.

src/cap.rs
```diff
@@ -9,7 +9,11 @@ use kvm_bindings::*;
 
 /// Enumeration of the extension capability list that KVM exposes.
```
dianpopa: Since cargo doc will put these under "Enums" category, I would remove "Enumeration of" and I would leave only "The list of capabilities exposed by KVM" or something of the kind.

src/cap.rs
```diff
@@ -9,7 +9,11 @@ use kvm_bindings::*;
...
 ///
-/// For details check the
+/// The capabilities list can be used in conjunction with
+/// [Kvm::check_extension()](struct.Kvm.html#method.check_extension) to check if a particular
+/// capability is available.
+///
+/// The list of capabilities is based on the the KVM_CAP_* defines from the
 /// [Linux KVM header](https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/kvm.h).
```
dianpopa, line 17: Could we do the same thing (put a link to the elixir sources) for the `VpuExit` documentation. Currently it looks like this : `The exit reasons are mapped to the KVM_EXIT_* defines from include/uapi/linux/kvm.h.` and it looks kind of ugly when doing `cargo doc --open`.

src/ioctls/system.rs
```diff
@@ -25,6 +25,13 @@ pub struct Kvm {
 impl Kvm {
     /// Opens `/dev/kvm/` and returns a `Kvm` object on success.
```
dianpopa:
```suggestion
    /// Opens `/dev/kvm` and returns a `Kvm` object on success.
```

src/ioctls/system.rs
```diff
@@ -48,10 +57,25 @@ impl Kvm {
 
     /// Opens `/dev/kvm` and returns the fd number on success.
     ///
+    /// One usecase for this method is opening `/dev/kvm` before exec-ing into a
+    /// process with seccomp filters enabled that blacklist the `sys_open` syscall.
+    /// For this usecase `open_with_cloexec` must be called with the`close_on_exec`
```
dianpopa:
```suggestion
    /// For this usecase `open_with_cloexec` must be called with the `close_on_exec`
```

dianpopa (commented):
Additional comments related to documentation:
* `set_user_memory_region` does not have arguments section.
* `create_pit2` does not have arguments section

src/ioctls/vm.rs
andreeaflorescu: I'll add it to a different commit (I was lazy).

The merge commit will vanish once we merge the PR to master. That's how we also do it in Firecracker to avoid `push -f` when the only thing you want to do is update the branch to master.

Trying to find an official documentation that explains this but totally failing.

sameo: > avoid push -f when the only thing you want to do is update the branch to master.

I figured that much. I personally don't mind push forcing to a development branch in order to keep the reviewed commit history clean.

alxiord (commented):
I ran `cargo doc --open` and have some general observations that don't necessarily pertain to these code changes but are documentation related nonetheless:

* The documentation for [`ioctl_ioc_nr`](https://github.com/rust-vmm/kvm-ioctls/blob/83c98b2724332c312cba7f79a2c1dbf3d089bf24/src/sys_ioctl.rs#L14) differs from other macros. It starts with `Raw macro to declare` whereas the others simply state `Declare...`. Please rephrase it.
```
ioctl_io_nr | Declare an ioctl that transfers no data.
ioctl_ioc_nr | Raw macro to declare a function that returns an ioctl number.
ioctl_ior_nr | Declare an ioctl that reads data.
ioctl_iow_nr | Declare an ioctl that writes data.
ioctl_iowr_nr | Declare an ioctl that reads and writes data.
```
* The documentation for wrappers states either `Wrapper for X` or `A wrapper for X`. I personally like the `Wrapper for` option better. In any case, please adjust the different ones.

```
CpuId | Wrapper for kvm_cpuid2 which has a zero length array at the end. Hides the zero length array behind a bounds check.
DeviceFd | Wrapper over the file descriptor obtained when creating an emulated device in the kernel.
Kvm | A wrapper over the /dev/kvm file.
KvmRunWrapper | A safe wrapper over the kvm_run struct.
VcpuFd | A wrapper around creating and using a kvm related vCPU file descriptor.
VmFd | A wrapper around creating and using a VM.
```

* The documentation for enums also differs: `CpuId` explicitly states that it is a list while `VcpuExit` states `Reasons`. Personal preference - I'd like `CpuId` to say `Capabilities exposed by KVM`. It's under the `Enums` category so it's bound to be a collection of something.

```
Cap | The list of capabilities exposed by KVM.
VcpuExit | Reasons for vCPU exits. The exit reasons are mapped to the KVM_EXIT_* defines from include/uapi/linux/kvm.h.
```

andreeaflorescu:
@aghecenco I will leave the ioctl declaration unchanged because having it inside the kvm-ioctls crate is a temporary state till vmm-sys-utils is merged. Once the crate vmm-sys-utils is merged we will remove the definition from this crate so I don't see a point in investing in the documentation just yet.

On the other points, I will take care of them once I get all the public interface documented. Right now, I only added documentation for cap.rs, system.rs and vm.rs is in progress. I'll ping you when I manage to document all items in the ioctls module so you can re-review it. Sorry if I haven't been upfront about it and wasted your time on reviewing. I will change the name of the PR to RFC.

andreeaflorescu:
I think I am done with the documentation of the public interface. Please re-review @sameo @dianpopa @aghecenco

dianpopa (approved):

sameo (approved):

Issue #14, opened 2019-04-03
CpuId, KvmRunWrapper and Result are no longer exported


sameo:
Commit 95279983 moved them from being exported to being private.

PR #15, opened 2019-04-03
lib: Export CpuId, KvmRunWrapper and Result

sameo:
They were unexported by commit 95279983.

Fixes: #14

Signed-off-by: Samuel Ortiz <sameo@linux.intel.com>

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -24,8 +24,8 @@ pub use ioctls::system::Kvm;
 pub use ioctls::vcpu::{VcpuExit, VcpuFd};
 pub use ioctls::vm::VmFd;
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
-use ioctls::CpuId;
-use ioctls::{KvmRunWrapper, Result};
+pub use ioctls::CpuId;
+pub use ioctls::{KvmRunWrapper, Result};
 
 /// Maximum number of CPUID entries that can be returned by a call to KVM ioctls.
 ///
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -23,6 +23,17 @@ pub use cap::Cap;
 pub use ioctls::system::Kvm;
 pub use ioctls::vcpu::{VcpuExit, VcpuFd};
 pub use ioctls::vm::VmFd;
+
+// The following example is used to verify that our public
```
andreeaflorescu, line 28: The CpuId example can be moved to Cpuid::new(). I think it makes more sense in that context.

I have no ideas yet for `Result`. We might be able to test `KvmRunWrapper` once we add the high level documentation and give a more complex example of using this crate that is actually running some code and spawns the vCPU threads.

I would see this documentation test as a temporary solution and I would also add a TODO to fix it once we have all the documentation in place. What do you say?

sameo: > I have no ideas yet for Result. 

Simply running an example that tries to use the structures would test just that.

```Rust
/// # Example
///
/// ```
/// #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
/// use kvm_ioctls::{KvmRunWrapper, Result};
/// ```

```

sameo: @andreeaflorescu I moved the `CpuId` example to the ioctl module, and kept the `use kvm_ioctls::{KvmRunWrapper, Result};` example. This does test the export of those structures.

andreeaflorescu: My concern was only that the lines testing the imports are not placed somewhere where it makes sense from the crate documentation point of view. What I was suggesting was to remove this example once we have other code examples that use `KvmRunWrapper` and `Result` for something meaningful.

sameo: Issue #17 created to track that.
The PR has also been updated to move the `CpuId` example next to the `new` method.

andreeaflorescu (approved):

sameo:
@sboeuf @rbradford Could you PTAL?

rbradford (approved):

sboeuf (approved):

PR #33, opened 2019-04-26
Aarch64/test run code

rhdrjones:
I still don't know Rust, but copying x86 code and modifying it for AArch64 wasn't too bad... Question: is there any way to share boilerplate test code? Creating VM's and VCPU's are required for each test, so we get a lot of redundant lines. Thanks, drew

### DIFF
--- a/src/ioctls/vcpu.rs
+++ b/src/ioctls/vcpu.rs
@@ -928,6 +928,93 @@ mod tests {
         }
     }
 
+    #[cfg(target_arch = "aarch64")]
+    #[test]
+    fn test_run_code() {
+        use std::io::Write;
+
+        let kvm = Kvm::new().unwrap();
+        let vm = kvm.create_vm().unwrap();
+        #[rustfmt::skip]
+        let code = [
+            0x40, 0x20, 0x80, 0x52, /* mov w0, #0x102 */
+            0x00, 0x01, 0x00, 0xb9, /* str w0, [x8]; test physical memory write */
+            0x81, 0x60, 0x80, 0x52, /* mov w1, #0x304 */
+            0x02, 0x00, 0x80, 0x52, /* mov w2, #0x0 */
+            0x20, 0x01, 0x40, 0xb9, /* ldr w0, [x9]; test MMIO read */
+            0x1f, 0x18, 0x14, 0x71, /* cmp w0, #0x506 */
+            0x20, 0x00, 0x82, 0x1a, /* csel w0, w1, w2, eq */
+            0x20, 0x01, 0x00, 0xb9, /* str w0, [x9]; test MMIO write */
+            0x00, 0x00, 0x00, 0x14, /* b <this address>; shouldn't get here, but if so loop forever */
+        ];
+
+        let mem_size = 0x20000;
+        let load_addr = mmap_anonymous(mem_size);
+        let guest_addr: u64 = 0x10000;
+        let slot: u32 = 0;
+        let mem_region = kvm_userspace_memory_region {
+            slot,
+            guest_phys_addr: guest_addr,
+            memory_size: mem_size as u64,
+            userspace_addr: load_addr as u64,
+            flags: KVM_MEM_LOG_DIRTY_PAGES,
+        };
+        vm.set_user_memory_region(mem_region).unwrap();
+
+        unsafe {
+            // Get a mutable slice of `mem_size` from `load_addr`.
+            // This is safe because we mapped it before.
+            let mut slice = std::slice::from_raw_parts_mut(load_addr, mem_size);
+            slice.write(&code).unwrap();
+        }
+
+        let vcpu_fd = vm.create_vcpu(0).unwrap();
+        let mut kvi = kvm_vcpu_init::default();
+        vm.get_preferred_target(&mut kvi).unwrap();
+        vcpu_fd.vcpu_init(&kvi).unwrap();
+
+        let core_reg_base: u64 = 0x6030_0000_0010_0000;
+        let mmio_addr: u64 = guest_addr + mem_size as u64;
+
+        // Set the PC to the guest address where we loaded the code.
+        vcpu_fd.set_one_reg(core_reg_base + 2 * 32, guest_addr);
+
+        // Set x8 and x9 to the addresses the guest test code needs
+        vcpu_fd.set_one_reg(core_reg_base + 2 * 8, guest_addr + 0x10000);
+        vcpu_fd.set_one_reg(core_reg_base + 2 * 9, mmio_addr);
+
+        loop {
+            match vcpu_fd.run().expect("run failed") {
+                VcpuExit::MmioRead(addr, data) => {
+                    assert_eq!(addr, mmio_addr);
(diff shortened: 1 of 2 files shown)


src/ioctls/vm.rs
```diff
@@ -380,6 +380,125 @@ impl VmFd {
...
+    ///             assert_eq!(dirty_pages, 1);
+    ///             break;
+    ///         }
+    ///         exit_reason => panic!("unexpected exit reason: {:?}", exit_reason),
+    ///     }
+    /// }
+    /// ```
+    ///
+    #[cfg(target_arch = "aarch64")]
+    pub fn get_dirty_log(&self, slot: u32, memory_size: usize) -> Result<Vec<u64>> {
```
andreeaflorescu, line 472: This function looks the same as the x86_64 variant. If it is suppose to work the same on both x86 and aarch64, we can just add the aarch64 to the target_arch macro.

I was living with the impression that the `KVM_GET_DIRTY_LOG` ioctl is only available on x86. Check this issue for details: https://github.com/firecracker-microvm/firecracker/issues/866. I am very surprised to see it work on aarch64 as well.

src/ioctls/vcpu.rs
```diff
@@ -928,6 +928,93 @@ mod tests {
         }
     }
 
+    #[cfg(target_arch = "aarch64")]
+    #[test]
+    fn test_run_code() {
```
andreeaflorescu, line 933: We can create a single test_run_code for both platforms to reduce the code duplication.
I don't know if it is possible, but ideally we would trigger the same exits on both x86 and arm with same values for reads and writes so we can have the same asserts in the exits. This would make the run loop to look the same for both platforms. Then the only differences between the two platforms would be:
- the actual code (we would have a aarch64 code and a x86_64 code)
- the register setup (sregs and regs for x86 and one_reg for arm)

src/ioctls/vm.rs
rhdrjones: > This function looks the same as the x86_64 variant. If it is suppose to work the same on both x86 and aarch64, we can just add the aarch64 to the target_arch macro.

I would be happy to share the code. The problem is, that without introducing wrapper functions, I didn't see a way to share it since the unit tests above them in the '///' are arch specific.

I would like to learn how to share vmm code among different architectures that need different unit tests, and arch-neutral unit test boilerplate code among those different unit tests.

> 
> I was living with the impression that the `KVM_GET_DIRTY_LOG` ioctl is only available on x86. Check this issue for details: [firecracker-microvm/firecracker#866](https://github.com/firecracker-microvm/firecracker/issues/866). I am very surprised to see it work on aarch64 as well.

KVM_GET_DIRTY_LOG is supported by all architectures that support KVM. I see the documentation (Documentation/virtual/kvm/api.txt) only states x86, but that's a doc bug.

src/ioctls/vcpu.rs
rhdrjones: > We can create a single test_run_code for both platforms to reduce the code duplication.

There's a bit of overlap, like creating a VM with a memory region and putting guest code named 'code' (which is of course different) there. But then that's it.

> I don't know if it is possible, but ideally we would trigger the same exits on both x86 and arm with same values for reads and writes so we can have the same asserts in the exits.

Only the MMIO exits can be the same. AArch64 doesn't have PortIO nor Hlt.

> This would make the run loop to look the same for both platforms. Then the only differences between the two platforms would be:
> 
>     * the actual code (we would have a aarch64 code and a x86_64 code)
> 
>     * the register setup (sregs and regs for x86 and one_reg for arm)

Besides these two things and the exits, the vcpu creation is even different. AArch64 has a funny little dance with the preferred target and a vcpu-init must be done too. I really don't think there's any value in trying to share this function between x86 and AArch64. In fact, I would argue the code savings would be nil after adding all the conditions.

andreeaflorescu: Ok, that's fine. Would you like to add the code with some description to the highlevel documentation as well? The highlevel docs and examples are placed in [src/lib.rs](https://github.com/rust-vmm/kvm-ioctls/blob/master/src/lib.rs).

You can add a new code block with #[cfg(target_arch = "aarch64"].

src/ioctls/vm.rs
andreeaflorescu: So it seems I am not very good at explaining. I modified your code a bit so that we do not have two functions with same code.
You can annotate parts of the code as architecture specific using macros. I did this in a commit on top of your existing ones. Check it out here: https://github.com/andreeaflorescu/kvm-ioctls/blob/test_run_code_archc/src/ioctls/vm.rs#L260

Basically, the only things that are platform specific right now are:
- setting up the registers
- initialize the vcpu (only on aarch64)
- and the VM exit where we check the dirty log. 

Btw, isn't it possible to have a dummy code similar like for x86_64 that only triggers the halt VM exit? I am asking because the example code could be further simplified if we do that as we wouldn't need to trigger two different VM exits to check dirty log.

andreeaflorescu: This is the diff between your last commit and mine to help visualize things: https://github.com/rust-vmm/kvm-ioctls/commit/8d0d62b7d7955157567c9676ef2e854a69e1fe24

rhdrjones: > Check it out here: https://github.com/andreeaflorescu/kvm-ioctls/blob/test_run_code_archc/src/ioctls/vm.rs#L260

That looks pretty good, but I'm not sure interleaving x86 and arm code is the most readable. Now that I know how to use the annotations I might try some refactoring and then create exactly one arm sequence and one x86 sequence.

> Btw, isn't it possible to have a dummy code similar like for x86_64 that only triggers the halt VM exit? I am asking because the example code could be further simplified if we do that as we wouldn't need to trigger two different VM exits to check dirty log.

arm doesn't have a halt exit reason, nor anything similar. Since x86 does also exit on mmio we could change the x86 code to use the same exit as arm though.

src/ioctls/vcpu.rs
rhdrjones: > Would you like to add the code with some description to the highlevel documentation as well? 

Sure, I'll take a look at that and also attempt the refactoring of common vm/vcpu setup code to see if we can get a bit more code sharing without introducing too many arch annotations.

src/ioctls/vm.rs
andreeaflorescu: We can also have x86_64 and arm examples separate. It just felt to me that a lot of code is similar so I didn't feel the need to split it further more. It's your choice of course.

We can use MmioWrite for both platforms, I just found the MmmioWrite to be a bit unrelated to what we actually want to test. I was looking for either halt or nop.

andreeaflorescu:
@rhdrjones are you still working on this PR? We need the functionality for get dirty log on aarch64 as it is currently blocking one of the issues we have in Firecracker: https://github.com/firecracker-microvm/firecracker/issues/866

@dianpopa can take over the `get_dirty_log` issue.

dianpopa:
Closing in favor of #62 which adjusts while also completing this PR. Read description for more details.

PR #42, opened 2019-05-31
Add support for KVM_SIGNAL_MSI

sboeuf:
This pull request primary goal is to add the support for the KVM ioctl `KVM_SIGNAL_MSI`.
The reason being to allow consumers of this crate to directly inject MSI messages instead of using legacy pin based interrupts.

The second part of this pull request implements the `Clone` trait for `VmFd` so that anyone can safely clone `VmFd`. The need for cloning is to be able to move the `VmFd` into a closure or a thread.


sboeuf:
@andreeaflorescu @sameo @rbradford I don't see how I can properly test this new ioctl...
The reason `signal_msi()` seems impossible to test is tied to the fact that MSI vectors are not chosen from the HW side. The guest OS (or anything that runs inside the VM) is supposed to allocate the MSI/MSI-X vectors, and usually communicate back through PCI configuration space.
If I simply send a random MSI vector, the call to `signal_msi()` fails.
It seems to me this kind of `ioctls` can only be tested from an integration test perspective, but not directly from the crate itself.
Please let me know what you think!

rbradford (dismissed):
I think testing such ioctls will often present challenges and we need to be tolerant of lowering the coverage where necessary. With such little logic in the handling of this ioctl I do not feel overly concerned that anything might break it.

src/ioctls/vm.rs
```diff
@@ -608,6 +608,15 @@ impl AsRawFd for VmFd {
     }
 }
 
+impl Clone for VmFd {
+    fn clone(&self) -> Self {
```
rbradford: nit: I know there is no TryClone trait available in the stable standard library, but did you consider instead manually implementing `try_clone()` instead as I think the error from underlying `try_clone()` could be recoverable.

src/ioctls/vm.rs
```diff
@@ -192,6 +192,45 @@ impl VmFd {
...
+    /// extern crate kvm_bindings;
+    /// # use kvm_ioctls::{Kvm, VmFd};
+    /// use kvm_bindings::kvm_msi;
+    ///
+    /// let kvm = Kvm::new().unwrap();
+    /// let vm = kvm.create_vm().unwrap();
+    /// let msi = kvm_msi::default();
+    /// #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    /// vm.create_irq_chip().unwrap();
+    /// //vm.signal_msi(msi).unwrap();
```
andreeaflorescu, line 237: Maybe add a comment here to why the signal_msi is commented out?

src/ioctls/vm.rs
```diff
@@ -733,4 +772,18 @@ mod tests {
...
+        target_arch = "x86",
+        target_arch = "x86_64",
+        target_arch = "arm",
+        target_arch = "aarch64"
+    ))]
+    fn test_signal_msi_failure() {
+        let kvm = Kvm::new().unwrap();
+        let vm = kvm.create_vm().unwrap();
+        let msi = kvm_msi::default();
+        assert!(vm.signal_msi(msi).is_err());
```
andreeaflorescu, line 813: Add an explanation to why we expect this to be an error.

src/ioctls/vm.rs
```diff
@@ -608,6 +608,15 @@ impl AsRawFd for VmFd {
     }
 }
 
+impl Clone for VmFd {
```
andreeaflorescu: Conceptually I find it a bit weird to have a clone of a VmFd. What is the usecase for adding this? I am thinking that instead of having clones of VmFd you should be able to work with a reference to VmFd.

src/ioctls/vm.rs
andreeaflorescu: I agree with Rob here. I think it's better to avoid expects in libraries and let the consumer handle the error.

sboeuf: Ok I have worked around the problem by using `Arc<>` from the code consuming the `VmFd`. There's no need for the library to handle the `VmFd` cloning.

src/ioctls/vm.rs
sboeuf: Ok I have worked around the problem by using Arc<> from the code consuming the VmFd. There's no need for the library to handle the VmFd cloning.

src/ioctls/vm.rs
sboeuf: Yes will do.

src/ioctls/vm.rs
sboeuf: Yes will do.

src/ioctls/vm.rs
sboeuf: done

src/ioctls/vm.rs
sboeuf: done

sboeuf:
@andreeaflorescu @rbradford PR updated, let me know what you think?

rbradford:
@sboeuf need to update your coverage amount.

rbradford (dismissed):

sboeuf:
Yeah so I lowered the coverage from 91.3 to 91.2 and now all green!
@andreeaflorescu PTAL

sameo (dismissed):

src/ioctls/vm.rs
```diff
@@ -193,6 +193,53 @@ impl VmFd {
...
+        target_arch = "x86_64",
+        target_arch = "arm",
+        target_arch = "aarch64"
+    ))]
+    pub fn signal_msi(&self, msi: kvm_msi) -> Result<()> {
+        // safe becuase we allocated the struct and we know the kernel will read
+        // exactly the size of the struct
+        let ret = unsafe { ioctl_with_ref(self, KVM_SIGNAL_MSI(), &msi) };
+        if ret != -1 {
+            Ok(())
```
andreeaflorescu: As per KVM API documentation this ioctl:
```Returns: >0 on delivery, 0 if guest blocked the MSI, and -1 on error```

We should keep the same return values in this wrapper as well.

Comment (sboeuf) [reply src/ioctls/vm.rs]: should I create a new error for 0?

Comment (andreeaflorescu) [reply src/ioctls/vm.rs]: I think you should return 0 and let the caller handle that exit code.

Comment (sboeuf) [reply src/ioctls/vm.rs]: Ok sounds good.

Comment (andreeaflorescu) [reply src/ioctls/vm.rs]: P.S. It's worth specifying this in the function documentation as well.

Comment (sboeuf) [reply src/ioctls/vm.rs]: Ok it's updated now, let me know what you think.

File: src/ioctls/vm.rs [hunk 6]
```diff
@@ -193,6 +193,55 @@ impl VmFd {
         }
     }
 
+    /// Directly injects a MSI message as per the `KVM_SIGNAL_MSI` ioctl.
```
andreeaflorescu, line 205: Please split documentation comment in summary and description. The summary should be followed by an empty comment line.

src/ioctls/vm.rs
```diff
@@ -193,6 +193,55 @@ impl VmFd {
...
+    /// As per KVM API documentation this ioctl returns >0 on delivery,
+    /// 0 if guest blocked the MSI, and -1 on error.
```
andreeaflorescu: In our case it doesn't return -1 on error, but std::io::Error which contains the last error number. Since this is what happens to all ioctls, I wouldn't mention the error case. You can say something like: on success it return > 0 when the MSI is successfully delivered and 0 when the guest blocked the MSI.

src/ioctls/vm.rs
sboeuf: done

src/ioctls/vm.rs
sboeuf: done

andreeaflorescu (dismissed):

rbradford (dismissed):

rbradford (approved):

andreeaflorescu (approved):

PR #44, opened 2019-06-03
mark `set_user_memory_region` as unsafe

### LINKED ISSUE #38: set_user_memory_region shouldn't be safe
This crate allows calling `set_user_memory_region` from safe code, but that function takes a raw pointer and length, and has no way to check that that refers to valid memory.

Shouldn't a safe version of that require passing a memory region type from the vm-memory crate?

andreeaflorescu:
This function uses as parameter kvm_userspace_memory_region which is
a wrapper over a raw pointer (userspace_addr) and the size
(memory_size) so cannot know if the pointer refers to a valid memory.

Fixes: https://github.com/rust-vmm/kvm-ioctls/issues/38

### DIFF
--- a/src/ioctls/vcpu.rs
+++ b/src/ioctls/vcpu.rs
@@ -620,7 +620,7 @@ impl VcpuFd {
     ///         userspace_addr: load_addr as u64,
     ///         flags: 0,
     ///     };
-    ///     vm.set_user_memory_region(mem_region).unwrap();
+    ///     unsafe { vm.set_user_memory_region(mem_region).unwrap() };
     ///
     ///     // Dummy x86 code that just calls halt.
     ///     let x86_code = [
@@ -961,7 +961,9 @@ mod tests {
             userspace_addr: load_addr as u64,
             flags: KVM_MEM_LOG_DIRTY_PAGES,
         };
-        vm.set_user_memory_region(mem_region).unwrap();
+        unsafe {
+            vm.set_user_memory_region(mem_region).unwrap();
+        }
 
         unsafe {
             // Get a mutable slice of `mem_size` from `load_addr`.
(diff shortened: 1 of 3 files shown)


acatangiu (approved):

rbradford (approved):

src/ioctls/vcpu.rs
```diff
@@ -620,7 +620,7 @@ impl VcpuFd {
     ///         userspace_addr: load_addr as u64,
     ///         flags: 0,
     ///     };
-    ///     vm.set_user_memory_region(mem_region).unwrap();
+    ///     unsafe { vm.set_user_memory_region(mem_region).unwrap() };
```
alexandruag, line 623: For maximum neatness, the `unwrap` here should've been outside of the `unsafe` block. This occurs in a number of other places as well :-s

src/ioctls/vcpu.rs
```diff
@@ -961,7 +961,9 @@ mod tests {
             userspace_addr: load_addr as u64,
             flags: KVM_MEM_LOG_DIRTY_PAGES,
         };
-        vm.set_user_memory_region(mem_region).unwrap();
+        unsafe {
```
alexandruag, line 964: It's nice to have a comment outside the `unsafe` block explaining why the code is actually safe to use. This seems to be pretty pervasive around here :(

src/ioctls/vm.rs
```diff
@@ -83,15 +83,16 @@ impl VmFd {
...
     ///                     flags: 0,
     ///                 };
-    /// vm.set_user_memory_region(mem_region).unwrap();
+    /// unsafe {
+    ///     vm.set_user_memory_region(mem_region).unwrap();
+    /// };
     /// ```
     ///
-    pub fn set_user_memory_region(
+    pub unsafe fn set_user_memory_region(
```
alexandruag, line 91: Whenever a method is declared as `unsafe`, we should add documentation stating what invariants must hold when calling the function such that the code is actually safe.

src/ioctls/vcpu.rs
andreeaflorescu: Not in the tests though. But if you want to have it in the tests as well, I can add a comment.

PR #49, opened 2019-06-10
Add support for KVM_ENABLE_CAP ioctl

sboeuf:
By adding the support for the ioctl KVM_ENABLE_CAP, this commit will
allow consumers of the kvm-ioctls crate to enable any KVM capability
they would be missing from their VMM implementation.

One example of capability that can be interesting to enable is the
newly added KVM_CAP_SPLIT_IRQCHIP, that allows the consumer to prevent
KVM from creating a virtual IOAPIC and PIC. This capability can let
a VMM provide its own userspace implementation of those components for
security purposes.


src/ioctls/vm.rs
```diff
@@ -812,4 +855,25 @@ mod tests {
         let msi = kvm_msi::default();
         assert!(vm.signal_msi(msi).is_err());
     }
+
+    /// Providing the `kvm_enable_cap` structure filled with default() should
```
andreeaflorescu: Can you move the comments before the assert? I don't think that's a common practice in Rust to write the comments before the test function. I haven't seen it at least.

src/ioctls/vm.rs
```diff
@@ -812,4 +855,25 @@ mod tests {
...
+    fn test_enable_cap_failure() {
+        let kvm = Kvm::new().unwrap();
+        let vm = kvm.create_vm().unwrap();
+        let cap: kvm_enable_cap = Default::default();
+        assert!(vm.enable_cap(&cap).is_err());
+    }
+
+    #[test]
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64",))]
+    fn test_enable_split_irqchip_cap() {
```
andreeaflorescu, line 885: Please add a comment and explain that cap.args[0] is the number of routes reserved for userspace IOAPICs or add a link to the full explanation: https://elixir.bootlin.com/linux/latest/source/Documentation/virtual/kvm/api.txt#L4606

It took me some time to understand what this does.

src/ioctls/vm.rs
```diff
@@ -610,6 +610,49 @@ impl VmFd {
...
+    /// * kvm_enable_cap - KVM capability structure. For details check the `kvm_enable_cap`
+    ///                    structure in the
+    ///                    [KVM API doc](https://www.kernel.org/doc/Documentation/virtual/kvm/api.txt).
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # extern crate kvm_ioctls;
+    /// extern crate kvm_bindings;
+    /// #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
```
andreeaflorescu: I am not a big fan of having x86_64 specific examples when the ioctl is available on other platforms as well. I would suggest to use `KVM_CAP_ARM_USER_IRQ` on aarch64 and `KVM_CAP_SPLIT_IRQCHIP` on x86_64. The only code that should be specific to the platform should be ```cap.cap = ...``` and ```cap.args[0] = 24```. Also add a comment about what is 24.

src/ioctls/vm.rs
```diff
@@ -610,6 +610,49 @@ impl VmFd {
         Ok(())
     }
 
+    /// Enable the specified capability as per the `KVM_ENABLE_CAP` ioctl.
+    ///
+    /// See the documentation for `KVM_ENABLE_CAP`.
+    ///
+    /// This ioctl returns Ok() when the capability is successfully enabled,
```
andreeaflorescu: I don't think you need the comment about the return type so we can keep the same comment style on all methods. Methods in this crate only specify the return type when it's not Result<()>.

sboeuf: Fixed.

src/ioctls/vm.rs
sboeuf: Done

src/ioctls/vm.rs
sboeuf: done

src/ioctls/vm.rs
sboeuf: Yes you're right, it's actually good to show an example with two different architectures.

sboeuf: Well let me take that back... The cap `KVM_CAP_ARM_USER_IRQ` that you suggested is only meant to be checked through `check_extension()`, but it cannot be set from the VMM.
And based on the KVM bindings 4.20, there is no `arm/aarch64` capability that we could take as an example.
I reworked the structure of the documentation test though, let me know if it's okay with you.

andreeaflorescu: Yeah, I didn't look in depth. Is there really no other cap that we can enable for aarch64?

sboeuf: Unless I'm reading KVM doc the wrong way (could be possible :p), I don't think there are some ARM caps that we can simply try to enable.
The only one could be `KVM_CAP_MANUAL_DIRTY_LOG_PROTECT2` but it's not part of our bindings because it's been introduced recently on the kernel.

andreeaflorescu: `KVM_CAP_MANUAL_DIRTY_LOG_PROTECT2` is only available starting with 5.0 so it will fail anyway since we don't have that capability on the host kernel.

src/cap.rs
```diff
@@ -23,6 +23,8 @@ use kvm_bindings::*;
 #[allow(missing_docs)]
 pub enum Cap {
     Irqchip = KVM_CAP_IRQCHIP,
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    SplitIrqchip = KVM_CAP_SPLIT_IRQCHIP,
```
alxiord: The capabilities are sorted by value. Please move this lower (should be between `KVM_CAP_S390_USER_SIGP` and `KVM_CAP_IMMEDIATE_EXIT`).

src/ioctls/vm.rs
```diff
@@ -610,6 +610,62 @@ impl VmFd {
...
+    ///     // provided through the first argument of the capability structure, as
+    ///     // specified in KVM documentation:
+    ///     //     args[0] - number of routes reserved for userspace IOAPICs
+    ///     //
+    ///     // Because an IOAPIC supports 24 pins, that's the reason why this test
+    ///     // picked this number as reference.
+    ///     cap.args[0] = 24;
+    ///     vm.enable_cap(&cap).unwrap();
+    /// }
+    ///
```
alxiord: nit: remove blank line

src/ioctls/vm.rs
```diff
@@ -610,6 +610,62 @@ impl VmFd {
...
+    /// # Arguments
+    ///
+    /// * kvm_enable_cap - KVM capability structure. For details check the `kvm_enable_cap`
+    ///                    structure in the
+    ///                    [KVM API doc](https://www.kernel.org/doc/Documentation/virtual/kvm/api.txt).
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # extern crate kvm_ioctls;
```
alxiord, line 628: Please remove his line (and the next one prefixed by `#`). They don't show up in the generated doc and the doctest passes without them.

src/ioctls/vm.rs
```diff
@@ -610,6 +610,62 @@ impl VmFd {
...
+    ///     //
+    ///     // Because an IOAPIC supports 24 pins, that's the reason why this test
+    ///     // picked this number as reference.
+    ///     cap.args[0] = 24;
+    ///     vm.enable_cap(&cap).unwrap();
+    /// }
+    ///
+    /// ```
+    ///
+    pub fn enable_cap(&self, cap: &kvm_enable_cap) -> Result<()> {
```
alxiord, line 658: What architectures is this available on? The [documentation](https://www.kernel.org/doc/Documentation/virtual/kvm/api.txt) only mentions `mips, ppc, s390`, I found patches for x86, but what about aarch64?

sboeuf: The documentation states that all architectures support this ioctl:
```
4.37 KVM_ENABLE_CAP

Capability: KVM_CAP_ENABLE_CAP
Architectures: mips, ppc, s390
Type: vcpu ioctl
Parameters: struct kvm_enable_cap (in)
Returns: 0 on success; -1 on error

Capability: KVM_CAP_ENABLE_CAP_VM
Architectures: all
Type: vcpu ioctl
Parameters: struct kvm_enable_cap (in)
Returns: 0 on success; -1 on error

+Not all extensions are enabled by default. Using this ioctl the application
can enable an extension, making it available to the guest.

On systems that do not support this ioctl, it always fails. On systems that
do support it, it only works for extensions that are supported for enablement.

To check if a capability can be enabled, the KVM_CHECK_EXTENSION ioctl should
be used.

struct kvm_enable_cap {
       /* in */
       __u32 cap;

The capability that is supposed to get enabled.

       __u32 flags;

A bitfield indicating future enhancements. Has to be 0 for now.

       __u64 args[4];

Arguments for enabling a feature. If a feature needs initial values to
function properly, this is the place to put them.

       __u8  pad[64];
};

The vcpu ioctl should be used for vcpu-specific capabilities, the vm ioctl
for vm-wide capabilities.

```

src/cap.rs
sboeuf: Fixed

src/ioctls/vm.rs
sboeuf: Fixed

src/ioctls/vm.rs
sboeuf: Fixed but those lines (prefixed by #) where there in every other doctest.

sboeuf:
@aghecenco comments addressed PTAL

src/ioctls/vm.rs
sboeuf: I can't remove it, the CI complains:
```

---- src/ioctls/vm.rs - ioctls::vm::VmFd::enable_cap (line 627) stdout ----
error[E0433]: failed to resolve: use of undeclared type or module `Kvm`
--> src/ioctls/vm.rs:632:11
 |
7 | let kvm = Kvm::new().unwrap();
 | ^^^ use of undeclared type or module `Kvm`
```

src/ioctls/vm.rs
sboeuf: Oh my bad! Talking with @andreeaflorescu I realized this snippet of the doc is not part of 4.14 or 4.20, it's only been introduced later in 5.x.
So I suggest that for now I can remove the arm/aarch64 target.

andreeaflorescu: Sounds good to me.

sboeuf: Ok I've updated the PR!

andreeaflorescu (dismissed):

sboeuf:
@andreeaflorescu @sameo @acatangiu @aghecenco please take a look.
The PR is ready, and @andreeaflorescu approved it, but the approval went away because I had to rebase it on `master` and repush...

andreeaflorescu:
@sboeuf you don't need to do push forces for updating to master. You can use the update branch button. The merge commit is not kept when merging the PR.

sboeuf:
@andreeaflorescu oh okay good to know!

sameo (approved):

andreeaflorescu (approved):

PR #51, opened 2019-06-12
vcpu: Return proper vector when receiving EOI

sboeuf:
The vector value returned along with End Of Interrupt (EOI) VM exit
is important as it indicates a userspace IOAPIC which interrupt needs
to be deasserted.

About KVM_EXIT_IOAPIC_EOI, KVM documentation mentions:
```
/* KVM_EXIT_IOAPIC_EOI */
struct {
    __u8 vector;
} eoi;

Indicates that the VCPU's in-kernel local APIC received an EOI for a
level-triggered IOAPIC interrupt. This exit only triggers when the
IOAPIC is implemented in userspace (i.e. KVM_CAP_SPLIT_IRQCHIP is
enabled); the userspace IOAPIC should process the EOI and retrigger
the interrupt if it is still asserted. Vector is the LAPIC interrupt
vector for which the EOI was received.
```


src/ioctls/vcpu.rs
```diff
@@ -725,7 +725,13 @@ impl VcpuFd {
                 KVM_EXIT_EPR => Ok(VcpuExit::Epr),
                 KVM_EXIT_SYSTEM_EVENT => Ok(VcpuExit::SystemEvent),
                 KVM_EXIT_S390_STSI => Ok(VcpuExit::S390Stsi),
-                KVM_EXIT_IOAPIC_EOI => Ok(VcpuExit::IoapicEoi),
+                KVM_EXIT_IOAPIC_EOI => {
+                    // Safe because the exit_reason (which comes from the kernel) told us which
+                    // union field to use.
+                    let eoi = unsafe { &mut run.__bindgen_anon_1.eoi };
+
```
alxiord: nit: remove blank line

tests/coverage
```diff
@@ -1 +1 @@
-91.3
+90.9
```
alxiord, line 1: I'm not too happy about this, is there no way this snippet can be properly covered?

andreeaflorescu: I also find it a bit weird that the coverage dropped with so many percentages.

sboeuf: Yeah I don't understand why it dropped so much given the minimal amount of code added

sboeuf: But I don't see how we could force an VM exit EOI to happen...

src/ioctls/vcpu.rs
sboeuf: done

tests/coverage
sboeuf: @bonzini @sameo @rbradford any idea if it would be possible to trigger an EOI for the purpose of unit testing.
This looks very complex to me, and it's worth dropping the coverage IMO.

sboeuf: A bit more feedback. The Intel SDM Volume 3A, Section 10.8.5 specifies how the software can generate an EOI by writing to the EOI register `0xFEE000B0`. That's why I have tried to add one line of assembly code to the existing test `test_run_code`. Also, I had to make sure to enable `KVM_CAP_SPLIT_IRQCHIP` so that the KVM local APIC would redirect the EOI by generating a VM exit EOI.
Here is the assembly line
```rust
0xc7, 0x04, 0x01, 0xb0, 0x00, 0xe0, 0xfe,  // movl   $0x2a, (0xfee000b0);
```
But unfortunately it does not work and I don't receive any VM exit EOI from the unit test.

I think this is simply too complex. From my understanding, we would need a full VMM to try this out (not a simple test like here), with an OS doing a bunch of things. I mean to say we need to advertise there is an IOAPIC taking care of pins 0-23, then the guest should read that from the mptables, and then based on what device is probed (like a serial device needing IRQ 4), the guest OS takes care of assigning some interrupt (which means configuring the local APIC).
Last part is about the need for the local APIC to be requested for an interrupt, so that later it will accept to forward the EOI to the userspace IOAPIC (translated into a VM exit in the case of KVM).

@bonzini because you are the KVM expert/maintainer, could you correct anything wrong in this statement, and maybe confirm/deny if having a unit test for that is feasible/reasonable?

sameo: @andreeaflorescu @aghecenco I think that unit testing an EOI would take a lot more than the code snippet in `test_run_code()`, unfortunately. Unless someone comes up with an easy way for KVM to forward an EOI to us, and given that this is a relatively trivial piece of code, we should accept that 0.4% coverage decrease.

alxiord: @sboeuf @sameo I agree. It's overkill. I'm OK with the coverage drop. Cool deep dive!

alxiord (approved):

tests/coverage
andreeaflorescu: Anyone checked why it's dropping so much by running kcov before and after this patch? I agree with decreasing the coverage, I just want to make sure we don't have another bug hidden somewhere in our tests. A small change shouldn't have such a big impact on the coverage.

sboeuf:
Talking with @andreeaflorescu, we agreed we can wait a few more days for @bonzini input as we want to make sure there's actually no way to test this out properly.
Based on input, or no input at all, we'll end up merging.

sboeuf:
@andreeaflorescu @sameo no feedback so far. Could we merge?
Also, I wanted to mention I didn't add this VM exit support, I simply didn't ignore a returned parameter, but the VM exit itself was not tested before this patch.

I think if someone has anything to add to this topic later, we can still raise another PR to fix those tests (but I doubt it'll be the case).

andreeaflorescu (approved):

PR #52, opened 2019-06-24
implement Drop for KvmRunWrapper

andreeaflorescu:
In KvmRunWrapper we are mmaping memory for the kvm_run wrapper.
The memory needs to be unmapped before dropping the reference to
KvmRunWrapper.


src/ioctls/mod.rs
```diff
@@ -223,6 +223,7 @@ impl CpuId {
 /// threads as raw pointers do not implement `Send` and `Sync`.
 pub struct KvmRunWrapper {
     kvm_run_ptr: *mut u8,
+    size: usize,
```
alxiord: nit: I'd call it something other than `size` (`mapping_size`?), because it's not obvious to someone who doesn't know/care about the underlying `mmap` that we need to know the mapping size for later.

acatangiu: I agree, otherwise LGTM 👍

andreeaflorescu: Done.

acatangiu (approved):

alxiord (approved):

PR #56, opened 2019-07-09
Add support for KVM_SET_GSI_ROUTING ioctl

sboeuf:
By adding the support for the ioctl KVM_SET_GSI_ROUTING, this commit
will allow consumers of the kvm-ioctls crate to set custom mappings for
GSI entries.

This would let a caller of this ioctl defines that GSI number X can
trigger an interrupt by delivering an MSI directly to the local APIC or
instead deliver a pin based interrupt to the in-kernel IOAPIC.

One example that comes to mind is to use this ioctl to let a VMM
interact with the VFIO API, as it requires an eventfd to be associated
with each interrupt. By combining this ioctl with KVM_IRQFD, one could
directly map an interrupt to an eventfd managed by KVM, and associated
to a specific GSI.

### DIFF
--- a/src/ioctls/vm.rs
+++ b/src/ioctls/vm.rs
@@ -254,6 +254,54 @@ impl VmFd {
         }
     }
 
+    /// Sets the GSI routing table entries, overwriting any previously set
+    /// entries, as per the `KVM_SET_GSI_ROUTING` ioctl.
+    ///
+    /// See the documentation for `KVM_SET_GSI_ROUTING`.
+    ///
+    /// Returns an io::Error when the table could not be updated.
+    ///
+    /// # Arguments
+    ///
+    /// * kvm_irq_routing - IRQ routing configuration. Describe all routes
+    ///                     associated with GSI entries. For details check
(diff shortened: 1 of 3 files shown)


src/ioctls/vm.rs
```diff
@@ -254,6 +254,53 @@ impl VmFd {
...
+    ///
+    /// See the documentation for `KVM_SET_GSI_ROUTING`.
+    ///
+    /// Returns an io::Error when the table could not be updated.
+    ///
+    /// # Arguments
+    ///
+    /// * kvm_irq_routing - IRQ routing configuration. Describe all routes
+    ///                     associated with GSI entries. For details check
+    ///                     the `kvm_msi` structure in the
```
alxiord:
```suggestion
    ///                     the `kvm_irq_routing` and `kvm_irq_routing_entry` structures in the
```

sboeuf: Fixed! thanks

alxiord (approved):

sboeuf:
@andreeaflorescu @sameo ptal

sameo (approved):

PR #57, opened 2019-08-10
Add unregister_irqfd() to VmFd

jiangliu:
Add unregister_irqfd() to VmFd to unregister an irqfd from the KVM,
which will be needed when enabling supoort of MSI.

Signed-off-by: Liu Jiang <gerry@linux.alibaba.com>

### DIFF
--- a/src/ioctls/vm.rs
+++ b/src/ioctls/vm.rs
@@ -535,6 +535,52 @@ impl VmFd {
         }
     }
 
+    /// Unregisters an event that will, when signaled, trigger the `gsi` IRQ.
+    ///
+    /// # Arguments
+    ///
+    /// * `fd` - Event to be signaled.
+    /// * `gsi` - IRQ to be triggered.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # extern crate kvm_ioctls;
+    /// # extern crate libc;
+    /// # use kvm_ioctls::{Kvm, VmFd};
+    /// # use libc::{eventfd, EFD_NONBLOCK};
+    /// let kvm = Kvm::new().unwrap();
+    /// let vm = kvm.create_vm().unwrap();
+    /// let evtfd = unsafe { eventfd(0, EFD_NONBLOCK) };
+    /// #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    /// vm.register_irqfd(evtfd, 0).unwrap();
+    /// #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    /// vm.unregister_irqfd(evtfd, 0).unwrap();
+    /// ```
+    ///
+    #[cfg(any(
+        target_arch = "x86",
+        target_arch = "x86_64",
+        target_arch = "arm",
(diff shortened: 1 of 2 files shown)


src/ioctls/vm.rs
```diff
@@ -844,6 +890,42 @@ mod tests {
...
+        // On x86_64 this fails as the event fd was already matched with a GSI.
+        assert!(vm_fd.register_irqfd(evtfd3, 4).is_err());
+        assert!(vm_fd.register_irqfd(evtfd3, 5).is_err());
+
+        // KVM irqfd doesn't report failure on this case:(
+        assert!(vm_fd.unregister_irqfd(evtfd3, 5).is_ok());
+
+        let evtfd4 = unsafe { eventfd(0, EFD_NONBLOCK) };
+        unsafe { libc::close(evtfd4) };
+        assert!(vm_fd.unregister_irqfd(evtfd4, 5).is_err());
```
acatangiu: This test is racy.
Might not occur now, but as we add more tests that deal with `fd`s we will hit the case where another test reuses the `fd` closed in the line above and then this check will sporadically fail.

I suggest you remove this part of the test, or test it with an always invalid `fd` like `-1`.

acatangiu (changes requested):
Functionality looks good, small change requested in a test.

Also, please fix typo in commit message: `s/supoort/suport`

src/ioctls/vm.rs
sameo: Nice catch, thanks.

sameo (dismissed):
Once @acatangiu comment is addressed, this could be merged.

src/ioctls/vm.rs
jiangliu: Thanks for review:)

src/ioctls/vm.rs
```diff
@@ -844,6 +890,39 @@ mod tests {
...
+            assert!(vm_fd.unregister_irqfd(evtfd2, 8).is_ok());
+        }
+
+        // On aarch64, this fails because setting up the interrupt controller is mandatory before
+        // registering any IRQ.
+        // On x86_64 this fails as the event fd was already matched with a GSI.
+        assert!(vm_fd.register_irqfd(evtfd3, 4).is_err());
+        assert!(vm_fd.register_irqfd(evtfd3, 5).is_err());
+
+        // KVM irqfd doesn't report failure on this case:(
```
sameo, line 922: Maybe we should specify what *this case* is: Duplicated eventfd registration.

src/ioctls/vm.rs
```diff
@@ -844,6 +890,39 @@ mod tests {
...
+        assert!(vm_fd.unregister_irqfd(evtfd3, 5).is_ok());
+        assert!(vm_fd.unregister_irqfd(-1, 5).is_err());
```
sameo, line 926: Could we separate (by a new line or a comment) this one from the previous one. It feels like the previous comment also applies to that one. Sorry for the nits...

acatangiu (approved):

alxiord (approved):

PR #58, opened 2019-08-14
reimplement CpuId as a FamStructWrapper

serban300:
The CpuId type was implemented as a wrapper over the kvm_cpuid2
structure. Since kvm_cpuid2 contains a flexible array member,
we can reimplement CpuId as a FamStructWrapper<kvm_cpuid2>.

Signed-off-by: Serban Iorga <seriorga@amazon.com>

### DIFF
--- a/src/ioctls/mod.rs
+++ b/src/ioctls/mod.rs
@@ -15,6 +15,8 @@ use std::result;
 use kvm_bindings::kvm_run;
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
 use kvm_bindings::{kvm_cpuid2, kvm_cpuid_entry2};
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+use vmm_sys_util::fam::{FamStruct, FamStructWrapper};
 
 /// Wrappers over KVM device ioctls.
 pub mod device;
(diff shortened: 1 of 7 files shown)


andreeaflorescu (changes requested):
As we discussed offline, we are going to block the merge of this PR until we publish vmm-sys-utils to crates.io so we don't have to consume it from git.

andreeaflorescu:
@serban300 vmm-sys-util is now published on crates.io https://crates.io/crates/vmm-sys-util

src/ioctls/mod.rs
```diff
@@ -15,6 +15,10 @@ use std::result;
 use kvm_bindings::kvm_run;
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
 use kvm_bindings::{kvm_cpuid2, kvm_cpuid_entry2};
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+use vmm_sys_util::fam::{FamStruct, FamStructWrapper};
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
```
andreeaflorescu: I think it is better to move the definition of MAX_KVM_CPUID_ENTRIES here instead of importing it from `src/lib.rs`.

src/ioctls/mod.rs
```diff
@@ -64,159 +68,68 @@ fn vec_with_array_field<T: Default, F>(count: usize) -> Vec<T> {
...
+        let KvmCpuId(cpuid) = self;
+        KvmCpuId(kvm_cpuid2 {
+            nent: cpuid.nent,
+            padding: cpuid.padding,
+            entries: cpuid.entries.clone(),
+        })
     }
 }
 
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
```
jiangliu, line 96: Could we use "generate_fam_struct_impl(KvmCpuId, kvm_cpuid_entry2, u32, 0.nent, MAX_KVM_CPUID_ENTRIES)" here to reduce code size?

src/ioctls/mod.rs
```diff
@@ -15,6 +15,10 @@ use std::result;
 use kvm_bindings::kvm_run;
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
 use kvm_bindings::{kvm_cpuid2, kvm_cpuid_entry2};
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+use vmm_sys_util::fam::{FamStruct, FamStructWrapper};
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+use MAX_KVM_CPUID_ENTRIES;
```
jiangliu: how about "use crate::MAX_KVM_CPUID_ENTRIES;" for better readability?

src/ioctls/mod.rs
serban300: It would be really nice, and I tried that, but `0.nent` can't be used in the macro call.

src/ioctls/mod.rs
serban300: Sure. Sounds good.

src/ioctls/mod.rs
serban300: Makes sense. Done.

andreeaflorescu (dismissed):

sameo (dismissed):

src/lib.rs
```diff
@@ -193,7 +194,8 @@ pub use ioctls::system::Kvm;
 pub use ioctls::vcpu::{VcpuExit, VcpuFd};
 pub use ioctls::vm::{IoEventAddress, NoDatamatch, VmFd};
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
-pub use ioctls::CpuId;
+pub use ioctls::{CpuId, MAX_KVM_CPUID_ENTRIES};
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
```
acatangiu: This is added by mistake I think.

serban300: Yes. Fixed.

acatangiu (approved):

sameo (approved):

PR #61, opened 2019-10-11
Implement FromRawFd for DeviceFd

jiangliu:
Implement FromRawFd for DeviceFd to recreate a DeviceFd object from a
raw fd.
Also export VmFd::new_vmfd() as pub.

Signed-off-by: Liu Jiang <gerry@linux.alibaba.com>

### DIFF
--- a/coverage_config.json
+++ b/coverage_config.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 91.0,
+  "coverage_score": 90.3,
   "exclude_path": "",
   "crate_features": ""
 }
--- a/src/ioctls/device.rs
+++ b/src/ioctls/device.rs
@@ -3,7 +3,7 @@
 
 use std::fs::File;
 use std::io;
-use std::os::unix::io::{AsRawFd, RawFd};
+use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
 
 use kvm_bindings::kvm_device_attr;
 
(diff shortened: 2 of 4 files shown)


src/lib.rs
```diff
@@ -192,7 +192,7 @@ pub use cap::Cap;
 pub use ioctls::device::DeviceFd;
 pub use ioctls::system::Kvm;
 pub use ioctls::vcpu::{VcpuExit, VcpuFd};
-pub use ioctls::vm::{IoEventAddress, NoDatamatch, VmFd};
+pub use ioctls::vm::{new_vmfd, IoEventAddress, NoDatamatch, VmFd};
```
andreeaflorescu: Hmmm, the `new_vmfd` was not intended to be exported as part of the public interface. Checkout the comment: https://github.com/rust-vmm/kvm-ioctls/blob/8ead44e3af3f803d3a1c228755db54fc004aa20d/src/ioctls/vm.rs#L769

Any particular reason you need this function to be public?

jiangliu: We want to restore the VmFd object on live upgrade destination side, otherwise we need to expose struct VmFd's private fields.

andreeaflorescu: This change should be added to a new commit in any case.

If you want to have a public constructor for VmFd, than you should add a public constructor in the VmFd structure implementation. When I first wrote this code this is what I was trying to avoid, hence I couldn't have a public constructor because VmFd would be a public structure hence the constructor will also be public. This is why I wrote the helper function `new_vmfd` so that the function can be public in the module, but not exported in the public interface.

jiangliu: Hi Andreea, I have reimplemented it in another way:)

acatangiu (dismissed):

andreeaflorescu (changes requested):
Please update the commit message:
- summary should be less than 50 chars
- please describe in the commit description **why** you are doing this change. **what** change you're doing is obvious from the changed lines, the why is not super obvious at a first look.

acatangiu (approved):

sameo (approved):

andreeaflorescu (approved):

PR #62, opened 2019-10-11
Support for getting dirty logs on aarch64

dianpopa:
# Description of changes
This PR completes while also adjusting #33:
* `get_dirty_log` function is not `cfg` labeled anymore. The cfg labels were moved inside the doc comments for grouping architecturally specific variables and the examples was brought to a state that has as many similar line of codes between different architectures
* `test_run_code` was left as it was
* added main doc in `src/lib.rs` for aarch64 also

### DIFF
--- a/src/ioctls/vm.rs
+++ b/src/ioctls/vm.rs
@@ -87,8 +87,8 @@ impl VmFd {
     /// let vm = kvm.create_vm().unwrap();
     /// let mem_region = kvm_userspace_memory_region {
     ///                     slot: 0,
-    ///                     guest_phys_addr: 0x1000 as u64,
-    ///                     memory_size: 0x4000 as u64,
+    ///                     guest_phys_addr: 0x10000 as u64,
+    ///                     memory_size: 0x10000 as u64,
     ///                     userspace_addr: 0x0 as u64,
     ///                     flags: 0,
     ///                 };
--- a/src/ioctls/vm.rs
+++ b/src/ioctls/vm.rs
@@ -7,7 +7,6 @@
 
 use kvm_bindings::*;
 use std::fs::File;
-#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
 use std::os::raw::c_void;
 use std::os::raw::{c_int, c_ulong};
 use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
@@ -697,7 +696,7 @@ impl VmFd {
     /// # use kvm_bindings::{kvm_userspace_memory_region, KVM_MEM_LOG_DIRTY_PAGES};
     /// # let kvm = Kvm::new().unwrap();
     /// # let vm = kvm.create_vm().unwrap();
-    /// // This examples is based on https://lwn.net/Articles/658511/.
+    /// // This example is based on https://lwn.net/Articles/658511/.
     /// let mem_size = 0x4000;
     /// let guest_addr: u64 = 0x1000;
     /// let load_addr: *mut u8 = unsafe {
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -93,39 +122,43 @@
...
+//!         vcpu_regs.rip = guest_addr;
+//!         vcpu_regs.rax = 2;
+//!         vcpu_regs.rbx = 3;
+//!         vcpu_regs.rflags = 2;
+//!         vcpu_fd.set_regs(&vcpu_regs).unwrap();
+//!     }
+//!
+//!     #[cfg(target_arch = "aarch64")]
+//!     {
+//!         // aarch64 specific registry setup.
```
alxiord, line 150: nit: add a similar comment to the `x86_64` snippet

src/lib.rs
```diff
@@ -155,26 +188,23 @@
...
-//!             VcpuExit::Hlt => {
 //!                 // The code snippet dirties 1 page when it is loaded in memory
 //!                 let dirty_pages_bitmap = vm.get_dirty_log(slot, mem_size).unwrap();
 //!                 let dirty_pages = dirty_pages_bitmap
 //!                     .into_iter()
 //!                     .map(|page| page.count_ones())
 //!                     .fold(0, |dirty_page_count, i| dirty_page_count + i);
 //!                 assert_eq!(dirty_pages, 1);
+//!                 #[cfg(target_arch = "aarch64")]
+//!                 break;
```
alxiord, line 200: nit: explain in a comment why the `break` occurs only on `aarch64` (I don't understand it)

src/lib.rs
```diff
@@ -61,15 +59,46 @@
...
 //!     use kvm_bindings::kvm_userspace_memory_region;
 //!
+//!	let mut mem_size = 0x0;
+//!	let mut guest_addr = 0x0;
+//!	let asm_code: &[u8];
+//!
+//!     // Setting up architectural dependent values.
+//!     #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+//!    {
+//!        // This examples is based on https://lwn.net/Articles/658511/.
```
alxiord:
```suggestion
//!        // This example is based on https://lwn.net/Articles/658511/.
```

src/ioctls/vm.rs
```diff
@@ -394,9 +393,22 @@ impl VmFd {
...
     /// # let vm = kvm.create_vm().unwrap();
-    /// // This examples is based on https://lwn.net/Articles/658511/.
-    /// let mem_size = 0x4000;
-    /// let guest_addr: u64 = 0x1000;
+    /// let mut mem_size = 0x0;
+    /// let mut guest_addr = 0x0;
+    ///
+    /// #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    /// {
+    ///     // This examples is based on https://lwn.net/Articles/658511/.
```
alxiord:
```suggestion
    ///     // This example is based on https://lwn.net/Articles/658511/.
```

src/ioctls/vm.rs
```diff
@@ -394,9 +393,22 @@ impl VmFd {
     /// # use kvm_bindings::{kvm_userspace_memory_region, KVM_MEM_LOG_DIRTY_PAGES};
     /// # let kvm = Kvm::new().unwrap();
     /// # let vm = kvm.create_vm().unwrap();
-    /// // This examples is based on https://lwn.net/Articles/658511/.
-    /// let mem_size = 0x4000;
-    /// let guest_addr: u64 = 0x1000;
+    /// let mut mem_size = 0x0;
+    /// let mut guest_addr = 0x0;
```
alxiord: Comment out these lines too so they don't appear in the generated documentation - they're not relevant to the example

src/lib.rs
```diff
@@ -61,15 +59,46 @@
...
+//!        ];
+//!    }
+//!    #[cfg(target_arch = "aarch64")]
+//!    {
+//!        mem_size = 0x10000;
+//!        guest_addr = 0x10000;
+//!        asm_code = &[
+//!             0x01, 0x00, 0x00, 0x10, /* adr x1, <this address> */
+//!             0x22, 0x10, 0x00, 0xb9, /* str w2, [x1, #16]; write to this page */
+//!             0x02, 0x00, 0x00, 0xb9, /* str w2, [x0]; force MMIO exit */
```
alxiord: Is it a read or a write?

src/lib.rs
andreeaflorescu: On aarch64 you don't have a halt instruction :(

src/ioctls/vm.rs
```diff
@@ -394,9 +393,22 @@ impl VmFd {
     /// # use kvm_bindings::{kvm_userspace_memory_region, KVM_MEM_LOG_DIRTY_PAGES};
     /// # let kvm = Kvm::new().unwrap();
     /// # let vm = kvm.create_vm().unwrap();
-    /// // This examples is based on https://lwn.net/Articles/658511/.
-    /// let mem_size = 0x4000;
-    /// let guest_addr: u64 = 0x1000;
+    /// let mut mem_size = 0x0;
```
andreeaflorescu: Any reason we cannot have mem_size and guest_addr the same on aarch64 and x86_64? This would simplify the code and remove some of the architecture specific code.

src/ioctls/vm.rs
andreeaflorescu: This comment feels a bit out of place here before the initialization of guest memory. Can we move it at the beginning of the example and say something like for x86_64 the example is based on bla bla?

src/ioctls/vm.rs
```diff
@@ -418,36 +430,61 @@ impl VmFd {
...
-    ///         0xf4,             /* hlt */
+    /// #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    /// // ASM code that just forces a MMIO Write.
+    /// let asm_code = [
+    ///         0xc6, 0x06, 0x00, 0x80, 0x00,
+    /// ];
+    /// #[cfg(target_arch = "aarch64")]
+    /// let asm_code = [
+    ///     0x01, 0x00, 0x00, 0x10, /* adr x1, <this address> */
+    ///     0x22, 0x10, 0x00, 0xb9, /* str w2, [x1, #16]; write to this page */
```
andreeaflorescu, line 731: Probably this is a stupid question, but I don't understand why you need this write. On x86_64 loading the code will dirty one page, on aarch64 it seems that this doesn't happen. Is this write required to dirty one page?

src/ioctls/vm.rs
```diff
@@ -418,36 +430,61 @@ impl VmFd {
...
+    ///     vcpu_fd.set_one_reg(core_reg_base + 2 * 32, guest_addr); // set PC
+    ///     vcpu_fd.set_one_reg(core_reg_base + 2 * 0, mmio_addr); // set X0
+    /// }
     ///
     /// loop {
     ///     match vcpu_fd.run().expect("run failed") {
-    ///         VcpuExit::Hlt => {
-    ///             // The code snippet dirties 1 page when loading the code in memory.
+    ///         VcpuExit::MmioWrite(addr, data) => {
+    ///             // The code snippet dirties its own page.
```
andreeaflorescu: On x86_64 loading the code dirties one page. On aarch64 it looks like you need a write to the guest memory. The comment needs to be updated as well.

src/lib.rs
```diff
@@ -24,8 +24,7 @@
 //! # Example - Running a VM on x86_64
 //!
 //! In this example we are creating a Virtual Machine (VM) with one vCPU.
-//! On the vCPU we are running x86_64 specific code. This example is based on
```
andreeaflorescu, line 27: I would still leave the LWN article as a reference here. Even though the arm code is different, we still follow the workflow as described in that article. We can say that for aarch64 the code is adapted to work on that platform.

dianpopa: Done.

src/lib.rs
dianpopa: Removed this comment since I was already mentioning it in the foregoing comment.

src/lib.rs
dianpopa: Done, take a look.

src/lib.rs
dianpopa: Indeed, you can see that the asm code for aarch64 contains no halt instruction so we specifically need to break out of the loop when the expected MMIO write exit happens.

src/ioctls/vm.rs
dianpopa: Done.

src/ioctls/vm.rs
dianpopa: Indeed, we can. Modified it accordingly.

src/ioctls/vm.rs
dianpopa: Given the last change, there is no need for that anymore. Take a look.

src/ioctls/vm.rs
dianpopa: Done.

src/lib.rs
dianpopa: It is a write, I updated the comment.

src/ioctls/vm.rs
dianpopa: Indeed, on aarch64 no dirty bit is set when the code is loaded into memory so yes, the write is required to dirty one page. 
After a short research it might seem that there are some specific dirty pages tracking on x86 (like for example [these](https://elixir.free-electrons.com/linux/v4.14.162/source/arch/x86/include/asm/kvm_host.h#L1040)), so it might be that those make a difference.

src/ioctls/vm.rs
dianpopa: I updated it, take a look.

dianpopa (commented):
@andreeaflorescu @aghecenco addressed comments. Please re-review.

src/ioctls/vm.rs
```diff
@@ -721,36 +720,62 @@ impl VmFd {
...
+    ///     vcpu_fd.set_one_reg(core_reg_base + 2 * 32, guest_addr); // set PC
+    ///     vcpu_fd.set_one_reg(core_reg_base + 2 * 0, mmio_addr); // set X0
+    /// }
     ///
     /// loop {
     ///     match vcpu_fd.run().expect("run failed") {
-    ///         VcpuExit::Hlt => {
-    ///             // The code snippet dirties 1 page when loading the code in memory.
+    ///         VcpuExit::MmioWrite(addr, data) => {
+    ///             // On x86_64, the code snippet dirties 1 page when loding the code in memory
```
alxiord:
```suggestion
    ///             // On x86_64, the code snippet dirties 1 page when loading the code in memory
```

dianpopa: Solved, take a look.

src/lib.rs
andreeaflorescu: I think we should still add a comment here to say that this break it's only there for the example (so it doesn't loop forever), because in an actual VMM loop you will not want to exit on MMIO writes.

dianpopa: Done, take a look.

dianpopa (commented):
@andreeaflorescu @aghecenco I applied your suggestions.

alxiord (approved):

andreeaflorescu (approved):

PR #64, opened 2019-10-16
Implement MsrList and Msrs as FamStructWrappers

acatangiu:
# Description of changes

`kvm_msr_list` and `kvm_msrs` each contain a flexible array member and were forcing users of `kvm-ioctls` crate to work with `unsafe` code.
Use the `kvm-bindings` `CpuId`, `MsrList` and `Msrs` safe wrappers over them and change our ioctl functions to use these safe wrappers.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 Cargo.lock
 /target
+.idea
 **/*.rs.bk
 **/.pytest_cache/
 **/__pycache__/*
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ license = "Apache-2.0 OR MIT"
 
 [dependencies]
 libc = ">=0.2.39"
-kvm-bindings = ">=0.1.0"
+kvm-bindings = { version = ">=0.2.0", features = ["fam-wrappers"] }
 vmm-sys-util = ">=0.1.1"
 
 [dev-dependencies]
--- a/coverage_config.json
+++ b/coverage_config.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 92.0,
+  "coverage_score": 91.9,
   "exclude_path": "",
   "crate_features": ""
 }
--- a/src/ioctls/mod.rs
+++ b/src/ioctls/mod.rs
@@ -13,10 +13,6 @@ use std::ptr::null_mut;
 use std::result;
 
 use kvm_bindings::kvm_run;
-#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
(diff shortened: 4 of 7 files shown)


jiangliu:
Could we implement this in kvm-bindings crate? It should be simpler and natural to implement mechanism to access syscall related structs within the kvm-bindings crate.

andreeaflorescu:
@jiangliu I don't think that implementing these traits in the kvm-bindings crate is a good idea at the moment. There are a few reasons for that:

1. kvm-bindings still separates clearly the autogenerated code from the one that is not autogenerated. I would prefer to keep it this way for now.
2. we are in discussions about merging/collaborating with ketuvim (https://github.com/rust-vmm/community/issues/62). Probably it will become more clear after the rust-vmm meetup what is the plan there.

While I believe that it's not the best thing now to move this functionality in kvm-bindings, we can discuss about it again once the rust-vmm components and their separation are more obvious.

What do you say?

jiangliu:
> @jiangliu I don't think that implementing these traits in the kvm-bindings crate is a good idea at the moment. There are a few reasons for that:
> 
> 1. kvm-bindings still separates clearly the autogenerated code from the one that is not autogenerated. I would prefer to keep it this way for now.
When enabling serde for kvm-bindings, I ran into the same issue to add non-auto-generated code to the kvm-bindings crate. In some sense, we are enhancing the bindgen tool to better support FAM structs, so 
propose to implement in the kvm-binding crate.
It's great to collaborate with the ketuvim project, which may help to solve all these issues.

> 2. we are in discussions about merging/collaborating with ketuvim ([rust-vmm/community#62](https://github.com/rust-vmm/community/issues/62)). Probably it will become more clear after the rust-vmm meetup what is the plan there.
> 
> While I believe that it's not the best thing now to move this functionality in kvm-bindings, we can discuss about it again once the rust-vmm components and their separation are more obvious.
> 
> What do you say?

acatangiu:
After some thought, I agree that the FamWrappers over the kvm FAM structures belong in the same crate as the structs themselves (kvm-bindings). Doing that also simplifies the implementation since we can directly implement the external FamStructWrapper trait on a local struct.

I've opened https://github.com/rust-vmm/kvm-bindings/pull/8 that exemplifies how that'd look.
The PR adds a new module in that crate that doesn't touch the generated files, it just imports the structs from there, so that's nice. The downside is that `kvm-bindings` would depend on `vmm-sys-utils`.

acatangiu:
The PR is updated to use the `kvm-bindings` safe wrappers. Please review when you get the chance.

alxiord (approved):

jiangliu (approved):

PR #65, opened 2019-10-24
vcpu: Add wrapper for KVM_GET_ONE_REG.

sandreim:
- add ioctl definition in kvm_ioctls.rs
- implement vcpu.get_one_reg()
- enable KVM_GET_ONE_REG / KVM_SET_ONE_REG for all platforms.

Signed-off-by: Andrei Sandu <sandreim@amazon.com>

### DIFF
--- a/src/ioctls/vcpu.rs
+++ b/src/ioctls/vcpu.rs
@@ -972,6 +972,30 @@ impl VcpuFd {
         Ok(())
     }
 
+    /// Returns the value of the specified vCPU register.
+    ///
+    /// The id of the register is encoded as specified in the kernel documentation
+    /// for `KVM_GET_ONE_REG`.
+    ///
+    /// # Arguments
+    ///
+    /// * `reg_id` - ID of the register.
+    ///
+    #[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
+    pub fn get_one_reg(&self, reg_id: u64) -> Result<u64> {
+        let mut reg_value = 0;
+        let mut onereg = kvm_one_reg {
+            id: reg_id,
+            addr: &mut reg_value as *mut u64 as u64,
+        };
+
+        let ret = unsafe { ioctl_with_mut_ref(self, KVM_GET_ONE_REG(), &mut onereg) };
+        if ret < 0 {
+            return Err(io::Error::last_os_error());
+        }
+        Ok(reg_value)
+    }
+
     /// Triggers the running of the current virtual CPU returning an exit reason.
     ///
     /// See documentation for `KVM_RUN`.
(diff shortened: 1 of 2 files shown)


src/ioctls/vcpu.rs
```diff
@@ -972,6 +972,29 @@ impl VcpuFd {
...
+    /// Returns the value of the specified vCPU register.
+    ///
+    /// The id of the register is encoded as specified in the kernel documentation
+    /// for `KVM_SET_ONE_REG`.
+    ///
+    /// # Arguments
+    ///
+    /// * `reg_id` - ID of the register.
+    ///
+    pub fn get_one_reg(&self, reg_id: u64) -> Result<u64> {
```
acatangiu, line 985:
```suggestion
    #[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
    pub fn get_one_reg(&self, reg_id: u64) -> Result<u64> {
```

sandreim: Thx.

src/ioctls/vcpu.rs
```diff
@@ -972,6 +972,30 @@ impl VcpuFd {
...
+    ///
+    /// The id of the register is encoded as specified in the kernel documentation
+    /// for `KVM_SET_ONE_REG`.
+    ///
+    /// # Arguments
+    ///
+    /// * `reg_id` - ID of the register.
+    ///
+    #[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
+    pub fn get_one_reg(&self, reg_id: u64) -> Result<u64> {
```
andreeaflorescu, line 985: This ioctl behaves (sort of) the same as `kvm_get_irqchip` and `kvm_get_msrs`. To keep the kvm-ioctls interface consistent (and as described in the KVM API documentation) this function should receive a mutable `kvm_one_reg` structure and return the register in that structure and the error code in the result.

For example, this is the signature of `get_irqchip`:
```rust
pub fn get_irqchip(&self, irqchip: &mut kvm_irqchip) -> Result<()>
```

sandreim: The structure **kvm_one_reg** doesn't actually contain any output data, and it is not written to by the kvm. The result of the KVM_GET_ONE_REG call is an u64 value  and will be written to the address pointed to by **kvm_one_reg**.**addr**. 

Based on the above I believe the current interface is better since it does not put any overhead of filling the kvm_one_reg struct on the user.

acatangiu: I kind of like the model where the ioctl wrapper does the heavy work (working with the kernel structs) as well.

In retrospect this is what should've been done for `get_irqchip` as well. I will open a RFC to that end 😄

src/ioctls/vcpu.rs
```diff
@@ -972,6 +972,30 @@ impl VcpuFd {
         Ok(())
     }
 
+    /// Returns the value of the specified vCPU register.
+    ///
+    /// The id of the register is encoded as specified in the kernel documentation
+    /// for `KVM_SET_ONE_REG`.
```
dianpopa:
```suggestion
    /// for `KVM_GET_ONE_REG`.
```

dianpopa (dismissed):

acatangiu:
Please squash your commits.

sandreim:
> Please squash your commits.

I applied Diana's suggestion in this web interface and it added a new commit. I'll force push to make it look good again.

src/ioctls/vcpu.rs
andreeaflorescu: Let's leave it as is right now. It's also a very subjective thing.

src/ioctls/vcpu.rs
sameo: No big deal, but this is not an `aarch64` ioctl.

sameo (approved):
You can merge it as is and address my comment with a follow-up PR.

andreeaflorescu (approved):

src/ioctls/vcpu.rs
sandreim: The set_one_reg implementation is also aarch64 only. I will followup with another PR and fix both of them.

sameo: Both `GET_ONE_REG` and `SET_ONE_REG` are available for all archs.

jiangliu: KVM_GET_ONE_REG/KVM_SET_ONE_REG are defined for all archs, but they are gated by KVM_CAP_ONE_REG. So no special treatment needed for x86.

Issue #39, opened 2019-05-15
functions that take a Rawfd should be unsafe


bonzini:
This crate allows registering eventfds with the kernel from safe code, but that function takes a raw file descriptor, and has no way to check that that the VM will not outlive the file descriptor. Thus, register_ioevent and register_irqfd should be unsafe (I didn't check if there are more instances).

andreeaflorescu:
We have two options here:
1. Mark register_ioevent and register_irqfd as unsafe
2. Use `EventFd` from vmm-sys-utils

The reason I replaced `EventFd` with `RawFd` was because I didn't want to take a dependency on vmm-sys-utils since we are using the eventfd as a rawfd anyway.

Any opinions on this one? @sameo @bonzini @alexandruag

bonzini:
Yes, using `EventFd` would be better (and safe). However, to do it properly would require to pass a `&EventFd`, so please first check in Firecracker that it would be easy to pass the EventFd object by reference there.

andreeaflorescu:
@bonzini That was the initial implementation in Firecracker, so it shouldn't be hard to switch back to it.

andreeaflorescu:
About the dependency on vmm-sys-utils, nvm my brain was not functioning properly. We will take the dependency anyway since we are going to use the ioctls macro defines so I might as well just use a reference to EventFd.

andreeaflorescu:
Fixed by #68

PR #68, opened 2019-10-28
Cleanup - use modules defined in vmm-sys-util

### LINKED ISSUE #23: Remove sys_ioctls.rs from project
Once we have the functionality merged in the vmm-sys-util crate, we should remove the sys_ioctls.rs file and use the ioctl defined from there.

andreeaflorescu:
Changes:
- Removed ioctl macros from repository and use the ones defined in vmm-sys-util instead
- Fixed irqchip related tests
- use Eventfd instead of RawFd
- Remove Cargo.lock

Fixes: https://github.com/rust-vmm/kvm-ioctls/issues/23 and https://github.com/rust-vmm/kvm-ioctls/issues/39


jiangliu (approved):

acatangiu (commented):
Why limit the interface to using `EventFd`? Why not use `& dyn AsRawFd`? Safety is guaranteed and the user gets more flexibility.

jiangliu:
It's good to explicitly state an KVM EventFd is needed here.

acatangiu:
> Why limit the interface to using EventFd? Why not use & dyn AsRawFd? Safety is guaranteed and the user gets more flexibility.

After going through the kernel KVM documentation and source, I see that it actually **requires** an `eventfd`. Disregard my previous comment.

src/ioctls/vm.rs
```diff
@@ -530,7 +531,7 @@ impl VmFd {
     ///
     /// # Arguments
     ///
-    /// * `fd` - FD which will be signaled. When signaling, the usual `vmexit` to userspace
+    /// * `fd` - `EventFd` which will be signaled. When signaling, the usual `vmexit` to userspace
```
acatangiu, line 534: nit: maybe use an explicit variable name
```suggestion
    /// * `evtfd` - `EventFd` which will be signaled. When signaling, the usual `vmexit` to userspace
```

acatangiu (approved):

alxiord (approved):

PR #70, opened 2019-10-29
vm: Add a function to unregister ioeventfd

sboeuf:
The KVM_IOEVENTFD ioctl allows for both assigning and deassigning events
to a specific PIO or MMIO guest address. This commit introduces the
missing function to perform the deassignment.

### DIFF
--- a/coverage_config.json
+++ b/coverage_config.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 92.3,
+  "coverage_score": 92.5,
   "exclude_path": "",
   "crate_features": ""
 }
--- a/src/ioctls/vm.rs
+++ b/src/ioctls/vm.rs
@@ -591,6 +591,73 @@ impl VmFd {
         }
     }
 
+    /// Unregisters an event from a certain address it has been previously registered to.
+    ///
+    /// See the documentation for `KVM_IOEVENTFD`.
+    ///
+    /// # Arguments
+    ///
+    /// * `fd` - FD which will be unregistered.
+    /// * `addr` - Address being written to.
+    ///
+    /// # Safety
+    ///
+    /// This function is unsafe because it relies on RawFd.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # extern crate kvm_ioctls;
+    /// extern crate libc;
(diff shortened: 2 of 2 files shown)


andreeaflorescu:
I am currently trying to move away from RawFd as a parameter for things and use the friendly (and safe) wrapper `EventFd`: https://github.com/rust-vmm/kvm-ioctls/pull/68

If you don't want to block on the other PR being merged, I would say to mark the function as unsafe. What do you think?

sboeuf:
@andreeaflorescu 
> I am currently trying to move away from RawFd as a parameter for things and use the friendly (and safe) wrapper EventFd: #68

Oh that's great :)

>If you don't want to block on the other PR being merged, I would say to mark the function as unsafe. What do you think?

Yes I will go ahead and tag it as unsafe for now, so that we can move both `register_ioeventfd` and `unregister_ioeventfd` to `EventFd` all at once.

sboeuf:
@andreeaflorescu PR is ready

Cargo.lock
```diff
@@ -12,7 +12,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "kvm-ioctls"
-version = "0.2.0"
+version = "0.3.0"
```
acatangiu: It's weird that for some crates the `Cargo.lock` file is kept in the subversioning system (tracked in `git`), while for other it isn't.

I know there have been some discussions around which is the best practice. Is there a conclusion? Can we align all crates on this?

acatangiu (dismissed):

Cargo.lock
sboeuf: Yes you're right, I don't know what's the common decision on this and I'll remove it if needed!

andreeaflorescu: We agreed on removing them, but I forgot to do so for this crate. @sboeuf do you want to remove it in your PR or should we open a new one?

sboeuf: Let me remove the Cargo commit from my PR.

acatangiu (approved):

andreeaflorescu (approved):

sboeuf:
@andreeaflorescu @acatangiu wow that was quick, thanks for merging :)

Issue #59, opened 2019-10-08
kvm-ioctls 0.4.0


sameo:
With cloud-hypervisor we're using the GSI routing ioctl (commit 37669f60). Since we've also added support `unregister_irqfd` since 0.2.0, and there are no pending new features, should we do a 0.3.0 release?

sameo:
cc @andreeaflorescu

andreeaflorescu:
Yap, I was planning on doing that. For this release I would also like to update kvm-ioctls to use vmm-sys-util. I'll do that these days.

sameo:
I'd also suggest you add the rust-vmm gatekeepers team to the crate owners, so that we can help with the publishing part.

andreeaflorescu:
Yap, will do. @acatangiu will help with the necessary changes.

jiangliu:
I still have one small PR for 0.3.0 version, will post it soon:)

andreeaflorescu:
Since vmm-sys-util is now available, we should also fix the following issues before publishing on crates.io:

- [x] use errno from vmm-sys-util instead of std::io::error: https://github.com/rust-vmm/kvm-ioctls/issues/35
- [x] remove ioctls.rs: https://github.com/rust-vmm/kvm-ioctls/issues/23
- [x] fix export of cpuid related consts: https://github.com/rust-vmm/kvm-ioctls/issues/16
- [x] use Eventfd instead of RawFds so that we don't have to have unsafe functions: https://github.com/rust-vmm/kvm-ioctls/issues/39
- [x] Use `FamStructWrapper` from vmm-sys-util: https://github.com/rust-vmm/kvm-ioctls/pull/58

andreeaflorescu:
We published 0.3.0, let's do the remaining items for our next release.

PR #71, opened 2019-11-22
use errno wrapper instead of io::Error in Result types

### LINKED ISSUE #35: use errno wrapper instead of io::Error in Result types
Because `io::Error` type can have an `ErrorKind` that is meaningless to a vmm, a simple errno wrapper is preferred. Occasionally, the `io::Error` function that converts an errno to `ErrorKind` has some surprising mappings, although many of them simply map to `Other`, so I wish to sidestep the problem entirely.

For display purposes, the `Display` impl can simply pass through to `io::Error`, which does have a useful `Display` impl for any relevant errno.

acatangiu:
Because io::Error type can have an ErrorKind that is meaningless to a vmm, a simple errno wrapper is preferred. Occasionally, the io::Error function that converts an errno to ErrorKind has some surprising mappings, although many of them simply map to Other, so I wish to sidestep the problem entirely.

For display purposes, the Display impl can simply pass through to io::Error, which does have a useful Display impl for any relevant errno.

Fixes #35

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ license = "Apache-2.0 OR MIT"
 [dependencies]
 libc = ">=0.2.39"
 kvm-bindings = { version = ">=0.2.0", features = ["fam-wrappers"] }
-vmm-sys-util = ">=0.1.1"
+vmm-sys-util = ">=0.2.1"
 
 [dev-dependencies]
 byteorder = ">=1.2.1"
--- a/src/ioctls/mod.rs
+++ b/src/ioctls/mod.rs
@@ -5,12 +5,11 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the THIRD-PARTY file.
 
-use std::io;
 use std::os::unix::io::AsRawFd;
 use std::ptr::null_mut;
-use std::result;
 
 use kvm_bindings::kvm_run;
+use vmm_sys_util::errno;
 
 /// Wrappers over KVM device ioctls.
 pub mod device;
@@ -23,9 +22,9 @@ pub mod vm;
 
 /// A specialized `Result` type for KVM ioctls.
 ///
-/// This typedef is generally used to avoid writing out io::Error directly and
+/// This typedef is generally used to avoid writing out errno::Error directly and
 /// is otherwise a direct mapping to Result.
-pub type Result<T> = result::Result<T, io::Error>;
+pub type Result<T> = std::result::Result<T, errno::Error>;
 
 /// Safe wrapper over the `kvm_run` struct.
 ///
@@ -64,7 +63,7 @@ impl KvmRunWrapper {
             )
         };
         if addr == libc::MAP_FAILED {
-            return Err(io::Error::last_os_error());
+            return Err(errno::Error::last());
         }
 
         Ok(KvmRunWrapper {
--- a/src/ioctls/mod.rs
+++ b/src/ioctls/mod.rs
@@ -49,7 +49,7 @@ impl KvmRunWrapper {
     /// # Arguments
     /// * `fd` - File descriptor to mmap from.
     /// * `size` - Size of memory region in bytes.
-    pub fn mmap_from_fd(fd: &AsRawFd, size: usize) -> Result<KvmRunWrapper> {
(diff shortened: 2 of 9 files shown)


src/ioctls/vcpu.rs
```diff
@@ -8,13 +8,14 @@
...
-use std::io;
 use std::os::unix::io::{AsRawFd, RawFd};
 
-use ioctls::{KvmRunWrapper, Result};
+use ioctls::KvmRunWrapper;
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
 use kvm_bindings::{CpuId, Msrs};
 use kvm_ioctls::*;
+use vmm_sys_util::errno::Error as Errno;
+use vmm_sys_util::errno::Result as ErrnoResult;
```
andreeaflorescu: I noticed that in the other commits you use errno::Error. I think it's better to be consistent here as well and import `vmm_sys_util::errno` instead.

acatangiu: Oh, this is leftover code, good catch!

acatangiu:
@andreeaflorescu @sameo @jiangliu I've addressed the last issue listed in https://github.com/rust-vmm/kvm-ioctls/issues/59 so that we can release `kvm-ioctls v0.4.0`

jiangliu (dismissed):

Cargo.toml
```diff
@@ -11,7 +11,7 @@ license = "Apache-2.0 OR MIT"
 [dependencies]
 libc = ">=0.2.39"
 kvm-bindings = { version = ">=0.2.0", features = ["fam-wrappers"] }
-vmm-sys-util = ">=0.1.1"
+vmm-sys-util = ">=0.2.1"
```
jiangliu, line 14: Should it be 0.2.0?

acatangiu: 0.2.1 contains a fix for clone of FamSafeWrappers which were released in 0.2.0, I chose to enforce using the fixed version of the crate.

src/ioctls/mod.rs
```diff
@@ -25,7 +24,7 @@ pub mod vm;
 ///
 /// This typedef is generally used to avoid writing out io::Error directly and
 /// is otherwise a direct mapping to Result.
-pub type Result<T> = result::Result<T, io::Error>;
+//pub type Result<T> = result::Result<T, io::Error>;
```
andreeaflorescu: If we don't want to use this anymore, we should also delete the comment and the commented code.

src/lib.rs
```diff
@@ -198,6 +198,6 @@ pub use ioctls::vm::{IoEventAddress, NoDatamatch, VmFd};
 ///
 /// ```
 /// #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
-/// use kvm_ioctls::{KvmRunWrapper, Result};
+/// use kvm_ioctls::KvmRunWrapper;
```
andreeaflorescu: Can you also add a comment in the commit message to say that the Result is no longer exported and the reason behind that?

acatangiu:
Changed `kvm_ioctls::Result` to work with `errno::Result` instead of `io::Result`.
    
No longer exporting `kvm_ioctls::Result`. Users of this crate should not have to use `kvm_ioctls::Result` outside the crate.
    
Now exporting `kvm_ioctls::Error` type definition so that users of this crate can create their own wrapping errors without having to know the `Error` type used internally by this crate.

```rust
enum UserError {
  FailedSomeOperation(kvm_ioctls::Error),
}
```
No need to assume `io::Error` or `errno::Error` used internally by `kvm_ioctls`.

@andreeaflorescu @jiangliu what do you think?

src/ioctls/mod.rs
acatangiu: done

src/lib.rs
acatangiu: done

andreeaflorescu (approved):
As a future improvement we might want to have a typedef for KvmError or similar and replace errno::Error with KvmError. This should help if we need to change the error type in the future as we only need to change it in one place.

jiangliu (approved):

Issue #67, opened 2019-10-28
Add proper tests for register/unregister irqfd on aarch64


andreeaflorescu:
Both the unit tests and the example only test the error cases on aarch64. We should add proper unit tests for these functions where the operations are successful & examples that actually call the functions.

MrXinWang:
Will propose a PR to address this soon.

andreeaflorescu:
Fixed by: https://github.com/rust-vmm/kvm-ioctls/pull/74

PR #74, opened 2019-12-09
Add proper unit tests for register/unregister irqfd on aarch64

MrXinWang:
Referencing [issue#67](https://github.com/rust-vmm/kvm-ioctls/issues/67) and existing x86_64 code, proposing some code for testing successful register/unregister irqfd case on aarch64 machines after the failure cases. 

It passes the unit tests on both my x86_64 and aarch64 machines for now.

### DIFF
--- a/src/ioctls/device.rs
+++ b/src/ioctls/device.rs
@@ -59,14 +59,11 @@ impl FromRawFd for DeviceFd {
 mod tests {
     use super::*;
     use ioctls::system::Kvm;
-    #[cfg(target_arch = "aarch64")]
-    use kvm_bindings::kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V2;
     #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
-    use kvm_bindings::kvm_device_type_KVM_DEV_TYPE_VFIO;
     use kvm_bindings::{
-        kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V3, KVM_CREATE_DEVICE_TEST, KVM_DEV_VFIO_GROUP,
-        KVM_DEV_VFIO_GROUP_ADD,
+        kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V3, kvm_device_type_KVM_DEV_TYPE_VFIO,
     };
+    use kvm_bindings::{KVM_CREATE_DEVICE_TEST, KVM_DEV_VFIO_GROUP, KVM_DEV_VFIO_GROUP_ADD};
 
     #[test]
     #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
@@ -110,6 +107,7 @@ mod tests {
     #[test]
     #[cfg(target_arch = "aarch64")]
     fn test_create_device() {
+        use ioctls::vm::create_gic_device;
         use kvm_bindings::kvm_device_type_KVM_DEV_TYPE_FSL_MPIC_20;
 
         let kvm = Kvm::new().unwrap();
@@ -124,16 +122,7 @@ mod tests {
         // the VGIC.
         assert!(vm.create_device(&mut gic_device).is_err());
 
-        gic_device.type_ = kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V3;
-
-        let device_fd = match vm.create_device(&mut gic_device) {
-            Ok(fd) => fd,
-            Err(_) => {
-                gic_device.type_ = kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V2;
-                vm.create_device(&mut gic_device)
-                    .expect("Cannot create KVM device")
-            }
-        };
+        let device_fd = create_gic_device(&vm, KVM_CREATE_DEVICE_TEST);
 
         // Following lines to re-construct device_fd are used to test
         // DeviceFd::from_raw_fd() and DeviceFd::as_raw_fd().
--- a/src/ioctls/vm.rs
+++ b/src/ioctls/vm.rs
@@ -1106,6 +1106,32 @@ impl AsRawFd for VmFd {
     }
 }
 
+/// Creates a dummy GIC device.
+///
+/// # Arguments
+///
+/// * `vm` - The vm file descriptor.
+/// * `flags` - Flags to be passed to `KVM_CREATE_DEVICE`.
+///
+#[cfg(test)]
+#[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
+pub(crate) fn create_gic_device(vm: &VmFd, flags: u32) -> DeviceFd {
+    let mut gic_device = kvm_bindings::kvm_create_device {
+        type_: kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V3,
+        fd: 0,
+        flags,
+    };
+    let device_fd = match vm.create_device(&mut gic_device) {
+        Ok(fd) => fd,
+        Err(_) => {
+            gic_device.type_ = kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V2;
+            vm.create_device(&mut gic_device)
+                .expect("Cannot create KVM vGIC device")
+        }
+    };
+    device_fd
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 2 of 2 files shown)


src/ioctls/vm.rs
```diff
@@ -1279,12 +1279,57 @@ mod tests {
             assert!(vm_fd.register_irqfd(&evtfd2, 8).is_ok());
             assert!(vm_fd.register_irqfd(&evtfd3, 4).is_ok());
         }
-
         // On aarch64, this fails because setting up the interrupt controller is mandatory before
         // registering any IRQ.
         // On x86_64 this fails as the event fd was already matched with a GSI.
         assert!(vm_fd.register_irqfd(&evtfd3, 4).is_err());
         assert!(vm_fd.register_irqfd(&evtfd3, 5).is_err());
+
```
andreeaflorescu: I think we are better of with 2 separate tests for x86_64 and for aarch64. The code is widely different depending on the platform. Maybe we can also have register and unregister tested in the same test?

MrXinWang: Hi @andreeaflorescu ! I think it perfectly makes sense, and therefore I have made some code changes:

- Separate the x86 test and the aarch64 test into two test functions.

- Rename the test function as `test_register_unregister_irqfd` and test both register and unregister irqfd in the body.

Please review.

src/ioctls/vm.rs
```diff
@@ -1261,62 +1261,97 @@ mod tests {
...
+        let raw_fd = unsafe { libc::dup(vgic_fd.as_raw_fd()) };
+        assert!(raw_fd >= 0);
+        let vgic_fd = unsafe { DeviceFd::from_raw_fd(raw_fd) };
+
+        // refer: https://github.com/firecracker-microvm/firecracker/blob/master/src/arch/src/aarch64/layout.rs
+        // First usable interrupt on aarch64.
+        const IRQ_BASE: u32 = 32;
+        // Last usable interrupt on aarch64.
+        const IRQ_MAX: u32 = 159;
+        let nr_irqs: u32 = IRQ_MAX - IRQ_BASE + 1;
```
andreeaflorescu: I don't think this logic is needed here. We can just use a dummy, but valid value for tests.

src/ioctls/vm.rs
```diff
@@ -1261,62 +1261,97 @@ mod tests {
...
+
+        //We need to tell the kernel how many irqs to support with this vgic.
```
andreeaflorescu: nit: please add a space after `//` in all comments.

MrXinWang: Done, sorry about that

src/ioctls/vm.rs
MrXinWang: Make sense, already changed.

src/ioctls/vm.rs
```diff
@@ -1261,62 +1261,94 @@ mod tests {
...
-        }
+
+        let mut gic_device = kvm_bindings::kvm_create_device {
+            type_: kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V3,
+            fd: 0,
+            flags: 0,
+        };
+
+        let vgic_fd = vm_fd.create_device(&mut gic_device);
+        let vgic_fd = match vgic_fd {
```
acatangiu: You can replace this with:
```rust
let vgic_fd = match vm_fd.create_device(&mut gic_device) {
```

src/ioctls/vm.rs
```diff
@@ -1261,62 +1261,94 @@ mod tests {
...
+                gic_device.type_ = kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V2;
+                vm_fd
+                    .create_device(&mut gic_device)
+                    .expect("Cannot create KVM vGIC device")
+            }
+        };
+
+        let raw_fd = unsafe { libc::dup(vgic_fd.as_raw_fd()) };
+        assert!(raw_fd >= 0);
+        let vgic_fd = unsafe { DeviceFd::from_raw_fd(raw_fd) };
```
acatangiu: Why is this necessary? `vgic_fd` is already a `DeviceFd` object.

Why `dup` its underlying `fd`, close the original `fd` (through dropping the original `vgic_fd`), then creating another `DeviceFd` from the duplicated `fd`?

The test should work the same without these 3 lines, right?

Production code is not playing with the underlying `fd`s like this, so unless you want to exercise some path that I'm missing, just use the original `DeviceFd` provided by `vm_fd.create_device()`.

acatangiu (commented):
Aside for the couple of inline comments, the code looks good. Thanks for the PR!

src/ioctls/vm.rs
MrXinWang: Done, thanks for this comment!

src/ioctls/vm.rs
MrXinWang: Ah yes of course, this is just because I saw similar code in [device.rs#97](https://github.com/rust-vmm/kvm-ioctls/blob/5a50456cd1afa18e079c97bf2e87be2572b4a3f9/src/ioctls/device.rs#L97) and was not sure whether I should adding them.....The test works for both cases.

MrXinWang: Then I just delete them :) Thanks for pointing it out!

acatangiu (dismissed):

src/ioctls/vm.rs
```diff
@@ -1258,62 +1258,89 @@ mod tests {
...
+
+        let vgic_fd = match vm_fd.create_device(&mut gic_device) {
+            Ok(fd) => fd,
+            Err(_) => {
+                gic_device.type_ = kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V2;
+                vm_fd
+                    .create_device(&mut gic_device)
+                    .expect("Cannot create KVM vGIC device")
+            }
+        };
```
andreeaflorescu: We probably want to add these lines to a test function so that they can be used in both test with irqfds and the test for create_device.

We can have a function such as:

```rust
#[cfg(test)]
/// Creates a GIC device using the KVM_CREATE_DEVICE_TEST flags.
pub (crate) fn create_test_gic_device() -> Result<DeviceFd> {
    let mut gic_device = kvm_bindings::kvm_create_device {
            type_: kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V3,
            fd: 0,
            flags: KVM_CREATE_DEVICE_TEST,
        };
       let device_fd = match vm.create_device(&mut gic_device) {
            Ok(fd) => fd,
            Err(_) => {
                gic_device.type_ = kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V2;
                vm.create_device(&mut gic_device)
            }
        };
}
```

Then in this test and in `test_create_device` from device.rs we can call this function and we don't need to have duplicated code. This will require some changes to the existing tests to test the error cases. What do you think?

MrXinWang: Hi @andreeaflorescu ! I got the idea and it really makes sense for reducing code duplication. 

However I do have one concern: refer to the [documentation of the kvm kernel api](https://www.kernel.org/doc/Documentation/virtual/kvm/api.txt), the `KVM_CREATE_DEVICE_TEST` flag will only check only test whether the device type is supported (instead of really create a vGIC I think), and therefore in the scope of this PR, if I set this flag, the `set_device_attr` will fail, leading to the test fail. Without this vGIC setup, I don't think we can test the successful register/unregister irqfd case on aarch64. Referencing the GIC implementation in [firecracker](https://github.com/firecracker-microvm/firecracker/blob/f55da99544f16280c31919b2596522c18d83e5e1/src/arch/src/aarch64/gic.rs#L116), this flag is also not set. Please do correct me if I am thinking in the wrong way, but I think the code in this PR is not related to the device.rs test cases? (I am really not sure about that......)

Wondering if you have any comments or ideas about this problem. Thank you!

MrXinWang: An idea occured to me that we probably can make this flag as an input argument of this `create_test_gic_device`, then the function would be:

```rust
#[cfg(test)]
/// Creates a GIC device for Arm architectures using either the KVM_CREATE_DEVICE_TEST flags or not.
pub (crate) fn create_gic_device(vm: &VmFd, test_flag: u32) -> DeviceFd {
    let mut gic_device = kvm_bindings::kvm_create_device {
        type_: kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V3,
        fd: 0,
        flags: test_flag,
    };
    let device_fd = match vm.create_device(&mut gic_device) {
        Ok(fd) => fd,
        Err(_) => {
            gic_device.type_ = kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V2;
            vm
             .create_device(&mut gic_device)
             .expect("Cannot create KVM vGIC device")
        }
    };
    device_fd
}
```
Calling that function in `test_register_unregister_irqfd` for `aarch64` would be either:
`let vgic_fd = create_gic_device(&vm_fd, 0);` or 
`let vgic_fd = create_gic_device(&vm_fd, KVM_CREATE_DEVICE_TEST);`

This proposal solves both cases, and I tested on my machine that it works. What do you think?

andreeaflorescu: Sounds good!

MrXinWang: Hi @andreeaflorescu, just updated the PR with regard of your comments. Would you mind reviewing it again? Thanks!

src/ioctls/vm.rs
```diff
@@ -1106,6 +1106,32 @@ impl AsRawFd for VmFd {
     }
 }
 
+/// Creates a GIC device for Arm architectures using either the KVM_CREATE_DEVICE_TEST flags or not.
```
andreeaflorescu: I think this should only specify that it creates a dummy GIC device (dummy because instead of a valid fd, we just write 0).
Isn't arm already implied when we say GIC? We should remove the part with "using either the KVM_CREATE_DEVICE_TEST flags or not" as this is covered by the Arguments section.

src/ioctls/vm.rs
```diff
@@ -1106,6 +1106,32 @@ impl AsRawFd for VmFd {
...
+///
+/// # Arguments
+///
+/// * `vm` - the vm file descriptor.
+/// * `test_flag` - use either KVM_CREATE_DEVICE_TEST or 0.
```
andreeaflorescu: We can name this to `flags` and update the description of this parameter. There is no enforcement for the flags to be either 0 or KVM_CREATE_DEVICE_TEST. We can update this to something along the lines of: "flags to be passed to `KVM_CREATE_DEVICE`."

src/ioctls/vm.rs
```diff
@@ -1258,62 +1284,76 @@ mod tests {
...
-            assert!(vm_fd.register_irqfd(&evtfd3, 4).is_ok());
-            assert!(vm_fd.unregister_irqfd(&evtfd2, 8).is_ok());
-            // KVM irqfd doesn't report failure on this case:(
-            assert!(vm_fd.unregister_irqfd(&evtfd2, 8).is_ok());
-        }
+
+        // Create the vGIC device.
+        let vgic_fd = create_gic_device(&vm_fd, 0);
+
+        // refer: https://github.com/firecracker-microvm/firecracker/blob/master/src/arch/src/aarch64/layout.rs
```
andreeaflorescu: In general it is better to have permanent links. If we decide to move `layout.rs` in Firecracker to another crate, that link will not work any more. I don't think we need a reference here because I believe we can just have a random (but correct value) directly written in kvm_device_attr.

src/ioctls/vm.rs
```diff
@@ -1258,62 +1284,76 @@ mod tests {
...
+        // Dummy interrupt for testing on aarch64.
+        let nr_irqs: u32 = 128;
+
+        // We need to tell the kernel how many irqs to support with this vgic.
+        let vgic_attr = kvm_bindings::kvm_device_attr {
+            group: kvm_bindings::KVM_DEV_ARM_VGIC_GRP_NR_IRQS,
+            attr: 0,
+            addr: &nr_irqs as *const u32 as u64,
```
andreeaflorescu, line 1331: There's no need for all the conversions. You can just write `addr: 128` or whatever else number you want to.

andreeaflorescu (commented):
Can you please also rework your commit message:
- follow the 50/72 rule: https://preslav.me/2015/02/21/what-s-with-the-50-72-rule/
- the summary can be something like: "Add unit tests for handling irqfd on aarch64"
- in the description you can mention that you added unit tests for register and unregister irqfd and that you added the new function `create_device` which is used for removing duplicated code when creating GIC devices.

src/ioctls/vm.rs
andreeaflorescu: I would still say to get rid of these two lines as it doesn't make a lot of sense to define `nr_irqs` (see below comment).

src/ioctls/vm.rs
MrXinWang: Hi @andreeaflorescu !

Sorry for questioning but according to the defination of kvm_device_attr from the kernel api:
```
struct kvm_device_attr {
	__u32	flags;		/* no flags currently defined */
	__u32	group;		/* device-defined */
	__u64	attr;		/* group-defined */
	__u64	addr;		/* userspace address of attr data */
};
```
I think this `addr` means the address of `nr_irqs` instead of the value of `nr_irqs`, and therefore I think we cannot simply use `addr: 128` (the test will fail too). Firecracker has a similar [implementation](https://github.com/firecracker-microvm/firecracker/blob/f55da99544f16280c31919b2596522c18d83e5e1/src/arch/src/aarch64/gic.rs#L99) of this and it also uses the pointer and similar conversion, so we probably need to keep this?

src/ioctls/vm.rs
MrXinWang: Ah yes sorry for being wordy...... Thanks very much for the suggstion, I have changed it and will push again after our discussion about this `nr_irq` stuff. Don't really want to bother you too many times for reviewing again and again (especially reviewing for these tiny issues which myself should have done better), so I will push the final version after all our discussions.

src/ioctls/vm.rs
MrXinWang: Same here.

src/ioctls/device.rs
```diff
@@ -110,30 +107,12 @@ mod tests {
...
         let kvm = Kvm::new().unwrap();
         let vm = kvm.create_vm().unwrap();
 
-        let mut gic_device = kvm_bindings::kvm_create_device {
-            type_: kvm_device_type_KVM_DEV_TYPE_FSL_MPIC_20,
-            fd: 0,
-            flags: KVM_CREATE_DEVICE_TEST,
-        };
-        // This fails on aarch64 as it does not use MPIC (MultiProcessor Interrupt Controller), it uses
-        // the VGIC.
```
MrXinWang, line 124: I think we can also keep this part as the error case? As this part is not really related to the `create_gic_device` function. @andreeaflorescu

src/ioctls/vm.rs
andreeaflorescu: Don't worry about it, I am not bothered at all. I'll reply to the other comments in line.

src/ioctls/vm.rs
andreeaflorescu: You are right, addr is the address of nr_irqs. My bad, I was not paying attention.

We should keep it as you wrote it right now.

src/ioctls/vm.rs
andreeaflorescu: I would say we can still name it flags. My logic here is that maybe flags will mean something else in future versions and maybe you will be able to pass something else other than 0 or KVM_CREATE_DEVICE_TEST (it might as well be that my logic is very flawed cause I really don't know much about arm). We can have the description of the parameter something like:
"flags to be passed to `KVM_CREATE_DEVICE`. Use `KVM_CREATE_DEVICE_TEST` flag to create a test device.". What do you think?

If we are sure that flags is always 0 or `KVM_CREATE_DEVICE`, than we can use a bool instead. We can have a parameter like `is_test_device: bool`.

src/ioctls/device.rs
andreeaflorescu: Yap, makes sense.

andreeaflorescu:
It looks like we still have to do the following before merging the PR:
- Update your commit message. See my comment here: https://github.com/rust-vmm/kvm-ioctls/pull/74#pullrequestreview-334575789
- Remove the link to Firecracker: https://github.com/rust-vmm/kvm-ioctls/pull/74#discussion_r359820350
- Update the documentation of create_gic_device: https://github.com/rust-vmm/kvm-ioctls/pull/74#discussion_r359817700
- re-add the error case as per your comment: https://github.com/rust-vmm/kvm-ioctls/pull/74#discussion_r360197818

I hope I didn't forget anything. Let me know if you have any questions.

MrXinWang:
Hi @andreeaflorescu ! Thanks soooo much for your patience and all these comments! I think I managed to change all of the code :) Now I think the code does look clean. Please review again.

src/ioctls/vm.rs
MrXinWang: Agreed :)

andreeaflorescu (approved):

alxiord (approved):

PR #75, opened 2019-12-11
Support GICv2 in testing KVM_CREATE_DEVICE.

michael2012z:
On Aarch64, retry to create VGICv2 device after failing to create
VGICv3, in case GICv2 is equipped (like RaspBerry Pi 4).
Divide test_create_device() into 2 architecture specific functions,
to avoid too much confusing conditional compiling sentences.

### DIFF
--- a/src/ioctls/device.rs
+++ b/src/ioctls/device.rs
@@ -59,41 +59,84 @@ impl FromRawFd for DeviceFd {
 mod tests {
     use super::*;
     use ioctls::system::Kvm;
+    #[cfg(target_arch = "aarch64")]
+    use kvm_bindings::kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V2;
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    use kvm_bindings::kvm_device_type_KVM_DEV_TYPE_VFIO;
     use kvm_bindings::{
-        kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V3, kvm_device_type_KVM_DEV_TYPE_VFIO,
-        KVM_CREATE_DEVICE_TEST, KVM_DEV_VFIO_GROUP, KVM_DEV_VFIO_GROUP_ADD,
+        kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V3, KVM_CREATE_DEVICE_TEST, KVM_DEV_VFIO_GROUP,
+        KVM_DEV_VFIO_GROUP_ADD,
     };
 
     #[test]
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
     fn test_create_device() {
-        #[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
+        let kvm = Kvm::new().unwrap();
+        let vm = kvm.create_vm().unwrap();
+
+        let mut gic_device = kvm_bindings::kvm_create_device {
+            type_: kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V3,
+            fd: 0,
+            flags: KVM_CREATE_DEVICE_TEST,
+        };
+        // This fails on x86_64 because there is no VGIC there.
+        assert!(vm.create_device(&mut gic_device).is_err());
+
+        gic_device.type_ = kvm_device_type_KVM_DEV_TYPE_VFIO;
+
+        let device_fd = vm
+            .create_device(&mut gic_device)
+            .expect("Cannot create KVM device");
+
+        // Following lines to re-construct device_fd are used to test
+        // DeviceFd::from_raw_fd() and DeviceFd::as_raw_fd().
+        let raw_fd = unsafe { libc::dup(device_fd.as_raw_fd()) };
+        assert!(raw_fd >= 0);
+        let device_fd = unsafe { DeviceFd::from_raw_fd(raw_fd) };
+
+        let dist_attr = kvm_bindings::kvm_device_attr {
+            group: KVM_DEV_VFIO_GROUP,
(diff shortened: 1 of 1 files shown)


andreeaflorescu (dismissed):

src/ioctls/device.rs
```diff
@@ -110,4 +104,50 @@ mod tests {
...
+            Err(_) => {
+                gic_device.type_ = kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V2;
+                vm.create_device(&mut gic_device)
+                    .expect("Cannot create KVM device")
+            }
+        };
+
+        let raw_fd = unsafe { libc::dup(device_fd.as_raw_fd()) };
+        assert!(raw_fd >= 0);
+        let device_fd = unsafe { DeviceFd::from_raw_fd(raw_fd) };
```
acatangiu: Why do this? `device_fd` is already a `DeviceFd`. Code should be equivalent without these 3 lines playing with the `unsafe` underlying `RawFd`.

src/ioctls/device.rs
```diff
@@ -110,4 +104,50 @@ mod tests {
...
+            flags: KVM_CREATE_DEVICE_TEST,
+        };
+        // This fails on aarch64 as it does not use MPIC (MultiProcessor Interrupt Controller), it uses
+        // the VGIC.
+        assert!(vm.create_device(&mut gic_device).is_err());
+
+        gic_device.type_ = kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V3;
+
+        let device_fd = vm.create_device(&mut gic_device);
+        let device_fd = match device_fd {
```
acatangiu: These 2 lines can be collapsed into:
```rust
let device_fd = match vm.create_device(&mut gic_device) {
```

michael2012z: Thanks.

src/ioctls/device.rs
michael2012z: Yes, these 3 lines are unnecessary. I introduced them when splitting from the old test function. :)
In x86 test case, the `raw fd` was not needed either. I deleted them as well.

coverage_config.json
```diff
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 91.9,
+  "coverage_score": 91.2,
```
michael2012z: I removed 3 unneeded code lines in x86, so the coverage degraded a little bit.

src/ioctls/device.rs
```diff
@@ -59,44 +59,75 @@ impl FromRawFd for DeviceFd {
...
-        } else if cfg!(any(target_arch = "arm", target_arch = "aarch64")) {
-            gic_device.type_ = kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V3;
-        }
+        gic_device.type_ = kvm_device_type_KVM_DEV_TYPE_VFIO;
+
         let device_fd = vm
             .create_device(&mut gic_device)
             .expect("Cannot create KVM device");
 
-        let raw_fd = unsafe { libc::dup(device_fd.as_raw_fd()) };
```
acatangiu, line 97: The code-coverage decrease prompted me to check and find out that these 3 lines in this particular test were there to also test (as coverage drop suggests) `DeviceFd::as_raw_fd()` and `DeviceFd::from_raw_fd()`.

I suggest you either:

- bring back those lines in this test only and properly document them with comments explaining what their purpose is in the test,
or
- add another test that covers `DeviceFd::as_raw_fd()` and `DeviceFd::from_raw_fd()`.

acatangiu (changes requested):
Decrease of coverage happens because `DeviceFd::as_raw_fd()` and `DeviceFd::from_raw_fd()` are no longer tested.

See inline comment.

src/ioctls/device.rs
michael2012z: Aha, yes, `as_raw_fd()` and `from_raw_fd()` are implemented for `DeviceFd`. Although they were not needed in our test cases, they should be tested.
I prefer to cover them in this test.

michael2012z:
An error was triggered in ARM unit tests:

> ---- ioctls::vm::tests::test_irq_chip stdout ------
> thread 'ioctls::vm::tests::test_irq_chip' panicked at 'assertion failed: vm.create_irq_chip().is_err()', src/ioctls/vm.rs:1153:13

Creating a device by `vm.create_irq_chip()` on a machine with GICv3 is supposed to fail. That is what the case asserted.
But if the GIC is v2, this operation will succeed. I think that's why this case failed: the job was deployed on a GICv2 machine (maybe a RPI4?).

michael2012z:
IMO, this line should be removed: https://github.com/rust-vmm/kvm-ioctls/blob/master/src/ioctls/vm.rs#L1153 .
And let `test_irq_chip()` be dedicated to X86 (skip this test for arm). Because:
1. For arm, this test is not a functioning one. It only asserted a failure case.
2. It's hardware dependent. On GIC v2 and v3, the result is totally different.

What do you think? @acatangiu

andreeaflorescu:
Hey @michael2012z the error is not related to your PR.

I changed the bare metal instance on which we run the Arm CI and in the process, I also upgraded the kernel version. These changes are triggering the faiIures you are seeing. I will fix the failures in a PR, but we should wait for the fixes to be merged before we can merge your PR. I'll keep you posted.

michael2012z:
Thank you, @andreeaflorescu .

acatangiu (approved):

michael2012z:
@andreeaflorescu Thank you for clearing the error with https://github.com/rust-vmm/kvm-ioctls/pull/78. 
Now all checks passed.

andreeaflorescu (approved):

Issue #11, opened 2019-04-01
Add the ioctls used by CrosVM


andreeaflorescu:
To be scoped.

MrXinWang:
Hi @andreeaflorescu ! 

For **aarch64** platform, there are currently 6 ioctls in CrosVM which are not included in rust-vmm, namely:

- `KVM_IRQ_LINE`

- `KVM_SET_SIGNAL_MASK`

- `KVM_SET_MEMORY_REGION`

- `KVM_HAS_DEVICE_ATTR`

- `KVM_SET_GUEST_DEBUG`

- `KVM_GET_REG_LIST`

Among these 6 ioctls, only the first two ioctls (`KVM_IRQ_LINE` and `KVM_SET_SIGNAL_MASK`) have some implemented functions to call them, respectively in [`/crosvm/kvm/src/lib.rs#744`](https://chromium.googlesource.com/chromiumos/platform/crosvm/+/refs/heads/master/kvm/src/lib.rs#744) and [`/crosvm/kvm/src/lib.rs#1628`](https://chromium.googlesource.com/chromiumos/platform/crosvm/+/refs/heads/master/kvm/src/lib.rs#1628). The other 4 ioctls are only declared but not implemented in [`/crosvm/kvm_sys/src/lib.rs`](https://chromium.googlesource.com/chromiumos/platform/crosvm/+/refs/heads/master/kvm_sys/src/lib.rs)

Wondering if we can add these ioctls for rust-vmm, if so, we @michael2012z will create some PRs to do that. Thanks!

andreeaflorescu:
We never planned for kvm-ioctls to have all the supported ioctls. I think it make sense to add ioctls as they're needed. So I would start with the ones used by CrosVM and we can continue from there. What do you think?

MrXinWang:
Hi @andreeaflorescu , thanks very much for the suggestion! Created a PR for the ioctl `KVM_IRQ_LINE` already, and I think my colleague will take care of the other one. Thanks.

andreeaflorescu:
Closing this as we are adding ioctls on an as-needed basis.

PR #76, opened 2019-12-13
Add ioctl KVM_IRQ_LINE from CrosVM to rust-vmm

MrXinWang:
As CrosVM has the ioctl `KVM_IRQ_LINE` and referencing issue #11 , this PR proposes a design for implementation and test for ioctl `KVM_IRQ_LINE`.

I think following the coding style in [vm.rs](https://github.com/rust-vmm/kvm-ioctls/blob/master/src/ioctls/vm.rs), we also need to add an example for this ioctl, however it seems that how to use this ioctl differs (a lot) between different machine architectures. Therefore I would like to get some comments about how we can write the example. Thanks! 

### DIFF
--- a/src/ioctls/vm.rs
+++ b/src/ioctls/vm.rs
@@ -878,6 +878,74 @@ impl VmFd {
         }
     }
 
+    /// Sets the level on the given irq to 1 if `active` is true, and 0 otherwise.
+    ///
+    /// # Arguments
+    ///
+    /// * `irq` - IRQ to be set.
+    /// * `active` - Level of the IRQ input.
+    ///
+    /// # Errors
+    ///
+    /// Returns an io::Error when the irq field is invalid
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # extern crate kvm_ioctls;
+    /// # extern crate libc;
+    /// # extern crate vmm_sys_util;
+    /// # use kvm_ioctls::{Kvm, VmFd};
+    /// # use libc::EFD_NONBLOCK;
+    /// # use vmm_sys_util::eventfd::EventFd;
+    /// fn arch_setup(vm_fd: &VmFd) {
+    ///     // Arch-specific setup:
+    ///     // For x86 architectures, it simply means calling vm.create_irq_chip().unwrap().
+    /// #   #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    /// #   vm_fd.create_irq_chip().unwrap();
+    ///     // For Arm architectures, the IRQ controllers need to be setup first.
+    ///     // Details please refer to the kernel documentation.
+    ///     // https://www.kernel.org/doc/Documentation/virtual/kvm/api.txt
+    /// #   #[cfg(any(target_arch = "arm", target_arch = "aarch64"))] {
+    /// #       vm_fd.create_vcpu(0).unwrap();
+    /// #       // ... rest of setup for Arm goes here
+    /// #   }
+    /// }
+    ///
+    /// let kvm = Kvm::new().unwrap();
+    /// let vm = kvm.create_vm().unwrap();
+    /// arch_setup(&vm);
+    /// #[cfg(any(target_arch = "x86", target_arch = "x86_64"))] {
+    ///     vm.set_irq_line(4, true);
+    ///     // ...
+    /// }
+    /// #[cfg(any(target_arch = "arm", target_arch = "aarch64"))] {
+    ///     vm.set_irq_line(0x01_00_0020, true);
+    ///     // ....
+    /// }
+    /// ```
+    ///
+    #[cfg(any(
+        target_arch = "x86",
+        target_arch = "x86_64",
+        target_arch = "arm",
+        target_arch = "aarch64"
+    ))]
+    pub fn set_irq_line(&self, irq: u32, active: bool) -> Result<()> {
+        let mut irq_level = kvm_irq_level::default();
+        irq_level.__bindgen_anon_1.irq = irq;
+        irq_level.level = if active { 1 } else { 0 };
+
+        // Safe because we know that our file is a VM fd, we know the kernel will only read the
+        // correct amount of memory from our pointer, and we verify the return result.
+        let ret = unsafe { ioctl_with_ref(self, KVM_IRQ_LINE(), &irq_level) };
+        if ret == 0 {
+            Ok(())
+        } else {
+            Err(errno::Error::last())
+        }
+    }
+
     /// Creates a new KVM vCPU file descriptor and maps the memory corresponding
     /// its `kvm_run` structure.
     ///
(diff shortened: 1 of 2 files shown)


src/ioctls/vm.rs
```diff
@@ -878,6 +878,34 @@ impl VmFd {
         }
     }
 
+    /// Sets the level on the given irq to 1 if `active` is true, and 0 otherwise.
+    ///
+    /// # Arguments
+    ///
+    /// * `irq` - IRQ to be set.
+    /// * `active` - Level of the IRQ input.
+    ///
```
alxiord, line 900: Please add an `# Errors` section (see other ioctl wrappers in this file).

alxiord (commented):
There appears to be a lot of setup to do on `aarch64` to get the (doc)tests to run, so I'd hide that away in the documentation, leaving it to the user to decide if they want to dive deep.

In the generated documentation, you could hide away the arch-specific details, but leave out a comment to get the user's attention.

Here's what I had in mind:
```rust
/// # Examples
///
/// ```rust
/// # extern crate kvm_ioctls;
/// # extern crate libc;
/// # extern crate vmm_sys_util;
/// # use kvm_ioctls::{Kvm, VmFd};
/// # use libc::EFD_NONBLOCK;
/// # use vmm_sys_util::eventfd::EventFd;
/// fn arch_setup(vm_fd: &VmFd) {
///     // Arch-specific setup.
///     // See kernel documentation.
/// #   #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
/// #   vm_fd.create_irq_chip().unwrap();
/// #   #[cfg(any(target_arch = "arm", target_arch = "aarch64"))] {
/// #       vm_fd.create_vcpu(0).unwrap();
/// #       // ... rest of setup goes here
/// #   }
/// }
///
/// let kvm = Kvm::new().unwrap();
/// let vm = kvm.create_vm().unwrap();
/// arch_setup(&vm);
///  #[cfg(any(target_arch = "x86", target_arch = "x86_64"))] {
///     vm.set_irq_line(4, true).unwrap();
///     // ...
/// }
/// #[cfg(any(target_arch = "arm", target_arch = "aarch64"))] {
///     vm_fd.set_irq_line(0x01_00_0020, true).unwrap();
///     // ....
/// }
/// ```
```

Here's what that would look like:

image

Of course something more helpful than `see kernel documentation` would be preferable and the indentation looks a bit off, but otherwise, WDYT?

src/ioctls/vm.rs
MrXinWang: Done. Thanks for pointing it out :)

MrXinWang:
Hi @aghecenco ! Thanks for this awesome suggestion! Already updated the examples with some other details based on your documentation framework. Please review.

src/ioctls/vm.rs
```diff
@@ -1358,6 +1426,68 @@ mod tests {
...
+        assert!(vm_fd.set_irq_line(4, false).is_ok());
+        assert!(vm_fd.set_irq_line(4, true).is_ok());
+    }
+
+    #[test]
+    #[cfg(target_arch = "aarch64")]
+    fn test_set_irq_line() {
+        let kvm = Kvm::new().unwrap();
+        let vm_fd = kvm.create_vm().unwrap();
+        // Create a vcpu for test case 2 of the KVM_IRQ_LINE API on aarch64.
```
andreeaflorescu, line 1503: What do you mean with test case 2 here?

src/ioctls/vm.rs
```diff
@@ -1358,6 +1426,68 @@ mod tests {
...
+        let nr_irqs: u32 = 128;
+
+        // We need to tell the kernel how many irqs to support with this vgic.
+        let vgic_attr = kvm_bindings::kvm_device_attr {
+            group: kvm_bindings::KVM_DEV_ARM_VGIC_GRP_NR_IRQS,
+            attr: 0,
+            addr: &nr_irqs as *const u32 as u64,
+            flags: 0,
+        };
+        assert!(vgic_fd.set_device_attr(&vgic_attr).is_ok());
```
andreeaflorescu: We can move this code in a helper function so we can also call that function from `test_register_unregister_irqfd`. We can call it something like `set_supported_nr_irqs`.

src/ioctls/vm.rs
```diff
@@ -1358,6 +1426,68 @@ mod tests {
...
+        // Finalize the GIC.
+        let vgic_attr = kvm_bindings::kvm_device_attr {
+            group: kvm_bindings::KVM_DEV_ARM_VGIC_GRP_CTRL,
+            attr: u64::from(kvm_bindings::KVM_DEV_ARM_VGIC_CTRL_INIT),
+            addr: 0,
+            flags: 0,
+        };
+        assert!(vgic_fd.set_device_attr(&vgic_attr).is_ok());
```
andreeaflorescu: This code can also be moved to a function and re-used in `test_register_unregister_irqfd`. We can call it something like `request_gic_init`. This name seems to be closer to the documentation than "finalize_gic":
> KVM_DEV_ARM_VGIC_CTRL_INIT
>       request the initialization of the VGIC, no additional parameter in
>       kvm_device_attr.addr.

src/ioctls/vm.rs
MrXinWang: In below comments:
```
// On arm/aarch64, irq field is interpreted like this:
// bits:  | 31 ... 24 | 23  ... 16 | 15    ...    0 |
// field: | irq_type  | vcpu_index |     irq_id     |
// The irq_type field has the following values:
// - irq_type[0]: out-of-kernel GIC: irq_id 0 is IRQ, irq_id 1 is FIQ
// - irq_type[1]: in-kernel GIC: SPI, irq_id between 32 and 1019 (incl.) (the vcpu_index field is ignored)
// - irq_type[2]: in-kernel GIC: PPI, irq_id between 16 and 31 (incl.)
// Hence, using irq_type = 1, irq_id = 32 (decimal), the irq field in hex is: 0x01_00_0020
// Case 2: using irq_type = 2, vcpu_index = 0, irq_id = 16 (decimal), the irq field in hex is: 0x02_00_0010
```

 `irq_type` can be either `1` or `2`, and I used case 1, 2 as two separate cases (for testing this same ioctl), namely `irq_type` equals `1` or `2`. Therefore case 2 means using `irq_type = 2`, `vcpu_index = 0`, `irq_id = 16 (decimal)`, the irq field in hex is then: `0x02_00_0010`

src/ioctls/vm.rs
MrXinWang: Done :)

src/ioctls/vm.rs
MrXinWang: Done :)

andreeaflorescu (approved):

alxiord (approved):

PR #77, opened 2019-12-17
Support KVM_SET_SIGNAL_MASK ioctl.

michael2012z:
Support KVM_SET_SIGNAL_MASK ioctl.

Reference to https://github.com/rust-vmm/kvm-ioctls/issues/11.

### DIFF
--- a/src/ioctls/vcpu.rs
+++ b/src/ioctls/vcpu.rs
@@ -5,15 +5,17 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the THIRD-PARTY file.
 
-use kvm_bindings::*;
-use libc::EINVAL;
-use std::fs::File;
-use std::os::unix::io::{AsRawFd, RawFd};
-
 use ioctls::{KvmRunWrapper, Result};
+use kvm_bindings::*;
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
 use kvm_bindings::{CpuId, Msrs};
 use kvm_ioctls::*;
+use libc::sigset_t;
+use libc::EINVAL;
+use std::fs::File;
+use std::mem::size_of;
+use std::os::unix::io::{AsRawFd, RawFd};
+use std::ptr::copy_nonoverlapping;
 use vmm_sys_util::errno;
 use vmm_sys_util::ioctl::{ioctl, ioctl_with_mut_ref, ioctl_with_ref};
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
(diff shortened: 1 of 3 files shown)


src/ioctls/vcpu.rs
```diff
@@ -233,6 +235,75 @@ impl VcpuFd {
...
+    /// # extern crate vmm_sys_util;
+    /// # extern crate libc;
+    /// # use kvm_ioctls::Kvm;
+    /// # use vmm_sys_util::signal;
+    /// # use libc::sigset_t;
+    /// # use libc::*;
+    /// let kvm = Kvm::new().unwrap();
+    /// let vm = kvm.create_vm().unwrap();
+    /// let vcpu = vm.create_vcpu(0).unwrap();
+    /// let sigset = signal::create_sigset(&[SIGINT, SIGKILL, SIGTERM]).unwrap();
```
alxiord, line 259: `SIGKILL` cannot be handled, please remove it from the example.

src/ioctls/vcpu.rs
```diff
@@ -233,6 +235,75 @@ impl VcpuFd {
...
+    /// vcpu.set_signal_mask(&sigset);
```
alxiord, line 260: Can you expand the example with a proof that signal blocking works?

src/ioctls/vcpu.rs
```diff
@@ -233,6 +235,75 @@ impl VcpuFd {
...
+    /// ```
+    ///
+    pub fn set_signal_mask(&self, sigset: &sigset_t) -> Result<()> {
+        // C struct of 'kvm_signal_mask' is in form of:
+        //   struct kvm_signal_mask {
+        //       __u32 len;
+        //       __u8  sigset[0];
+        //   };
+        // The payload array 'sigset' doesn't take any space in the struct.
+        // In Rust we construct a vector of 'kvm_signal_mask' with continuous space,
```
alxiord, line 270: This looks like it could be implemented with a [FAM struct](https://github.com/rust-vmm/vmm-sys-util/blob/master/src/fam.rs). Here's an example: https://github.com/rust-vmm/kvm-ioctls/blob/f372a9d046aa741860799e7003e65996d7734103/src/ioctls/system.rs#L235

src/ioctls/vcpu.rs
```diff
@@ -1254,6 +1325,17 @@ mod tests {
...
 
+    #[test]
+    fn test_set_signal_mask() {
+        use libc::*;
+        use vmm_sys_util::signal;
+        let kvm = Kvm::new().unwrap();
+        let vm = kvm.create_vm().unwrap();
+        let vcpu = vm.create_vcpu(0).unwrap();
+        let sigset = signal::create_sigset(&[SIGINT, SIGKILL, SIGTERM]).unwrap();
+        assert!(vcpu.set_signal_mask(&sigset).is_ok());
```
alxiord, line 1336: I'd prefer the test to prove that the functionality works, not just cover the lines.

src/ioctls/vcpu.rs
```diff
@@ -233,6 +235,75 @@ impl VcpuFd {
         Ok(())
     }
 
+    /// Specifies set of signals that are blocked during execution of KVM_RUN.
```
andreeaflorescu, line 238: I believe this functionality is needed for older kernel versions that do not support KVM_IMMEDIATE_EXIT to be able to kick vcpus. I am not super sure we should add this as we currently "support" only newer kernel versions. By support here I mean that we are always using KVM bindings coming from a kernel version > 4.20 where KVM_IMMEDIATE_EXIT is available.

CC: @rust-vmm/gatekeepers are there any usecases for which we also want to support this?

bonzini: I agree, KVM_CAP_IMMEDIATE_EXIT should be enough; if we want to support both, I'd rather have higher-level functionality that on one hand sets up the signal handler properly for KVM_CAP_IMMEDIATE_EXIT, and on the other hand knows how to "eat" the signal with sigwaitinfo (after running KVM_RUN) when not using KVM_CAP_IMMEDIATE_EXIT.

michael2012z: @andreeaflorescu , @bonzini . Thanks for reviewing.

Truely, `KVM_IMMEDIATE_EXIT` replaces `KVM_SET_SIGNAL_MASK` ioctl, there is good explanation in this page: http://patchwork.ozlabs.org/patch/726459 .

I have one thing to confirm about the last comment:
`KVM_SET_SIGNAL_MASK` ioctl is not needed even if KVM_CAP_IMMEDIATE_EXIT extension is not supported, right?

I will close this PR after getting confirmed.

src/ioctls/vcpu.rs
michael2012z: That's a good point! Thank you.
But as the later comments mentioned, it's quite probable that this PR will be canceled.
I will improve this if this PR continue.

michael2012z:
`KVM_SET_SIGNAL_MASK ` is not needed, `KVM_IMMEDIATE_EXIT` replaces it well.
Cancel this PR.

PR #78, opened 2019-12-18
make test_irq_chip x86 specific

andreeaflorescu:
The test was previously marked as x86_64 and aarch64 compatible.

On aarch64 we were testing only the error case. On older kernel
versions, creating an irqchip before creating the vCPU would return
an error. This restrictions is not in place anymore on Linux Kernel
4.15.

Since we cannot test this error path anymore on aarch64, the test is
now labeled as x86 specific.

Also updated the test to set the irq_base of the PIC and check that
this value is appropriately updated.

### DIFF
--- a/src/ioctls/vm.rs
+++ b/src/ioctls/vm.rs
@@ -1109,7 +1109,7 @@ impl AsRawFd for VmFd {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use {Cap, Kvm};
+    use Kvm;
 
     use libc::EFD_NONBLOCK;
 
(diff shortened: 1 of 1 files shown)


src/ioctls/vm.rs
```diff
@@ -1136,34 +1136,29 @@ mod tests {
...
+
+        // We initialize a dummy irq chip (`other_irqchip`) in which the
+        // function `get_irqchip` returns its result.
+        let mut other_irqchip = kvm_irqchip::default();
+        other_irqchip.chip_id = KVM_IRQCHIP_PIC_MASTER;
+        assert!(vm.get_irqchip(&mut other_irqchip).is_ok());
+
+        // Safe because we know that the irqchip type is PIC.
+        unsafe {
+            assert_eq!(irqchip.chip.pic.irq_base, other_irqchip.chip.pic.irq_base)
```
acatangiu: Why not `assert_eq!` on the whole `pic` instead of just the field we set?

andreeaflorescu: I was thinking that it is better to assert only on the values we intentionally set because maybe there are other fields in there that the kernel sets without input from us. I can change it if you feel strongly about this.

acatangiu (dismissed):

acatangiu (approved):

alxiord (approved):

PR #80, opened 2019-12-25
Add unit test for KVM_CREATE_IRQCHIP on ARM.

### LINKED ISSUE #79: [aarch64] Add test for create_irq_chip
On aarch64 we initialize the interrupt controller using `create_device`. This can be achieved with `create_irq_chip()` as well. I believe it only works with GICv2 (need to double check when scoping this issue). We should also add a test with `create_irq_chip`.

michael2012z:
Fix https://github.com/rust-vmm/kvm-ioctls/issues/79.

On ARM, KVM_CREATE_IRQCHIP is a bit more complex than on X86.
KVM_CREATE_IRQCHIP only creates VGICv2. It usually succeed with a GICv2 hardware. 
But if the hardware is GICv3, it's probable that GICv2 can't be emulated, 
so KVM_CREATE_IRQCHIP would fail.

In the test, we'd better check the ability of creating VGICv2 in advance, 
and assert correspondingly.

### DIFF
--- a/src/ioctls/vm.rs
+++ b/src/ioctls/vm.rs
@@ -146,12 +146,25 @@ impl VmFd {
     ///
     /// ```rust
     /// # extern crate kvm_ioctls;
+    /// # extern crate kvm_bindings;
     /// # use kvm_ioctls::Kvm;
     /// let kvm = Kvm::new().unwrap();
     /// let vm = kvm.create_vm().unwrap();
     ///
     /// #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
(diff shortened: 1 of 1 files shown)


andreeaflorescu (dismissed):

andreeaflorescu:
Can you also add the description from the PR in the commit message?

It is nice to also included the issue you are fixing in the commit message because it is easier to track changes.

michael2012z:
Updated commit message.

andreeaflorescu (approved):

alxiord (approved):

PR #81, opened 2020-01-09
vcpu: Add the system-level event type

xuyandong2:
This newly added KVM_EXIT_SYSTEM_EVENT type will be used by KVM
ARM/ARM64 in-kernel PSCI v0.2 support to reset/shutdown VMs.

Signed-off-by: Xu Yandong <xuyandong2@huawei.com>

### DIFF
--- a/src/ioctls/vcpu.rs
+++ b/src/ioctls/vcpu.rs
@@ -19,6 +19,29 @@ use vmm_sys_util::ioctl::{ioctl, ioctl_with_mut_ref, ioctl_with_ref};
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
 use vmm_sys_util::ioctl::{ioctl_with_mut_ptr, ioctl_with_ptr};
 
+/// The system-level event type.
+/// If exit_reason is KVM_EXIT_SYSTEM_EVENT then the vcpu has triggered
+/// a system-level event using some architecture specific mechanism.
+///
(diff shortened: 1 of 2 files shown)


src/ioctls/vcpu.rs
```diff
@@ -1150,7 +1173,22 @@ impl VcpuFd {
                 KVM_EXIT_WATCHDOG => Ok(VcpuExit::Watchdog),
                 KVM_EXIT_S390_TSCH => Ok(VcpuExit::S390Tsch),
                 KVM_EXIT_EPR => Ok(VcpuExit::Epr),
-                KVM_EXIT_SYSTEM_EVENT => Ok(VcpuExit::SystemEvent),
+                KVM_EXIT_SYSTEM_EVENT => unsafe {
```
andreeaflorescu, line 1176: Is there any way to test this exit code in [test_run_code](https://github.com/rust-vmm/kvm-ioctls/blob/18bc6b95a86df9feaa44d8a2ad3383a1f0d179eb/src/ioctls/vcpu.rs#L1433) on aarch64/x86?

Asking because I am curious how would that be used and also because our CI is complaining about coverage :D

xuyandong2: Ok，I will try to add test unit of this exit code.

crab2313:
@xuyandong2 Hello, any progress here? I'm going to make a new PR about this feature with a test case for aarch64.

acatangiu:
Closed in favor of #91

PR #93, opened 2020-05-01
Add support for the KVM_KVMCLOCK_CTRL ioctl

jabedude:
Hello,

I added the KVM_KVMCLOCK_CTRL ioctl. This would be useful for implementing suspend/resume in a VMM. 

I left this as a draft because I'm not sure what the best way to test this API would be. Thoughts?

### DIFF
--- a/src/ioctls/vcpu.rs
+++ b/src/ioctls/vcpu.rs
@@ -1009,6 +1009,18 @@ impl VcpuFd {
         Ok(reg_value)
     }
 
+    /// Signals to the host that the running VCPU is being paused by userspace.
+    ///
+    /// Supported on architectures that implement a paravirt clock (currently x86 only)
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    pub fn kvmclock_ctl(&self) -> Result<()> {
+        let ret = unsafe { ioctl(self, KVM_KVMCLOCK_CTRL()) };
+        if ret != 0 {
+            return Err(errno::Error::last());
+        }
+        Ok(())
+    }
+
     /// Triggers the running of the current virtual CPU returning an exit reason.
     ///
     /// See documentation for `KVM_RUN`.
--- a/src/ioctls/vcpu.rs
+++ b/src/ioctls/vcpu.rs
@@ -1222,6 +1222,18 @@ mod tests {
     ))]
     use Cap;
 
+    #[repr(C)]
+    struct pvclock_vcpu_time_info {
+        version: u32,
+        pad0: u32,
+        tsc_timestamp: u64,
+        system_time: u64,
+        tsc_to_system_mul: u32,
+        tsc_shift: i8,
+        flags: u8,
+        pad: [u8; 2],
+    }
+
     // Helper function for memory mapping `size` bytes of anonymous memory.
     // Panics if the mmap fails.
     fn mmap_anonymous(size: usize) -> *mut u8 {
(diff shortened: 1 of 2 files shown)


andreeaflorescu:
> I left this as a draft because I'm not sure what the best way to test this API would be. Thoughts?

Not sure we can test this with the current test infrastructure. We would need some sort of integration test for that, that does something more useful than some dummy MMIO writes (which is basically the most complex test case we have so far).

We could add a "negative" unit test I guess. I assume that if the vcpus are not running, then calling `KVM_KVMCLOCK_CTRL` returns an error? We could check that scenario. It isn't super helpful, but I don't really have a better idea.

alxiord:
We have tests that run simple guest workloads. We could mark a vcpu as paused in one such workload, provided that control can be passed back to the "vmm". A hack would be to hijack an MMIO exit and mark the vcpu paused in the exit handler. A somewhat cleaner way would be to mimic a real life pause scenario and interrupt the vcpu thread with a signal, *then* call this ioctl.
These may work and cover the code, but this ioctl afaik is useful after long time jumps between vcpu pause and resume, so a test that truly shows that it works should take this into account, and as @andreeaflorescu said that's not feasible with the infra in place.

jabedude:
I went ahead with the MMIO exit method and here's the status after doing some research:

Right now kvmclock_ctl is failing with EINVAL. I looked at the documentation and the kernel source to find what causes EINVAL in this ioctl (latest 5.7.0-rc2). EINVAL is returned if the pvclock is not enabled for the given VCPU. The pvclock is enabled when the MSR_KVM_SYSTEM_TIME MSR is set. Setting this MSR requires that the data in the MSR entry is a "4-byte aligned physical address of a memory area which must be in guest RAM, plus an enable bit in bit 0. This memory is expected to hold a copy of the following structure:"

```
struct pvclock_vcpu_time_info {
      u32   version;
      u32   pad0;
      u64   tsc_timestamp;
      u64   system_time;
      u32   tsc_to_system_mul;
      s8    tsc_shift;
      u8    flags;
      u8    pad[2];
} __attribute__((__packed__)); /* 32 bytes */
```
(I couldn't find this type defined in rust-vmm/kvm-bindings)

After placing this data in the guest physical memory (immediately after the code at 0x1024), I set the KVM_SYSTEM_TIME MSR but the kvmclock_ctl call ioctl still fails with EINVAL. 

I'm guessing that either:

1. I'm not initializing the pvclock struct correctly
2. I've gone down a completely wrong rabbit hole :)

andreeaflorescu:
> I went ahead with the MMIO exit method and here's the status after doing some research:
> 
> Right now kvmclock_ctl is failing with EINVAL. I looked at the documentation and the kernel source to find what causes EINVAL in this ioctl (latest 5.7.0-rc2). EINVAL is returned if the pvclock is not enabled for the given VCPU. The pvclock is enabled when the MSR_KVM_SYSTEM_TIME MSR is set. Setting this MSR requires that the data in the MSR entry is a "4-byte aligned physical address of a memory area which must be in guest RAM, plus an enable bit in bit 0. This memory is expected to hold a copy of the following structure:"
> 

I've been done the same rabbit hole in my try to test this somehow :laughing: 

The interesting thing is that it seems to be working on the CI machine because we don't get the EINVAL error there. Besides the coverage (which fails for different reasons), all tests pass. On my machine though it also fails. So maybe it is related to the kernel version?

The coverage seems to drop quite a lot for just a few lines (0.90%). I do not know the effects of that ioctl on the guest, but is it possible that because of calling `kvmclock_ctl` when processing MMIO write, the MMIO read and htl instructions are no longer executed? This could potentially justify the drop in coverage.

jabedude:
> The interesting thing is that it seems to be working on the CI machine because we don't get the EINVAL error there. Besides the coverage (which fails for different reasons), all tests pass. On my machine though it also fails. So maybe it is related to the kernel version?

Yes, I think that could definitely be a part of it.

> The coverage seems to drop quite a lot for just a few lines (0.90%). I do not know the effects of that ioctl on the guest, but is it possible that because of calling kvmclock_ctl when processing MMIO write, the MMIO read and htl instructions are no longer executed? This could potentially justify the drop in coverage.

I tested this out by not `.unwrap()`-ing the kvmclock_ctl ioctl and confirmed that the HLT is still issued. I'm really not sure why coverage dropped when I exercised the new functionality :(

andreeaflorescu:
The functionality was merged as part of #113 

Closing this.

PR #97, opened 2020-05-27
Add support to flexibly configure AArch64 IPA size of the guest

MrXinWang:
As `KVM_CREATE_VM` ioctl supports configuration of AArch64 IPA size of the guest after kernel v4.20, this commit adds support to flexibly configure AArch64 IPA size of the guest in rust-vmm.

Signed-off-by: Henry Wang <<Henry.Wang@arm.com>>

### DIFF
--- a/src/ioctls/system.rs
+++ b/src/ioctls/system.rs
@@ -12,6 +12,8 @@ use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
 use cap::Cap;
 use ioctls::vm::{new_vmfd, VmFd};
 use ioctls::Result;
+#[cfg(any(target_arch = "aarch64"))]
+use kvm_bindings::KVM_VM_TYPE_ARM_IPA_SIZE_MASK;
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
 use kvm_bindings::{CpuId, MsrList, KVM_MAX_MSR_ENTRIES};
 use kvm_ioctls::*;
@@ -374,6 +376,58 @@ impl Kvm {
         }
     }
 
+    /// AArch64 specific function to create a VM fd using the KVM fd with flexible IPA size.
+    ///
+    /// See the arm64 section of KVM documentation for `KVM_CREATE_VM`.
+    /// A call to this function will also initialize the size of the vcpu mmap area using the
+    /// `KVM_GET_VCPU_MMAP_SIZE` ioctl.
+    ///
+    /// Note: `Cap::ArmVmIPASize` should be checked using `check_extension` before calling
+    /// this function to determine if the host machine supports the IPA size capability.
+    ///
+    /// # Arguments
+    ///
+    /// * `ipa_size` - Guest VM IPA size, 32 <= ipa_size <= Host_IPA_Limit.
+    ///                The value of `Host_IPA_Limit` may be different between hardware
+    ///                implementations and can be extracted by calling `get_host_ipa_limit`.
+    ///                Possible values can be found in documentation of registers `TCR_EL2`
+    ///                and `VTCR_EL2`.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use kvm_ioctls::{Kvm, Cap};
+    /// let kvm = Kvm::new().unwrap();
+    /// // Check if the ArmVmIPASize cap is supported.
+    /// if kvm.check_extension(Cap::ArmVmIPASize) {
+    ///     let host_ipa_limit = kvm.get_host_ipa_limit();
+    ///     let vm = kvm.create_vm_with_ipa_size(host_ipa_limit as u32).unwrap();
+    ///     // Check that the VM mmap size is the same reported by `KVM_GET_VCPU_MMAP_SIZE`.
+    ///     assert!(vm.run_size() == kvm.get_vcpu_mmap_size().unwrap());
+    /// }
+    /// ```
+    ///
+    #[cfg(any(target_arch = "aarch64"))]
+    pub fn create_vm_with_ipa_size(&self, ipa_size: u32) -> Result<VmFd> {
+        // Safe because we know `self.kvm` is a real KVM fd as this module is the only one that
+        // create Kvm objects.
+        let ret = unsafe {
+            ioctl_with_val(
+                &self.kvm,
+                KVM_CREATE_VM(),
+                (ipa_size & KVM_VM_TYPE_ARM_IPA_SIZE_MASK).into(),
+            )
+        };
+        if ret >= 0 {
+            // Safe because we verify the value of ret and we are the owners of the fd.
+            let vm_file = unsafe { File::from_raw_fd(ret) };
+            let run_mmap_size = self.get_vcpu_mmap_size()?;
+            Ok(new_vmfd(vm_file, run_mmap_size))
+        } else {
+            Err(errno::Error::last())
+        }
+    }
+
     /// Creates a VmFd object from a VM RawFd.
     ///
     /// This function is unsafe as the primitives currently returned have the contract that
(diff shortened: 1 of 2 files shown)


alxiord (dismissed):

MrXinWang:
Hi @andreeaflorescu, any comments about this? Thanks

src/ioctls/system.rs
```diff
@@ -374,6 +376,46 @@ impl Kvm {
         }
     }
 
+    /// AArch64 specific function to create a VM fd using the KVM fd with flexible IPA size.
+    ///
+    /// See the arm64 section of KVM documentation for `KVM_CREATE_VM`.
+    /// A call to this function will also initialize the size of the vcpu mmap area using the
+    /// `KVM_GET_VCPU_MMAP_SIZE` ioctl.
```
andreeaflorescu, line 383: Can you also add an `# Arguments` section and specify what are the valid values for `ipa_size`?

From the kvm api documentation: " 32 <= N <= Host_IPA_Limit"

src/ioctls/system.rs
```diff
@@ -374,6 +376,46 @@ impl Kvm {
...
+    /// // Check if the ArmVmIPASize cap is supported.
+    /// if kvm.check_extension(Cap::ArmVmIPASize) {
+    ///     let vm = kvm.create_vm_with_ipa_size(40).unwrap();
+    ///     // Check that the VM mmap size is the same reported by `KVM_GET_VCPU_MMAP_SIZE`.
+    ///     assert!(vm.run_size() == kvm.get_vcpu_mmap_size().unwrap());
+    /// }
+    /// ```
+    ///
+    #[cfg(any(target_arch = "aarch64"))]
+    pub fn create_vm_with_ipa_size(&self, ipa_size: u32) -> Result<VmFd> {
```
andreeaflorescu, line 411: nit: Besides the ioctl call, these function looks pretty much like `create_vm`. It would be nice to not duplicate the code if possible. Since they're only two functions for now, I wouldn't stress to much on it though.

src/ioctls/system.rs
```diff
@@ -374,6 +376,46 @@ impl Kvm {
         }
     }
 
+    /// AArch64 specific function to create a VM fd using the KVM fd with flexible IPA size.
```
andreeaflorescu, line 379: Should we also specify that `Cap::ArmVmIPASize` should be checked before calling this function?

src/ioctls/system.rs
```diff
@@ -374,6 +376,46 @@ impl Kvm {
...
+    /// `KVM_GET_VCPU_MMAP_SIZE` ioctl.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use kvm_ioctls::{Kvm, Cap};
+    /// let kvm = Kvm::new().unwrap();
+    /// // Check if the ArmVmIPASize cap is supported.
+    /// if kvm.check_extension(Cap::ArmVmIPASize) {
+    ///     let vm = kvm.create_vm_with_ipa_size(40).unwrap();
```
andreeaflorescu: Should we use another value here? It looks like 40 is the default.

src/ioctls/system.rs
```diff
@@ -462,6 +504,33 @@ mod tests {
...
+    fn test_create_vm_with_ipa_size() {
+        let kvm = Kvm::new().unwrap();
+        if kvm.check_extension(Cap::ArmVmIPASize) {
+            let vm = kvm.create_vm_with_ipa_size(40).unwrap();
+
+            // Test create_vmfd_from_rawfd()
+            let rawfd = unsafe { libc::dup(vm.as_raw_fd()) };
+            assert!(rawfd >= 0);
+            let vm = unsafe { kvm.create_vmfd_from_rawfd(rawfd).unwrap() };
+            assert_eq!(vm.run_size(), kvm.get_vcpu_mmap_size().unwrap());
```
andreeaflorescu: What are we trying to check here?

src/ioctls/system.rs
```diff
@@ -462,6 +504,33 @@ mod tests {
         assert_eq!(vm.run_size(), kvm.get_vcpu_mmap_size().unwrap());
     }
 
+    #[test]
+    #[cfg(any(target_arch = "aarch64"))]
+    fn test_create_vm_with_ipa_size() {
+        let kvm = Kvm::new().unwrap();
+        if kvm.check_extension(Cap::ArmVmIPASize) {
+            let vm = kvm.create_vm_with_ipa_size(40).unwrap();
```
andreeaflorescu: Are there other valid values that we can use? 40 seems to be the default.

andreeaflorescu:
@MrXinWang sorry about the late review, I somehow missed this PR.

src/ioctls/system.rs
MrXinWang: Yes. I have added this part. Thanks for reminding.

src/ioctls/system.rs
MrXinWang: Done (also rebased to the current master).

src/ioctls/system.rs
MrXinWang: hmmm yes. I personally don't have a special preference. Actually the reason why I used this default value is that I found it very difficult to choose a proper value, as different arm servers might have a different implementation of the `Host_IPA_Size`. Therefore the guest IPA is limited. 

The registers [TCR_EL2](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0488c/CIHJBCHI.html) and [VTCR_EL2](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0488d/CIHEDIII.html) indicate that the possible values of `Host_IPA_Size` are 32, 36, 40, 42, 44, 48 (I tested on 6 machines in my team and got 42, 44, 48). Any suggestions about this?

src/ioctls/system.rs
MrXinWang: Same above :)

src/ioctls/system.rs
MrXinWang: This is actually my original thought. However I was a little bit worried that this might be a breaking change, as the interface should be changed. If we add another parameter called `ipa_size` to `create_vm`, it might break anything which used `create_vm` function. Although we probably can give it a default value, this is still not ideal and risky...Therefore I just used a safer way, and anyone needs to create a VM with flexible IPA size can call `create_vm_with_ipa_size` instead of `create_vm`.

src/ioctls/system.rs
MrXinWang: Same thing in `create_vm` (see [here](https://github.com/rust-vmm/kvm-ioctls/blob/1072239e47c6b5dc6e143be8712ecde4ecc1347e/src/ioctls/system.rs#L457)), namely `create_vmfd_from_rawfd()`. As this `create_vm_with_ipa_size` is an alternative function of `create_vm`, I personally think anything is covered in `create_vm` test should also be covered in `create_vm_with_ipa_size` test.

andreeaflorescu: I don't think it is necessarily needed.

In that test we were checking that the `create_vmfd_from_rawfd` function works when the file descriptor passed is valid. Since this is already tested, we don't need to double check it now.

MrXinWang: Sure! I will remove this part after all your comments are addressed. Thanks.

src/ioctls/system.rs
andreeaflorescu: Hm, let's stick with 40 then, but we should add a comment that this is the default value. I think it would be useful to also say one line about the registers as well.

MrXinWang: Thanks @andreeaflorescu, all done. Please take a look.

rhdrjones: Why not test with the maximum that the host supports? That would both test the discoverability of supported IPA sizes and it would likely test with something other than 40.

andreeaflorescu (dismissed):

src/ioctls/system.rs
MrXinWang: Yeah actually that makes a lot of sense...@andreeaflorescu what do you think? I can change to this way.

andreeaflorescu: Yap, great idea!

MrXinWang: Thanks for the suggestion @rhdrjones , I updated the code and would you mind taking another look?

@andreeaflorescu Sorry to keep you busy but I think you have to approve again :P

rhdrjones: wrt the test below, it looks good to me. Thanks

andreeaflorescu: > @andreeaflorescu Sorry to keep you busy but I think you have to approve again :P

Don't worry about it. It's my job to maintain this crate. When I don't have time, I might end up ghosting people (by mistake) :))

andreeaflorescu: @MrXinWang should we also update the code example to have the ipa_size = host_ipa_size?

MrXinWang: Done. @andreeaflorescu  Thought in the code example we can have any values instead of being as strict as in the unit test.

andreeaflorescu (approved):

src/ioctls/system.rs
andreeaflorescu: The code examples are also run in the CI, but it should be fine.

alxiord (approved):

PR #100, opened 2020-06-23
Add support for `KVM_GET_DEVICE_ATTR` ioctl

### LINKED ISSUE #99: Supporting `KVM_GET_DEVICE_ATTR` ioctl
Hi guys, I am wondering if we are missing supporting this `KVM_GET_DEVICE_ATTR` ioctl (doc see [here](https://github.com/torvalds/linux/blob/master/Documentation/virt/kvm/api.rst#480-kvm_set_device_attrkvm_get_device_attr)) or it is me understanding it incorrectly... Currently I can only find `KVM_SET_DEVICE_ATTR` and `KVM_HAS_DEVICE_ATTR` in master branch rust-vmm.

As far as I know, using this ioctl is the only way to extract the vGIC states on AArch64.

@andreeaflorescu @michael2012z could you please share any ideas?

MrXinWang:
The `KVM_GET_DEVICE_ATTR` ioctl is useful when we need to extract the state/information of devices in the VM. In AArch64 VMs, using this ioctl is the only method to get the vGIC states.

This commit implements the `KVM_GET_DEVICE_ATTR` ioctl with its unit test on AArch64.

Fixes: #99 

Signed-off-by: Henry Wang <Henry.Wang@arm.com>

### DIFF
--- a/coverage_config_x86_64.json
+++ b/coverage_config_x86_64.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 91.4,
+  "coverage_score": 91.3,
   "exclude_path": "",
   "crate_features": ""
 }
--- a/src/ioctls/device.rs
+++ b/src/ioctls/device.rs
@@ -6,9 +6,9 @@ use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
 
 use ioctls::Result;
 use kvm_bindings::kvm_device_attr;
-use kvm_ioctls::{KVM_HAS_DEVICE_ATTR, KVM_SET_DEVICE_ATTR};
+use kvm_ioctls::{KVM_GET_DEVICE_ATTR, KVM_HAS_DEVICE_ATTR, KVM_SET_DEVICE_ATTR};
 use vmm_sys_util::errno;
-use vmm_sys_util::ioctl::ioctl_with_ref;
+use vmm_sys_util::ioctl::{ioctl_with_mut_ref, ioctl_with_ref};
 
 /// Wrapper over the file descriptor obtained when creating an emulated device in the kernel.
 pub struct DeviceFd {
@@ -75,6 +75,76 @@ impl DeviceFd {
     ///
     pub fn set_device_attr(&self, device_attr: &kvm_device_attr) -> Result<()> {
         let ret = unsafe { ioctl_with_ref(self, KVM_SET_DEVICE_ATTR(), device_attr) };
+        if ret != 0 {
+            return Err(errno::Error::last());
+        }
+        Ok(())
+    }
+
+    /// Gets a specified piece of device configuration and/or state.
+    ///
+    /// See the documentation for `KVM_GET_DEVICE_ATTR`.
+    ///
+    /// # Arguments
+    ///
+    /// * `device_attr` - The device attribute to be get.
+    ///                   Note: This argument serves as both input and output.
+    ///                   When calling this function, the user should explicitly provide
+    ///                   valid values for the `group` and the `attr` field of the
+    ///                   `kvm_device_attr` structure, and a valid userspace address
+    ///                   (i.e. the `addr` field) to access the returned device attribute
+    ///                   data.
+    ///
+    /// # Returns
+    ///
+    /// * Returns the last occured `errno` wrapped in an `Err`.
+    /// * `device_attr` - The `addr` field of the `device_attr` structure will point to
+    ///                   the device attribute data.
+    ///
+    /// # Examples
+    /// ```rust
+    /// # extern crate kvm_ioctls;
+    /// # extern crate kvm_bindings;
+    /// # use kvm_ioctls::Kvm;
+    ///
+    /// let kvm = Kvm::new().unwrap();
+    /// let vm = kvm.create_vm().unwrap();
+    ///
+    /// // As on x86_64, `get_device_attr` is not necessarily needed. Therefore here
+    /// // the code example is only for AArch64.
+    /// #[cfg(any(target_arch = "aarch64"))]
+    /// {
+    ///     use kvm_bindings::{
+    ///         KVM_DEV_ARM_VGIC_GRP_NR_IRQS, kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V2,
+    ///         kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V3,
+    ///     };
+    ///
+    ///     // Create a GIC device.
+    ///     let mut gic_device = kvm_bindings::kvm_create_device {
+    ///         type_: kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V3,
+    ///         fd: 0,
+    ///         flags: 0,
+    ///     };
+    ///     let device_fd = match vm.create_device(&mut gic_device) {
+    ///        Ok(fd) => fd,
+    ///        Err(_) => {
+    ///        gic_device.type_ = kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V2;
+    ///        vm.create_device(&mut gic_device)
+    ///            .expect("Cannot create KVM vGIC device")
+    ///        }
+    ///     };
+    ///
+    ///     let mut data: u32 = 0;
+    ///     let mut gic_attr = kvm_bindings::kvm_device_attr::default();
+    ///     gic_attr.group = KVM_DEV_ARM_VGIC_GRP_NR_IRQS;
+    ///     gic_attr.addr = &mut data as *const u32 as u64;
+    ///
+    ///     device_fd.get_device_attr(&mut gic_attr).unwrap();
+    /// }
+    /// ```
+    ///
+    pub fn get_device_attr(&self, device_attr: &mut kvm_device_attr) -> Result<()> {
+        let ret = unsafe { ioctl_with_mut_ref(self, KVM_GET_DEVICE_ATTR(), device_attr) };
         if ret != 0 {
             return Err(errno::Error::last());
         }
@@ -155,9 +225,12 @@ mod tests {
             flags: 0,
         };
 
+        let mut dist_attr_mut = dist_attr;
+
         // We are just creating a test device. Creating a real device would make the CI dependent
         // on host configuration (like having /dev/vfio). We expect this to fail.
         assert!(device_fd.has_device_attr(&dist_attr).is_err());
+        assert!(device_fd.get_device_attr(&mut dist_attr_mut).is_err());
         assert!(device_fd.set_device_attr(&dist_attr).is_err());
         assert_eq!(errno::Error::last().errno(), 25);
     }
@@ -165,8 +238,11 @@ mod tests {
     #[test]
     #[cfg(target_arch = "aarch64")]
     fn test_create_device() {
-        use ioctls::vm::create_gic_device;
-        use kvm_bindings::kvm_device_type_KVM_DEV_TYPE_FSL_MPIC_20;
+        use ioctls::vm::{create_gic_device, set_supported_nr_irqs};
+        use kvm_bindings::{
+            kvm_device_type_KVM_DEV_TYPE_FSL_MPIC_20, KVM_DEV_ARM_VGIC_GRP_NR_IRQS,
+        };
+        use vmm_sys_util::errno::Error;
 
         let kvm = Kvm::new().unwrap();
         let vm = kvm.create_vm().unwrap();
@@ -197,6 +273,9 @@ mod tests {
         };
         assert!(device_fd.has_device_attr(&dist_attr).is_err());
 
+        // Set maximum supported number of IRQs of the vGIC device to 128.
+        set_supported_nr_irqs(&device_fd, 128);
+
         // Following attribute works with VGIC, they should be accepted.
         let dist_attr = kvm_bindings::kvm_device_attr {
             group: KVM_DEV_ARM_VGIC_GRP_CTRL,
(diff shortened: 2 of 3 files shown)


src/ioctls/device.rs
```diff
@@ -80,6 +80,21 @@ impl DeviceFd {
...
     }
+
+    /// Gets a specified piece of device configuration and/or state.
+    ///
+    /// See the documentation for `KVM_GET_DEVICE_ATTR`.
+    /// # Arguments
+    ///
+    /// * `device_attr` - The device attribute to be get.
+    ///
+    pub fn get_device_attr(&self, device_attr: &mut kvm_device_attr) -> Result<()> {
```
andreeaflorescu, line 146: Can we return the `kvm_device_attr` instead of receiving it as a parameter? Function parameters that act as output are to be avoided unless there is a good reason to have them.

andreeaflorescu (changes requested):

andreeaflorescu:
Please also update the commit message and add the following line:
`Fixes: https://github.com/rust-vmm/kvm-ioctls/issues/99`

src/ioctls/device.rs
MrXinWang: Hi @andreeaflorescu. It makes sense but I am not sure if you can accept the following idea:

As far as I know, this ioctl works in this way: users put the `group` and `attr` into this `kvm_device_attr` structure as "requests", send it to the `KVM_GET_DEVICE_ATTR` ioctl, and the ioctl will return the information in a pointer which points to the "user-requested" data in the `addr` field.

In this way, the `kvm_device_attr` structure is both part of the input and part of the output. Maybe we can only return the `addr` field? I personally don't have a specific preference.

coverage_config_x86_64.json
```diff
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 91.4,
+  "coverage_score": 90.3,
```
dianpopa: I am guessing that if we add a `assert!(device_fd.get_device_attr(&dist_attr).is_err());` [here](https://github.com/MrXinWang/kvm-ioctls/blob/kvm_get_device_attr/src/ioctls/device.rs#L176), this will not be needed anymore.

src/ioctls/device.rs
```diff
@@ -207,5 +227,18 @@ mod tests {
...
         assert!(device_fd.has_device_attr(&dist_attr).is_ok());
         assert!(device_fd.set_device_attr(&dist_attr).is_ok());
+
+        // Test `get_device_attr`. Here we try to extract the maximum supported number of IRQs.
+        // This value should be saved in the address provided to the ioctl.
+        let mut data: u32 = 0;
+
+        let mut gic_attr_get = kvm_bindings::kvm_device_attr::default();
+        gic_attr_get.group = KVM_DEV_ARM_VGIC_GRP_NR_IRQS;
+        gic_attr_get.addr = &mut data as *const u32 as u64;
```
dianpopa: I would also exercise the error path for this. For example, `EFAULT` will be returned if the attribute points to an invalid address. So, I'd add this before checking the success case:
```
gic_attr_get.addr = data as u64;
// Without properly providing the address to where the 
// value will be stored, the ioctl fails with EFAULT.
let res = device_fd.get_device_attr(&mut gic_attr_get);
assert!(res.is_err() && errno::Error::last().errno() == 14);

gic_attr_get.addr = &mut data as *const u32 as u64;
...
...
```
Expect that coverage may need adjustments .

src/ioctls/device.rs
dianpopa: You are right @MrXinWang. The user needs to provide `group`, `attr` and a valid address (i.e `addr` field) otherwise the ioctl will fail.
As a side note, in many places (qemu included) KVM_GET_ATTR AND KVM_SET_ATTR are called through the same function.

MrXinWang: > As a side note, in many places (qemu included) KVM_GET_ATTR AND KVM_SET_ATTR are called through the same function.

Exactly, they used a bool parameter to determine whether it is read or write.

coverage_config_x86_64.json
MrXinWang: Thanks! Updated.

src/ioctls/device.rs
MrXinWang: Ahh yes yes this is a really good example. Thanks for proposing this! Updated.

MrXinWang:
> Please also update the commit message and add the following line:
> `Fixes: https://github.com/rust-vmm/kvm-ioctls/issues/99`

Thanks for mentioning this! Done :)

dianpopa (dismissed):

coverage_config_x86_64.json
MrXinWang: Sorry @dianpopa , this coverage varies 0.1% after the update (see https://buildkite.com/rust-vmm/kvm-ioctls-ci/builds/489#bce57505-df0d-4032-8f47-355c8a36bab2/38-63)...

Now it should be fine.

dianpopa (dismissed):

src/ioctls/device.rs
andreeaflorescu: Oh, I see.

The point would be to make this function (and every other one for that matter) easy to use, and hard to misuse. When you have parameters that work both as input and as output, that's hard to achieve.

Since it seems like some of the fields are used for input and some are for output, can you make the documentation more explicit? Also, a code example should be very useful. Most of the other functions have one. For the code example you don't need to do asserts (or use SET_DEVICE_ATTR). Just showcasing how you use the parameter as an input and as output should suffice.

src/ioctls/device.rs
```diff
@@ -80,6 +80,21 @@ impl DeviceFd {
         }
         Ok(())
     }
+
+    /// Gets a specified piece of device configuration and/or state.
+    ///
+    /// See the documentation for `KVM_GET_DEVICE_ATTR`.
+    /// # Arguments
```
andreeaflorescu, line 88: Also add a `# Returns` section so we can add details such as this ioctl returns its result in the input parameter (and what fields should be used as return).

MrXinWang: Done.

src/ioctls/device.rs
MrXinWang: Hi @andreeaflorescu. I added the `#Returns` and the `#Examples` section in the doc and updated some details about the input and output arguments. Would you mind taking a look again?

src/ioctls/device.rs
```diff
@@ -80,6 +80,104 @@ impl DeviceFd {
...
+    ///         .expect("Cannot create KVM VFIO device");
+    ///
+    ///     let mut dist_attr_mut = kvm_bindings::kvm_device_attr {
+    ///         group: KVM_DEV_VFIO_GROUP,
+    ///         attr: u64::from(KVM_DEV_VFIO_GROUP_ADD),
+    ///         addr: 0x0,
+    ///         flags: 0,
+    ///     };
+    ///
+    ///     // On x86_64, getting device attributes from `KVM_DEV_VFIO_GROUP`
```
andreeaflorescu: In this case we shouldn't have an example with it.
Can you please remove it so it is less verbose?

The point of the examples is to show people how to use the function. It seems like on x86_64 you don't necessarily need it.

src/ioctls/device.rs
```diff
@@ -80,6 +80,104 @@ impl DeviceFd {
...
+    ///                   When calling this function, the user should explicitly provide
+    ///                   valid values for the `group` and the `attr` field of the
+    ///                   `kvm_device_attr` structure, and a valid userspace address
+    ///                   (i.e. the `addr` field) to access the returned device attribute
+    ///                   data.
+    ///
+    /// # Returns
+    ///
+    /// * Returns the last occured `errno` wrapped in an `Err`.
+    ///
```
andreeaflorescu: nit: remove this whitespace.

src/ioctls/device.rs
```diff
@@ -80,6 +80,104 @@ impl DeviceFd {
         }
         Ok(())
     }
+
+    /// Gets a specified piece of device configuration and/or state.
+    ///
+    /// See the documentation for `KVM_GET_DEVICE_ATTR`.
```
andreeaflorescu, line 86: nit: add whitespace before each section.

src/ioctls/device.rs
```diff
@@ -80,6 +80,104 @@ impl DeviceFd {
...
+    ///            .expect("Cannot create KVM vGIC device")
+    ///        }
+    ///     };
+    ///
+    ///     let mut data: u32 = 0;
+    ///     let mut gic_attr_get = kvm_bindings::kvm_device_attr::default();
+    ///     gic_attr_get.group = KVM_DEV_ARM_VGIC_GRP_NR_IRQS;
+    ///     gic_attr_get.addr = &mut data as *const u32 as u64;
+    ///
+    ///     assert!(device_fd.get_device_attr(&mut gic_attr_get).is_ok());
```
andreeaflorescu: Can you please replace the assert with `unwrap`? Asserts obfuscate the code a bit and the code does not flow naturally anymore.

src/ioctls/device.rs
```diff
@@ -80,6 +80,104 @@ impl DeviceFd {
...
+    ///        Ok(fd) => fd,
+    ///        Err(_) => {
+    ///        gic_device.type_ = kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V2;
+    ///        vm.create_device(&mut gic_device)
+    ///            .expect("Cannot create KVM vGIC device")
+    ///        }
+    ///     };
+    ///
+    ///     let mut data: u32 = 0;
+    ///     let mut gic_attr_get = kvm_bindings::kvm_device_attr::default();
```
andreeaflorescu: nit: objects should have substantives as names. This one should probably be named `gic_attr`.

src/ioctls/device.rs
```diff
@@ -207,5 +313,24 @@ mod tests {
 
         assert!(device_fd.has_device_attr(&dist_attr).is_ok());
         assert!(device_fd.set_device_attr(&dist_attr).is_ok());
+
+        // Test `get_device_attr`. Here we try to extract the maximum supported number of IRQs.
+        // This value should be saved in the address provided to the ioctl.
+        let mut data: u32 = 0;
+
+        let mut gic_attr_get = kvm_bindings::kvm_device_attr::default();
```
andreeaflorescu: This should be named `gic_attr`.

src/ioctls/device.rs
```diff
@@ -207,5 +313,24 @@ mod tests {
...
+        gic_attr_get.group = KVM_DEV_ARM_VGIC_GRP_NR_IRQS;
+        gic_attr_get.addr = data as u64;
+
+        // Without properly providing the address to where the
+        // value will be stored, the ioctl fails with EFAULT.
+        let res = device_fd.get_device_attr(&mut gic_attr_get);
+        assert!(res.is_err() && errno::Error::last().errno() == 14);
```
andreeaflorescu: This should assert on `res`, not on `errno::Error::last()`.

More of a nit, but can you use `libc::EFAULT` instead of magic numbers (i.e. 14)?

MrXinWang: I think what @dianpopa wants to express here is the ioctl will fail AND the error number in this case (i.e. without properly providing the address to where the value will be stored) should be a fixed number 14 (`EFAULT`). Hence we assert both of them by using `&&` here.

The error number has been changed to `libc::EFAULT` :)

src/ioctls/device.rs
MrXinWang: Done

src/ioctls/device.rs
MrXinWang: Done

src/ioctls/device.rs
MrXinWang: Done

src/ioctls/device.rs
MrXinWang: Done

src/ioctls/device.rs
MrXinWang: Done

src/ioctls/device.rs
MrXinWang: Done

src/ioctls/device.rs
```diff
@@ -207,5 +285,24 @@ mod tests {
...
+        let mut data: u32 = 0;
+
+        let mut gic_attr = kvm_bindings::kvm_device_attr::default();
+        gic_attr.group = KVM_DEV_ARM_VGIC_GRP_NR_IRQS;
+        gic_attr.addr = data as u64;
+
+        // Without properly providing the address to where the
+        // value will be stored, the ioctl fails with EFAULT.
+        let res = device_fd.get_device_attr(&mut gic_attr);
+        assert!(res.is_err() && errno::Error::last().errno() == libc::EFAULT);
```
andreeaflorescu: `res` is already `errno::Error::last`. No need to call that again. I believe that it is actually incorrect to call it again. If you add any ioctl between line 299 and line 300 than errno will be the error of that last call and not the result from `get_device_attr`. If this happens in another part of the kvm, I can open an issue with that.

MrXinWang: Thanks @andreeaflorescu for explaining! Then I think I should remove the assertion of error code here and only leave the assertion of calling this ioctl, namely `res`.

andreeaflorescu (approved):

src/ioctls/device.rs
andreeaflorescu: Sure, thanks for your PR!

dianpopa (approved):

alxiord (approved):

PR #102, opened 2020-06-30
Add capability definition for KVM_CAP_MSI_DEVID

michael2012z:
The per-VM KVM_CAP_MSI_DEVID capability advertises the requirement to
provide the device ID in kvm_irq_routing_msi.

Signed-off-by: Michael Zhao <michael.zhao@arm.com>

PR #103, opened 2020-07-07
Implement 'check_extension()' for Vm

michael2012z:
We have defined `check_extension()` in `Kvm` to perform `KVM_CHECK_EXTENSION` ioctl. While the same API is also needed in `Vm`.

One example of such extensions is `KVM_CAP_MSI_DEVID`. After adding Cap `KVM_CAP_MSI_DEVID` with https://github.com/rust-vmm/kvm-ioctls/pull/102, I realized it need to be checked against `Vm` rather than `Kvm`, but we haven't defined the API to do it.

### DIFF
--- a/src/ioctls/vm.rs
+++ b/src/ioctls/vm.rs
@@ -11,6 +11,7 @@ use std::os::raw::c_void;
 use std::os::raw::{c_int, c_ulong};
 use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
 
+use cap::Cap;
 use ioctls::device::new_device;
 use ioctls::device::DeviceFd;
 use ioctls::vcpu::new_vcpu;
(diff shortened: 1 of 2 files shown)


src/ioctls/vm.rs
```diff
@@ -1216,6 +1217,40 @@ impl VmFd {
     pub fn run_size(&self) -> usize {
         self.run_size
     }
+
+    /// Wrapper over `KVM_CHECK_EXTENSION`.
+    ///
+    /// Returns 0 if the capability is not available and a positive integer otherwise.
+    fn check_extension_int(&self, c: Cap) -> i32 {
```
michael2012z, line 1224: I reused the code in system.rs.
Now `check_extension()` is the only caller of this function. Defining it is mainly for possible future extension.

andreeaflorescu, line 1224: Are there no cases where you would need to get the result of calling `KVM_CHECK_EXTENSION` rather than just knowing it is available or not?

One example for KVM capabilities where you need to know the actual value returned by KVM_CHECK_EXTENSION is getting the kvm api number. The return of that ioctl should be 12 and you actually need to check that is 12, not that it is higher than 0.

My question comes from trying to understand if we need to make `check_extention_int` public instead of solely relying on `check_extension`.

src/ioctls/vm.rs
```diff
@@ -1751,4 +1786,11 @@ mod tests {
         let irq_routing = kvm_irq_routing::default();
         assert!(vm.set_gsi_routing(&irq_routing).is_ok());
     }
+
+    #[test]
+    fn test_check_extension() {
+        let kvm = Kvm::new().unwrap();
+        let vm = kvm.create_vm().unwrap();
+        assert!(vm.check_extension(Cap::MpState));
```
michael2012z, line 1794: MpState seems a Cap that constantly tested true per-Vm on both Arm and X86.

src/ioctls/vm.rs
michael2012z: Looking into kernel code handling `KVM_CHECK_EXTENSION` (ARM only): https://github.com/torvalds/linux/blob/v5.4/virt/kvm/kvm_main.c#L3198 and https://github.com/torvalds/linux/blob/v5.4/virt/kvm/arm/arm.c#L180, I didn't find any numerical extension that have to be checked per VM. 

So I think we don't need to expose it public, now.

michael2012z:
@andreeaflorescu, anything I should improve? :)

andreeaflorescu (approved):

rbradford (approved):

Issue #28, opened 2019-04-11
Add unit test for `set_cpuid2`


andreeaflorescu:
We have no unit test for testing the `set_cpuid2` method. Right now it is tested with the code example but that's not an ideal state.

wey-gu:
Hi, I'd like to look into this and help ASAP in my free time.
The `good first issue` label is very eye-catching to me :-), thanks.

andreeaflorescu:
@wey-gu cool! Let me know if you need help with anything.

PR #110, opened 2020-07-31
Some more Unit Tests for set_cpuid2

wey-gu:
To introduce more UTs for set_cpuid2
 * Setting Manufacturer ID
 * Disabling Intel SHA extensions

Issue: #28

Signed-off-by: Wey Gu <weyl.gu@gmail.com>

> This is the initial attempt for #28 , please team to help review/ advice on:
1. Whether this is the right direction?
2. Any other cases to be covered?

This is my very first PR, your reviews/ corrections are appreciated :-)

Thanks.

### DIFF
--- a/src/ioctls/vcpu.rs
+++ b/src/ioctls/vcpu.rs
@@ -1282,7 +1282,7 @@ mod tests {
 
     #[cfg(target_arch = "x86_64")]
     #[test]
-    fn cpuid2_test() {
+    fn test_get_cpuid() {
         let kvm = Kvm::new().unwrap();
         if kvm.check_extension(Cap::ExtCpuid) {
             let vm = kvm.create_vm().unwrap();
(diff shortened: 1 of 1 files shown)


andreeaflorescu:
The PR looks good. Thanks for the contribution!

Can you check the failing tests and fix the errors? They look mostly like style problems.

wey-gu:
> The PR looks good. Thanks for the contribution!
> 
> 
> 
> Can you check the failing tests and fix the errors? They look mostly like style problems.

Thanks a lot Andreea 😄. Fixed.

src/ioctls/vcpu.rs
```diff
@@ -1286,16 +1286,72 @@ mod tests {
         let kvm = Kvm::new().unwrap();
         if kvm.check_extension(Cap::ExtCpuid) {
             let vm = kvm.create_vm().unwrap();
-            let cpuid = kvm.get_supported_cpuid(KVM_MAX_CPUID_ENTRIES).unwrap();
+            let mut cpuid = kvm.get_supported_cpuid(KVM_MAX_CPUID_ENTRIES).unwrap();
             let ncpuids = cpuid.as_slice().len();
             assert!(ncpuids <= KVM_MAX_CPUID_ENTRIES);
-            let nr_vcpus = kvm.get_nr_vcpus();
-            for cpu_idx in 0..nr_vcpus {
```
andreeaflorescu: Any reason for not keeping the for loop? The cpuid should probably be set by all vcpus.

wey-gu: Dear @andreeaflorescu ,

I thought if we create `n` or `nr_vcpus` of vCPUs, tests will be running in a slower way, which duplicated the execution of `create_vcpu`, `set_cpuid2` and assert, or maybe I missed something here, we may gain something from this right?
What do you think, do you suggest we put all tests in the loop of `nr_vcpus`? I can do that if it's the case.

Thanks! :)

andreeaflorescu: We shouldn't worry too much about how much it takes to run this test. It is still probably sub-second, so it's fine.

When we are not doing the for anymore, we actually loose a few things that were tested before. For example, we do not test `kvm.get_nr_vcpus()`.

What we could do instead is leave this current test as it was before, rename it to `test_get_cpuid`, and create a new test `test_set_cpuid` in which we can only create one vcpu and set cpuid for it. Typically you will set cpuid for all the vcpus in real life scenarios, so the test will be closer to that scenario. What do you say? I am also fine with other options as long as we do not remove any existing tests.

wey-gu: Aha, got it Andreea, it used to cover both `get/ set cpuid2` and `get_nr_vcpus`, and I agree with this proposal to split both with the `get` test kept. :-)

andreeaflorescu (approved):

alxiord (approved):

PR #111, opened 2020-08-04
Enable `KVM_MEM_ENCRYPT_OP` ioctl for AMD SEV

kailun-qin:
The `KVM_MEM_ENCRYPT_OP` ioctl is used to access AMD SEV (Secure
Encrypted Virtualization) feature. When enabled, memory contents of a VM
will be transparently encrypted with a key unique to that VM.

This patch added implementation, documentation and related tests for the
following SEV commands:
- KVM_SEV_INIT
- KVM_SEV_LAUNCH_START
- KVM_SEV_LAUNCH_UPDATE_DATA
- KVM_SEV_LAUNCH_MEASURE
- KVM_SEV_LAUNCH_FINISH
- KVM_SEV_LAUNCH_SECRET

Note: this work is based on enarx/ketuvim. See [1]_ for details.

[1] https://github.com/rust-vmm/community/issues/62

Signed-off-by: Kailun Qin <kailun.qin@hotmail.com>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,19 @@ license = "Apache-2.0 OR MIT"
 libc = ">=0.2.39"
 kvm-bindings = { version = ">=0.2.0", features = ["fam-wrappers"] }
 vmm-sys-util = ">=0.2.1"
+bitflags = { version = ">=1.2.1", optional = true }
+sev = { path = "./sev", features = ["openssl"], optional = true }
 
 [dev-dependencies]
 byteorder = ">=1.2.1"
+codicon = "2.1.0"
+reqwest = "0.9.24"
+raw-cpuid = "8.1.1"
+regex = "1.3.9"
+lazy_static = "1.4.0"
+
+[features]
+default = ["amd-sev"]
+
+# We may use `target_feature` attribute when supported
+amd-sev = ["bitflags", "sev"]
--- a/src/ioctls/sev.rs
+++ b/src/ioctls/sev.rs
@@ -1,0 +1,680 @@
+#![cfg(feature = "amd-sev")]
+
+use kvm_bindings::*;
+use std::fs::{File, OpenOptions};
+use std::mem::size_of_val;
+use std::os::unix::io::{AsRawFd, RawFd};
+
+use ioctls::{vm::VmFd, Result};
+use kvm_ioctls::*;
+use sev::launch::*;
+use vmm_sys_util::errno;
+use vmm_sys_util::ioctl::ioctl_with_mut_ref;
+
+#[repr(u32)]
+#[allow(dead_code)]
+#[derive(Copy, Clone, Debug)]
+pub enum SevCmd {
+    Init = 0,
+    EsInit,
+
+    LaunchStart,
+    LaunchUpdateData,
+    LaunchUpdateVmsa,
+    LaunchSecret,
+    LaunchMeasure,
+    LaunchFinish,
+
+    SendStart,
+    SendUpdateData,
+    SendUpdateVmsa,
+    SendFinish,
+
+    ReceiveStart,
+    ReceiveUpdateData,
+    ReceiveUpdateVmsa,
+    ReceiveFinish,
+
+    GuestStatus,
+    DebugDecrypt,
+    DebugEncrypt,
+    CertExport,
+}
+
+#[derive(Copy, Clone)]
+pub struct Handle(u32);
+/// Learn from ``SevState`` introduced in QEMU, adding few changes.
+/// An enumeration of SEV state information used during @query-sev.
+///
+/// @Uninit: The guest is uninitialized.
+///
+/// @Init: The guest is initialized.
+///
+/// @LaunchUpdate: The guest is currently being launched; plaintext data and
+///                register state is being imported.
+///
+/// @LaunchSecret: The guest is currently being launched; ciphertext data
+///                is being imported.
+///
+/// @Running: The guest is fully launched or migrated in.
+///
+/// @SendUpdate: The guest is currently being migrated out to another machine.
+///
+/// @ReceiveUpdate: The guest is currently being migrated from another machine.
+#[derive(PartialEq, Debug)]
+pub struct Uninit;
+#[derive(PartialEq, Debug)]
+pub struct Init;
+pub struct LaunchUpdate(Handle);
+pub struct LaunchSecret(Handle, Measurement);
+pub struct Running(Handle, Measurement);
+#[allow(dead_code)]
+pub struct SendUpdate(Handle, Measurement);
+#[allow(dead_code)]
+pub struct ReceiveUpdate(Handle, Measurement);
+
+/// Wrapper over `/dev/sev` Fd
+pub struct SevFd {
+    sev: File,
+}
+
+impl AsRawFd for SevFd {
+    fn as_raw_fd(&self) -> RawFd {
+        self.sev.as_raw_fd()
+    }
+}
+
+/// Helper function to create a new `SevFd`.
+///
+/// This should not be exported as a public function because the preferred way is to use
+/// `new_with_vm` from `SevLaunch`.
+fn new_sevfd(sev: File) -> SevFd {
+    SevFd { sev }
+}
+
+/// KVM SEV launch contex, using a `state machine` pattern.
+pub struct SevLaunch<S> {
+    pub sev: SevFd,
+    pub vm: VmFd,
+    pub state: S,
+}
+
+impl<T> SevLaunch<T> {
+    /// Wrapper over `KVM_MEMORY_ENCRYPT_OP`.
+    ///
+    /// See documentation for `KVM_MEMORY_ENCRYPT_OP`.
+    ///
+    /// # Arguments
+    /// * `cmd` - command argument to `KVM_MEM_ENCRYPT_OP` (in). For details check the `kvm_sev_cmd`
+    ///           structure in the
+    ///           [KVM SEV API doc](https://www.kernel.org/doc/html/latest/virt/kvm/amd-memory-encryption.html).
+    /// * `data` - struct containing arguments specific to command (out).
+    ///
+    /// Returns data or an error otherwise.
+    fn sev_ioctl<U>(&self, cmd: SevCmd, mut data: U) -> Result<U> {
+        let mut input = kvm_sev_cmd {
+            id: cmd as u32,
+            data: &mut data as *mut _ as u64,
+            error: 0,
+            sev_fd: self.sev.as_raw_fd() as u32,
+        };
+
+        // The ioctl is safe because we allocated the struct and we know the
+        // kernel will write exactly the size of the struct.
+        let ret = unsafe { ioctl_with_mut_ref(&self.vm, KVM_MEMORY_ENCRYPT_OP(), &mut input) };
+        if ret == 0 {
+            Ok(data)
+        } else {
+            Err(errno::Error::last())
+        }
+    }
+}
+
+impl SevLaunch<Uninit> {
+    /// Instantiate a new SEV launch context with VmFd.
+    pub fn new_with_vm(vm: VmFd) -> Result<Self> {
+        let sev = OpenOptions::new().read(true).write(true).open("/dev/sev")?;
+
+        // Safe because we are the owners of the fd.
+        Ok(SevLaunch {
+            sev: new_sevfd(sev),
+            vm: vm,
+            state: Uninit,
+        })
+    }
+
+    /// Initialize the SEV platform context.
+    /// In a typical workflow, this command should be the first command issued.
+    pub fn init(self) -> Result<SevLaunch<Init>> {
+        self.sev_ioctl(SevCmd::Init, ())?;
+
+        Ok(SevLaunch {
+            sev: self.sev,
+            vm: self.vm,
+            state: Init,
+        })
+    }
+}
+
+impl From<SevLaunch<Uninit>> for SevLaunch<Init> {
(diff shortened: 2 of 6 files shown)


jiangliu:
This introduces too much dependency crates:(

kailun-qin:
> This introduces too much dependency crates:(

Hi @jiangliu ,

Thanks for the comments.

Most of the dependency crates introduced are dev-only and the rest for runtime are feature-controlled.

For the test-only dependencies, a large part of them aim to facilitate users with the construction of SEV cert chain (from remote). We can get rid of them if we assume that users have already had their own available locally. This depends on how we trade-off and I'd like to see more feedbacks.

Actually, the only one runtime dependency needed is the ``sev`` crate from [enarx](https://github.com/enarx/enarx) project and let's see how the enarx community suggests.

Cargo.toml
```diff
@@ -12,6 +12,19 @@ license = "Apache-2.0 OR MIT"
 libc = ">=0.2.39"
 kvm-bindings = { version = ">=0.2.0", features = ["fam-wrappers"] }
 vmm-sys-util = ">=0.2.1"
+bitflags = { version = ">=1.2.1", optional = true }
+sev = { path = "./sev", features = ["openssl"], optional = true }
```
alxiord, line 16: There's no `sev` crate, does this refer to the `sev` mod?

alxiord: ~Or did you have the `sev` crate locally in your workspace? The PR doesn't build without it.~
LE: I *think* this is the `sev` crate from the [enarx](https://github.com/enarx/enarx/tree/e044d49cb8c0e19bad8cf98b42129ff89bed1d8e/sev) repo; but, as `enarx` doesn't have a `Cargo.toml` (they seem to use a different build system, `cargo make`?), Cargo can't get the crate dependency right with
```toml
sev = { git = "https://github.com/enarx/enarx.git", features = ["openssl"], optional = true }
```
This would be a blocker in publishing a new version of `kvm-ioctls` with `sev` support: the `sev` crate needs to be published first.
In the meantime, for playing around, maybe you could get away with a multiple-dependency like explained [here](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations). Cargo would look for the `sev` crate on crates.io and, until it finds it there, pick it up from the local repo (to get the CI to pass, you'd have to yank it into your fork and add it to the PR).

Cargo.toml
```diff
@@ -12,6 +12,19 @@ license = "Apache-2.0 OR MIT"
...
 [dev-dependencies]
 byteorder = ">=1.2.1"
+codicon = "2.1.0"
+reqwest = "0.9.24"
+raw-cpuid = "8.1.1"
+regex = "1.3.9"
+lazy_static = "1.4.0"
+
+[features]
+default = ["amd-sev"]
```
alxiord, line 27: I don't think this should be a default feature.

src/ioctls/sev.rs
```diff
@@ -0,0 +1,680 @@
...
+    DebugEncrypt,
+    CertExport,
+}
+
+#[derive(Copy, Clone)]
+pub struct Handle(u32);
+/// Learn from ``SevState`` introduced in QEMU, adding few changes.
+/// An enumeration of SEV state information used during @query-sev.
+///
+/// @Uninit: The guest is uninitialized.
```
alxiord, line 49: nit: these should be converted to [doc comments](https://doc.rust-lang.org/stable/rust-by-example/meta/doc.html#doc-comments).

src/ioctls/sev.rs
```diff
@@ -0,0 +1,680 @@
...
+    }
+}
+
+impl SevLaunch<Init> {
+    /// Create the memory encryption context according to user-provided `Start` information,
+    /// incl. a guest policy the owner's public Diffie-Hellman (PDH) key and session information.
+    pub fn start(self, start: Start) -> Result<SevLaunch<LaunchUpdate>> {
+        let start = kvm_sev_launch_start {
+            handle: 0,
+            policy: start.policy.into(),
```
alxiord, line 173: This doesn't compile on my machine because there's no `Into<u32>` for `Policy`.

src/ioctls/sev.rs
```diff
@@ -0,0 +1,680 @@
...
+        self.sev_ioctl(SevCmd::LaunchUpdateData, update)?;
+
+        Ok(())
+    }
+
+    /// Retrieve the measurement of the data encrypted.
+    /// The guest owner may wait to provide the guest with confidential information until
+    /// it can verify the measurement thru comparison.
+    pub fn measure(self) -> Result<SevLaunch<LaunchSecret>> {
+        let mut mr = Measurement::default();
```
alxiord, line 210: This doesn't compile on my machine because there's no `impl Default` for `Measurement`.

Cargo.toml
alxiord: What's more, maybe you should also restrict the `sev` dependencies to `target=x86_64` as they won't build on ARM.

src/ioctls/sev.rs
```diff
@@ -0,0 +1,680 @@
...
+            state: Init,
+        })
+    }
+}
+
+impl From<SevLaunch<Uninit>> for SevLaunch<Init> {
+    fn from(launch: SevLaunch<Uninit>) -> SevLaunch<Init> {
+        launch
+            .init()
+            .expect("Launch state transition failed from Uninit to Init")
```
alxiord, line 163: I'm not familiar with how things are supposed to work on AMD SEV. If the transition fails, is it considered a serious enough program error that it terminates the VMM? Or is it recoverable and we can keep the VMM alive, turn this into a `TryFrom`, and propagate the error?

src/ioctls/sev.rs
```diff
@@ -0,0 +1,680 @@
...
+    /// that the memory was encrypted correctly by the firmware.
+    pub fn update_data(&self, data: &[u8]) -> Result<()> {
+        let update = kvm_sev_launch_update_data {
+            uaddr: data.as_ptr() as u64,
+            len: data.len() as u32,
+        };
+
+        self.sev_ioctl(SevCmd::LaunchUpdateData, update)?;
+
+        Ok(())
```
alxiord, line 203:
```suggestion
        self.sev_ioctl(SevCmd::LaunchUpdateData, update)
```

src/ioctls/sev.rs
```diff
@@ -0,0 +1,680 @@
...
+            state: LaunchSecret(self.state.0, mr),
+        })
+    }
+}
+
+impl From<SevLaunch<LaunchUpdate>> for SevLaunch<LaunchSecret> {
+    fn from(launch: SevLaunch<LaunchUpdate>) -> SevLaunch<LaunchSecret> {
+        launch
+            .measure()
+            .expect("Launch state transition failed from LaunchUpdate to LaunchSecret")
```
alxiord, line 230: Same question as earlier re:whether this is recoverable (applies to all state transition failures)

src/ioctls/sev.rs
```diff
@@ -0,0 +1,680 @@
...
+    }
+}
+
+impl SevLaunch<LaunchSecret> {
+    /// Get the measurement kept by the SEV launch context (in a `LaunchSecret` state)
+    pub fn get_measure(&self) -> Measurement {
```
alxiord, line 236: nit: getters usually lack the `get_` prefix in Rust

src/ioctls/sev.rs
```diff
@@ -0,0 +1,680 @@
...
+            hdr_uaddr: &mut secret.header as *mut _ as u64,
+            hdr_len: size_of_val(&secret.header) as u32,
+            guest_uaddr: gaddr,
+            guest_len: size,
+            trans_uaddr: secret.ciphertext.as_mut_ptr() as u64,
+            trans_len: secret.ciphertext.len() as u32,
+        };
+
+        self.sev_ioctl(SevCmd::LaunchSecret, secret)?;
+        Ok(())
```
alxiord, line 252:
```suggestion
        self.sev_ioctl(SevCmd::LaunchSecret, secret)
```

src/ioctls/sev.rs
```diff
@@ -0,0 +1,680 @@
...
+/// @LaunchSecret: The guest is currently being launched; ciphertext data
+///                is being imported.
+///
+/// @Running: The guest is fully launched or migrated in.
+///
+/// @SendUpdate: The guest is currently being migrated out to another machine.
+///
+/// @ReceiveUpdate: The guest is currently being migrated from another machine.
+#[derive(PartialEq, Debug)]
+pub struct Uninit;
```
alxiord, line 65: This is meant to be an enumeration, so why not express it as an `Enum`? Also, modelling the state transitions with the `From` trait is a bit weird to me, and not very idiomatic IMO - `From` conceptually implies equivalence between the 2 operands, while a state machine implies not just non-equivalence, but order between the states. I believe an `Enum` here and explicit transition functions between the states would make the code clearer.

src/ioctls/vm.rs
```diff
@@ -1259,6 +1261,35 @@ impl VmFd {
...
+    /// let launch = vm.create_sev_launch();
+    /// assert!(launch.is_ok());
+    /// ```
+    ///
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    #[cfg(feature = "amd-sev")]
+    pub fn create_sev_launch(self) -> Result<SevLaunch<Init>> {
+        let launch = SevLaunch::new_with_vm(self)?;
+
+        Ok(launch.init()?)
```
alxiord, line 1291:
```suggestion
        launch.init()
```

src/kvm_ioctls.rs
```diff
@@ -93,6 +95,10 @@ ioctl_ior_nr!(KVM_GET_PIT2, KVMIO, 0x9f, kvm_pit_state2);
 /* Available with KVM_CAP_PIT_STATE2 */
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
 ioctl_iow_nr!(KVM_SET_PIT2, KVMIO, 0xa0, kvm_pit_state2);
+/* Memory Encryption Commands */
+/* KVM_MEMORY_ENCRYPT_OP */
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+ioctl_iowr_nr!(KVM_MEMORY_ENCRYPT_OP, KVMIO, 0xba, c_ulong);
```
alxiord, line 101: Should this be gated to `feature = "amd-sev"` too?

alxiord (commented):
Apart from some code quirks (left comments), my main concern is how to consume the `sev` crate dependency (currently a crate in [`enarx`](https://github.com/enarx/enarx/tree/e044d49cb8c0e19bad8cf98b42129ff89bed1d8e/sev)) :thinking:

Cargo.toml
kailun-qin: > ~Or did you have the `sev` crate locally in your workspace? The PR doesn't build without it.~
> LE: I _think_ this is the `sev` crate from the [enarx](https://github.com/enarx/enarx/tree/e044d49cb8c0e19bad8cf98b42129ff89bed1d8e/sev) repo; but, as `enarx` doesn't have a `Cargo.toml` (they seem to use a different build system, `cargo make`?), Cargo can't get the crate dependency right with
> 
> ```toml
> sev = { git = "https://github.com/enarx/enarx.git", features = ["openssl"], optional = true }
> ```
> 
> This would be a blocker in publishing a new version of `kvm-ioctls` with `sev` support: the `sev` crate needs to be published first.
> In the meantime, for playing around, maybe you could get away with a multiple-dependency like explained [here](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations). Cargo would look for the `sev` crate on crates.io and, until it finds it there, pick it up from the local repo (to get the CI to pass, you'd have to yank it into your fork and add it to the PR).

Yes, I do have a modified `sev` crate in my workspace, which also implements the traits to resolve compilation issues you mentioned below. (that is why I marked this PR as a draft version ;))
As enarx team has cited in another thread. Let's wait a moment for their coming release of `sev` crate and APIs. I'll have an update based on that.

kailun-qin: > What's more, maybe you should also restrict the `sev` dependencies to `target=x86_64` as they won't build on ARM.

Yes agree, will update.

Cargo.toml
kailun-qin: Right, will remove the default.

src/ioctls/sev.rs
kailun-qin: > This is meant to be an enumeration, so why not express it as an `Enum`? Also, modelling the state transitions with the `From` trait is a bit weird to me, and not very idiomatic IMO - `From` conceptually implies equivalence between the 2 operands, while a state machine implies not just non-equivalence, but order between the states. I believe an `Enum` here and explicit transition functions between the states would make the code clearer.

Thanks for the suggestions. Seems reasonable to me. I'll make an update in the next version.

src/kvm_ioctls.rs
kailun-qin: Thanks, this should be gated to `feature = "amd-sev"`.

lauralt:
Hi, @kailun-qin! Are you still interested in merging this PR?

kailun-qin:
> Hi, @kailun-qin! Are you still interested in merging this PR?

Hi @lauralt ,
Thanks for following this up!
Sure, I'll continue working on this PR once enarx team stabilizes the dependent `sev` and publishes a release in the next few days. Please kindly see the previous discussions in [another thread](https://github.com/rust-vmm/community/issues/62#issuecomment-686556422).

andreeaflorescu:
This support was merged in #178. Closing this.

PR #113, opened 2020-09-03
ioctls: vcpu: Add KVM_KVMCLOCK_CTRL support

sboeuf:
The KVM_KVMCLOCK_CTRL lets the guest know that it has been paused, which
will prevent from detecting soft lockups due to pausing and resuming
operations.

### DIFF
--- a/src/ioctls/vcpu.rs
+++ b/src/ioctls/vcpu.rs
@@ -1049,6 +1049,22 @@ impl VcpuFd {
         Ok(reg_value)
     }
 
+    /// Notify the guest about the vCPU being paused.
+    ///
+    /// See the documentation for `KVM_KVMCLOCK_CTRL` in the
+    /// [KVM API documentation](https://www.kernel.org/doc/Documentation/virtual/kvm/api.txt).
+    ///
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    pub fn kvmclock_ctrl(&self) -> Result<()> {
+        // Safe because we know that our file is a KVM fd and that the request
+        // is one of the ones defined by kernel.
+        let ret = unsafe { ioctl(self, KVM_KVMCLOCK_CTRL()) };
+        if ret != 0 {
+            return Err(errno::Error::last());
+        }
+        Ok(())
+    }
+
     /// Triggers the running of the current virtual CPU returning an exit reason.
     ///
     /// See documentation for `KVM_RUN`.
@@ -1877,6 +1893,10 @@ mod tests {
             badf_errno
         );
         assert_eq!(faulty_vcpu_fd.run().unwrap_err().errno(), badf_errno);
+        assert_eq!(
+            faulty_vcpu_fd.kvmclock_ctrl().unwrap_err().errno(),
+            badf_errno
+        );
     }
 
     #[test]
--- a/coverage_config_x86_64.json
+++ b/coverage_config_x86_64.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 91.3,
+  "coverage_score": 91.2,
   "exclude_path": "",
   "crate_features": ""
 }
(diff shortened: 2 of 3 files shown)


jiangliu (dismissed):

src/ioctls/vcpu.rs
```diff
@@ -1049,6 +1049,31 @@ impl VcpuFd {
...
+    /// field of the pvclock_vcpu_time_info structure.  It will be set exclusively by
+    /// the host and read/cleared exclusively by the guest.  The guest operation of
+    /// checking and clearing the flag must an atomic operation so
+    /// load-link/store-conditional, or equivalent must be used.  There are two cases
+    /// where the guest will clear the flag: when the soft lockup watchdog timer resets
+    /// itself or when a soft lockup is detected.  This ioctl can be called any time
+    /// after pausing the vcpu, but before it is resumed.
+    ///
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    pub fn kvmclock_ctrl(&self) -> Result<()> {
```
lauralt, line 1058: Just curious, is it possible to add an `# Example` here and also some kind of test (at least in `test_faulty_vcpu_fd()` for example)?

sboeuf: Yes I'll add those :)

src/ioctls/vcpu.rs
```diff
@@ -1049,6 +1049,31 @@ impl VcpuFd {
         Ok(reg_value)
     }
 
+    /// Notify the guest about the vCPU being paused.
+    ///
+    /// This signals to the host kernel that the specified guest is being paused by
```
andreeaflorescu: Hmm, isn't this a signal from the host to the guest kernel?

jiangliu: Seems it means "signal host to notify the guest":)

sboeuf: Yes this comes directly from the KVM API documentation. Basically you invoke this ioctl to let your KVM host know that it should set a flag which will be later read by the guest.

src/ioctls/vcpu.rs
sboeuf: @lauralt I've updated `test_faulty_vcpu_fd()`, and added the example, but unfortunately for this ioctl to succeed, we need the VM to be running with the vCPU being paused at some point, which makes the whole setup a bit complex. That's why in the example, the call to `kvmclock_ctrl()` does not invoke `unwrap()` as it is failing otherwise.

jiangliu (dismissed):

src/ioctls/vcpu.rs
lauralt: :( The way the example looks now, I'm not so sure it's useful as it doesn't show some minimal use case scenario. I would either call `unwrap_err()` instead and explain in a comment why this call fails and eventually what conditions should be met for it to work or drop the whole example. What do you think?

src/ioctls/vcpu.rs
```diff
@@ -1049,6 +1049,47 @@ impl VcpuFd {
...
+    /// is shared between guest and host, specifically the second bit of the flags
+    /// field of the pvclock_vcpu_time_info structure.  It will be set exclusively by
+    /// the host and read/cleared exclusively by the guest.  The guest operation of
+    /// checking and clearing the flag must an atomic operation so
+    /// load-link/store-conditional, or equivalent must be used.  There are two cases
+    /// where the guest will clear the flag: when the soft lockup watchdog timer resets
+    /// itself or when a soft lockup is detected.  This ioctl can be called any time
+    /// after pausing the vcpu, but before it is resumed.
+    ///
+    /// See documentation for `KVM_KVMCLOCK_CTRL`.
```
lauralt: It looks like the description above contains the whole `KVM_KVMCLOCK_CTRL` documentation from https://www.kernel.org/doc/Documentation/virtual/kvm/api.txt. I would either try to summarize a bit the description and keep this line and maybe update it to:

`
See the documentation for KVM_KVMCLOCK_CTRL in the
[KVM API doc](https://www.kernel.org/doc/Documentation/virtual/kvm/api.txt).
`

or remove this line.

src/ioctls/vcpu.rs
sboeuf: I agree. That's why I'd drop the whole example as it does not provide any added value, which is supposed to be the point of these examples usually.

src/ioctls/vcpu.rs
sboeuf: Ok I'm gonna change that.

lauralt (approved):

andreeaflorescu (approved):

rbradford (approved):

PR #121, opened 2020-10-08
Add support for KVM_SET_GUEST_DEBUG ioctl call

baciumar:
This enables us to handle software breakpoints, perform single-stepping and possibly, implement hardware breakpoints

### DIFF
--- a/src/ioctls/vcpu.rs
+++ b/src/ioctls/vcpu.rs
@@ -1045,6 +1045,59 @@ impl VcpuFd {
         Ok(())
     }
 
+    /// Sets processor-specific debug registers and configures the vcpu for handling
+    /// certain guest debug events using the `KVM_SET_GUEST_DEBUG` ioctl.
+    ///
+    /// # Arguments
+    ///
+    /// * `debug_struct` - control bitfields and debug registers, depending on the specific architecture.
+    ///             For details check the `kvm_guest_debug` structure in the
+    ///             [KVM API doc](https://www.kernel.org/doc/Documentation/virtual/kvm/api.txt).
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # extern crate kvm_ioctls;
+    /// # extern crate kvm_bindings;
+    /// # use kvm_ioctls::Kvm;
+    /// # use kvm_bindings::{
+    /// #     KVM_GUESTDBG_ENABLE, KVM_GUESTDBG_USE_SW_BP, kvm_guest_debug_arch, kvm_guest_debug
+    /// # };
+    /// let kvm = Kvm::new().unwrap();
+    /// let vm = kvm.create_vm().unwrap();
+    /// let vcpu = vm.create_vcpu(0).unwrap();
+    ///
+    /// #[cfg(any(target_arch = "x86", target_arch = "x86_64"))] {
+    ///     let debug_struct = kvm_guest_debug {
+    ///         // Configure the vcpu so that a KVM_DEBUG_EXIT would be generated
+    ///         // when encountering a software breakpoint during execution
+    ///         control: KVM_GUESTDBG_ENABLE | KVM_GUESTDBG_USE_SW_BP,
+    ///         pad: 0,
+    ///         // Reset all x86-specific debug registers
+    ///         arch: kvm_guest_debug_arch {
+    ///             debugreg: [0, 0, 0, 0, 0, 0, 0, 0],
+    ///         },
+    ///     };
+    ///
+    ///     vcpu.set_guest_debug(&debug_struct).unwrap();
+    /// }
+    /// ```
+    ///
+    #[cfg(any(
+        target_arch = "x86",
+        target_arch = "x86_64",
+        target_arch = "arm64",
+        target_arch = "s390",
+        target_arch = "ppc"
+    ))]
+    pub fn set_guest_debug(&self, debug_struct: &kvm_guest_debug) -> Result<()> {
+        let ret = unsafe { ioctl_with_ref(self, KVM_SET_GUEST_DEBUG(), debug_struct) };
+        if ret < 0 {
+            return Err(errno::Error::last());
+        }
+        Ok(())
+    }
+
     /// Sets the value of one register for this vCPU.
     ///
     /// The id of the register is encoded as specified in the kernel documentation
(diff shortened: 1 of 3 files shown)


src/ioctls/vcpu.rs
```diff
@@ -1045,6 +1045,15 @@ impl VcpuFd {
         Ok(())
     }
 
+    /// Enables guest debugging and sets specific debug registers for x86
```
andreeaflorescu: Looking at the KVM documentation, it seems it is supported on: x86, s390, ppc, arm64. Can you update the description?

Can you add the arguments and example sections the documentation for this function?  You can check the documentation of other functions to get an inspiration of how it should look like.

src/ioctls/vcpu.rs
```diff
@@ -1045,6 +1045,15 @@ impl VcpuFd {
...
+    pub fn set_guest_debug(&self, debug_struct: &kvm_guest_debug) -> Result<()> {
```
andreeaflorescu, line 1093: Can this ioctl be tested in any way? The coverage drops by 1%.

andreeaflorescu (dismissed):

src/kvm_ioctls.rs
```diff
@@ -228,6 +228,8 @@ ioctl_iow_nr!(KVM_GET_DEVICE_ATTR, KVMIO, 0xe2, kvm_device_attr);
 /* Available with KVM_CAP_DEVICE_CTRL */
 ioctl_iow_nr!(KVM_HAS_DEVICE_ATTR, KVMIO, 0xe3, kvm_device_attr);
 
+ioctl_iow_nr!(KVM_SET_GUEST_DEBUG, KVMIO, 0x9b, kvm_guest_debug);
```
alxiord: Please add a comment that specifies the required capability (`KVM_CAP_SET_GUEST_DEBUG`).

src/ioctls/vcpu.rs
```diff
@@ -1045,6 +1045,15 @@ impl VcpuFd {
...
+        let ret = unsafe { ioctl_with_ref(self, KVM_SET_GUEST_DEBUG(), debug_struct) };
```
alxiord, line 1094: I think we should check the contents for `debug_struct`. The struct can contain different things for different architectures and there's no guarantee that they make sense at this point - for instance the caller could set an `aarch64` specific flag and expect it to work the same on `x86_64`.

dianpopa: What exactly do you expect this check to contain?  Also, what flags are you referring to?
If someone calls the `set_guest_debug` from `kvm-iocrtls` on aarch64 and then try to use a field from  `kvm_guest_debug_arch` that is x86_64, would not they fail the compilation step? What do you meaning by " set an aarch64 specific flag and expect it to work the same on x86_64"?

baciumar: @aghecenco  Are you maybe referring, for example, to the `control` bits supported individually by specific architectures?

andreeaflorescu: I hope that KVM will return an error if you're trying to enable debug registers on x86_64 that are only available on aarch64. It might be worth it to have it as a negative test, and check that KVM actually returns an error. Would that make sense?

baciumar: My understanding is that the `kvm_guest_debug_arch`struct is interpreted differently by each architecture (definitions for it differ among architectures) so there seems to be no problem in setting registers - the compiler will complain when using the wrong structure definition. However, regarding the `control` bitfield, when setting KVM_GUESTDBG_EXIT_PENDING (supported on s390 only) for example, and running on x86, no problem would be signaled; it would just have no effect - this is something I have tried.

alxiord: @dianpopa @baciumar yup, I was referring to the `control` field setting flags that mean different things on different architectures. 
* on `aarch64`, the only [valid flags](https://elixir.bootlin.com/linux/v4.20/source/arch/arm64/kvm/guest.c#L459) ([other link](https://elixir.bootlin.com/linux/v4.20/source/arch/arm64/include/uapi/asm/kvm.h#L148))are 0x20000 and 0x10000.
* on `x86_64` the [flags](https://elixir.bootlin.com/linux/v4.20/source/arch/x86/kvm/x86.c#L8421) are much more complex, and many of the `x86_64` flags don't make sense for `aarch64`.

Additionally, there are other flags like @baciumar mentioned, for s390 and so on.
My preference would be for this function to notify the caller when they pass flags that don't make sense, even if KVM ignores them.

dianpopa: Ok, now I understand, but are we sure we want to validate kvm-level logic in something that is intended to be a wrapper over low level ioctls? Wouldn t this mean that whenever a new control flag is added for some architecture we would also need to update the logic here?

src/ioctls/vcpu.rs
baciumar: Updated the description

src/ioctls/vcpu.rs
baciumar: Added a unittest for it

src/kvm_ioctls.rs
baciumar: Done

src/ioctls/vcpu.rs
andreeaflorescu: I second what @dianpopa said. We've used kvm-ioclts as a proxy crate that just forwards calls to KVM and checks the result. I think there are cases where we want to do the additional due diligence (like with ever growing input, or with output coming from untrusted actors), but I am not sure it applies here.

src/ioctls/vcpu.rs
```diff
@@ -2072,4 +2125,81 @@ mod tests {
...
+            control: KVM_GUESTDBG_ENABLE | KVM_GUESTDBG_SINGLESTEP,
+            pad: 0,
+            arch: kvm_guest_debug_arch {
+                debugreg: [0, 0, 0, 0, 0, 0, 0, 0],
+            },
+        };
+        vcpu_fd.set_guest_debug(&debug_struct).unwrap();
+
+        let mut instr_idx = 0;
+        loop {
```
andreeaflorescu: Can this be incorporated in `test_run_code` instead? It adds quite a lot of code that is duplicated.

baciumar: Indeed; I merged the two. However, the mmio events seem to interfere with the debug events so I left those instructions out of the check.

andreeaflorescu:
This LGTM. @baciumar can you squash your commits into one?

baciumar:
> This LGTM. @baciumar can you squash your commits into one?

Done

src/ioctls/vcpu.rs
sameo: I agree as well with @dianpopa. This crate is not the right place to validate this very much arch (and in some cases model) specific structure.

sameo (approved):

andreeaflorescu (approved):

Issue #119, opened 2020-10-07
Set a valid size for Cpuid


andreeaflorescu:
KVM only supports returning & setting 80 entries. We are not checking the number of entries passed as a parameter (`max_entries_count`) to `get_supported_cpuid` and `get_emulated_cpuid`. This is not really a huge issue because KVM checks the size against `KVM_MAX_CPUID_ENTRIES` and returns the min between the passed value and `KVM_MAX_CPUID_ENTRIES`. Instead of letting the user specify the maximum entries count, we can just remove that parameter and use the maximum value that KVM supports (`KVM_MAX_CPUID_ENTRIES = 80`). Adding another check here just uselessly complicates the logic, and we can just pick a good default instead.

The same logic should be applied to `Vcpu::get_cpuid2`.

For `Vcpu::set_cpuid2` we should just write a test to make sure that passing a too large `Cpuid` makes KVM return an error.

ahmedaabouzied:
I can work on that

andreeaflorescu:
@ahmedaabouzied there is seems to be a PR (https://github.com/rust-vmm/kvm-ioctls/pull/120) in progress already, but there is no activity on it. Let me first check if the submitter is still interested in the issue.

ahmedaabouzied:
Sure. I didn't know that because it wasn't linked to this issue

lauralt:
Hi @ahmedaabouzied, it looks like the submitter is not responding. Are you still interested in working on this? :D

ahmedaabouzied:
Indeed, will start with it and let's see how it goes

lauralt:
Hi @ahmedaabouzied. How is it going, do you need help with anything?

gabhijit:
Is anyone working on this? This looks somewhat simple to try. 

I have a question @andreeaflorescu - If we take the approach you suggested, where we can simply get rid of the parameters, will it not be breaking a public API, or is it okay to break that API? Also, since the actual underlying `ioctl` does allow setting user specified value, it's a good idea to keep the parameter than getting rid of it. So simply a change that checks for the passed value to be lesser than the `KVM_MAX_CPUID_ENTRIES` is perhaps a better way to go, that avoids an `unsafe` `ioctl` call. If you think this is fine, I will try to work on this and submit a PR soon.

andreeaflorescu:
> Is anyone working on this? This looks somewhat simple to try.
> 

[...]

PR #129, opened 2021-01-05
Fix Check For num_entries for CpuId #119

gabhijit:
Added a check to verify that the number of entires passed is less than
or equal to `KVM_MAX_CPUID_ENTRIES` (`Kvm.get_cpuid`,=`VcpuFd.get_cpuid2`).

Added test cases for failure scenarios for `get_emulated_cpuid` and
`get_supported_cpuid`.

Renamed param `max_entries_count` to `num_entries` consistent with
`VcpuFd.get_cpuid2`

### DIFF
--- a/src/ioctls/system.rs
+++ b/src/ioctls/system.rs
@@ -258,6 +258,9 @@ impl Kvm {
     /// * `num_entries` - Maximum number of CPUID entries. This function can return less than
     ///                         this when the hardware does not support so many CPUID entries.
     ///
+    /// Returns Error `errno::Error(libc::ENOMEM)` when the input `num_entries` is greater than
+    /// `KVM_MAX_CPUID_ENTRIES`.
+    ///
     /// # Example
     ///
     /// ```
@@ -285,6 +288,9 @@ impl Kvm {
     /// * `num_entries` - Maximum number of CPUID entries. This function can return less than
     ///                         this when the hardware does not support so many CPUID entries.
     ///
+    /// Returns Error `errno::Error(libc::ENOMEM)` when the input `num_entries` is greater than
+    /// `KVM_MAX_CPUID_ENTRIES`.
+    ///
     /// # Example
     ///
     /// ```
@@ -293,7 +299,7 @@ impl Kvm {
     /// use kvm_ioctls::Kvm;
     ///
     /// let kvm = Kvm::new().unwrap();
-    /// let mut cpuid = kvm.get_emulated_cpuid(KVM_MAX_CPUID_ENTRIES).unwrap();
+    /// let mut cpuid = kvm.get_supported_cpuid(KVM_MAX_CPUID_ENTRIES).unwrap();
     /// let cpuid_entries = cpuid.as_mut_slice();
     /// assert!(cpuid_entries.len() <= KVM_MAX_CPUID_ENTRIES);
     /// ```
--- a/src/ioctls/system.rs
+++ b/src/ioctls/system.rs
@@ -15,7 +15,7 @@ use ioctls::Result;
 #[cfg(any(target_arch = "aarch64"))]
 use kvm_bindings::KVM_VM_TYPE_ARM_IPA_SIZE_MASK;
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
-use kvm_bindings::{CpuId, MsrList, KVM_MAX_MSR_ENTRIES};
+use kvm_bindings::{CpuId, MsrList, KVM_MAX_CPUID_ENTRIES, KVM_MAX_MSR_ENTRIES};
 use kvm_ioctls::*;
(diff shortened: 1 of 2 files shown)


gabhijit:
Odd that I do not get the same warnings if I do a local `cargo build` and or `cargo test`? Not sure why those warnings. 

Does commit format require a 'signed-off-by'? line May be that's causing the failure?

src/ioctls/system.rs
```diff
@@ -228,13 +228,18 @@ impl Kvm {
     }
 
     #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
-    fn get_cpuid(&self, kind: u64, max_entries_count: usize) -> Result<CpuId> {
-        let mut cpuid = CpuId::new(max_entries_count);
+    fn get_cpuid(&self, kind: u64, num_entries: usize) -> Result<CpuId> {
+        if num_entries > KVM_MAX_CPUID_ENTRIES {
+            // Returns the same error the underlying `ioctl` would have sent.
+            return Err(errno::Error::new(ENOMEM));
```
acatangiu: nit: Actually newer kernels return `E2BIG`, you could do the same.

src/ioctls/system.rs
```diff
@@ -595,6 +600,14 @@ mod tests {
         assert!(cpuid_entries.len() <= KVM_MAX_CPUID_ENTRIES);
     }
 
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    #[test]
+    fn test_get_supported_cpuid_fail_nentries_high() {
```
acatangiu: nit: this is small enough to be another check in the `test_get_supported_cpuid()` test instead of a standalone test

src/ioctls/system.rs
```diff
@@ -605,6 +618,14 @@ mod tests {
         assert!(cpuid_entries.len() <= KVM_MAX_CPUID_ENTRIES);
     }
 
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    #[test]
+    fn test_get_emulated_cpuid_fail_nentries_high() {
```
acatangiu: nit: this is small enough to be another check in the `test_get_emulated_cpuid()` test instead of a standalone test

src/ioctls/vcpu.rs
```diff
@@ -376,6 +376,11 @@ impl VcpuFd {
     ///
     #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
     pub fn get_cpuid2(&self, num_entries: usize) -> Result<CpuId> {
+        if num_entries > KVM_MAX_CPUID_ENTRIES {
+            // Returns the same error the underlying `ioctl` would have sent.
+            return Err(errno::Error::new(ENOMEM));
```
acatangiu: nit: same here, `E2BIG` is better

acatangiu (commented):
The kernel already does all the validations included in this PR. The overhead here is minimal but also not necessary. I'll leave the final decision to the other maintainers.

src/ioctls/system.rs
gabhijit: Actually `E2BIG` is returned if the passed number of entries is too low and `ENOMEM` is returned if the passed number of entries is too high. So I believe in this case `ENOMEM` is right - as per [this documentation](https://www.kernel.org/doc/html/latest/virt/kvm/api.html?highlight=kvm%20ioctl)

src/ioctls/system.rs
gabhijit: Okay will do it there.

src/ioctls/system.rs
gabhijit: Okay will do it there.

src/ioctls/vcpu.rs
gabhijit: Same as above. I think `ENOMEM` is correct.

src/ioctls/system.rs
```diff
@@ -267,8 +272,8 @@ impl Kvm {
     /// ```
     ///
     #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
-    pub fn get_emulated_cpuid(&self, max_entries_count: usize) -> Result<CpuId> {
-        self.get_cpuid(KVM_GET_EMULATED_CPUID(), max_entries_count)
+    pub fn get_emulated_cpuid(&self, num_entries: usize) -> Result<CpuId> {
```
andreeaflorescu, line 278: I think we should also update the documentation of the function to specify that it returns an error when `num_entries` > KVM_MAX_CPUID_ENTRIES. Same for `get_supported_cpuid`.

andreeaflorescu (dismissed):

jiangliu (approved):

andreeaflorescu (approved):

PR #130, opened 2021-01-05
vcpu: Test Case for num_entries too small CpuId

gabhijit:
Kernel returns an error `E2BIG` if the number of entries passed to
`get_cpuid2` is too small. Added a test case for that assertion. Also,
added assertion to check `errno` to `ENOMEM`, when passed number of
entries is greater than `KVM_MAX_CPUID_ENTRIES` (previously the
assertion was there only for `is_err`).

Signed-off-by: Abhijit Gadgil <gabhijit@iitbombay.org>

### DIFF
--- a/src/ioctls/vcpu.rs
+++ b/src/ioctls/vcpu.rs
@@ -1427,8 +1427,27 @@ mod tests {
         if kvm.check_extension(Cap::ExtCpuid) {
             let vm = kvm.create_vm().unwrap();
             let vcpu = vm.create_vcpu(0).unwrap();
-            let err_cpuid = vcpu.get_cpuid2(KVM_MAX_CPUID_ENTRIES + 1 as usize);
-            assert!(err_cpuid.is_err());
+            let err_cpuid = vcpu.get_cpuid2(KVM_MAX_CPUID_ENTRIES + 1 as usize).err();
+            assert_eq!(err_cpuid.unwrap().errno(), libc::ENOMEM);
+        }
+    }
+
+    #[cfg(target_arch = "x86_64")]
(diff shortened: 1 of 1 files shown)


andreeaflorescu (approved):

src/ioctls/vcpu.rs
```diff
@@ -1427,8 +1427,27 @@ mod tests {
...
+
+    #[cfg(target_arch = "x86_64")]
+    #[test]
+    fn test_get_cpuid_fail_num_entries_too_small() {
+        let kvm = Kvm::new().unwrap();
+        if kvm.check_extension(Cap::ExtCpuid) {
+            let vm = kvm.create_vm().unwrap();
+            let cpuid = kvm.get_supported_cpuid(KVM_MAX_CPUID_ENTRIES).unwrap();
+            let ncpuids = cpuid.as_slice().len();
+            assert!(ncpuids <= KVM_MAX_CPUID_ENTRIES);
```
alexandruag, line 1443: Hmm, just wondering, can `ncpuids` ever be less than `KVM_MAX_CPUID_ENTRIES` here?

gabhijit: Yes [See here](https://www.kernel.org/doc/html/latest/virt/kvm/api.html#kvm-get-supported-cpuid) 

>  If the number is just right, the ‘nent’ field is adjusted to the number of valid entries in the ‘entries’ array, which is then filled.

alexandruag: thanks!

alexandruag (approved):

Issue #123, opened 2020-11-19
use u64 for specifying the id of the vcpu


andreeaflorescu:
Right now the `create_vcpu` function expects an u8, but KVM ioctl expects u64 here. We can update this number to match the KVM API.

gabhijit:
I think this is quite simple. Simply changing the parameter type should be good enough. No new test cases are required to be added. Right? As discussed on the other issue above - changing it to `u32` is fine? Although I think `u16` should be good enough for foreseeable future :-) 

Edit: `u16` will be borderline suffice. For powerpc `KVM_MAX_VCPU_ID` can be up to 65536. [reference](https://elixir.bootlin.com/linux/latest/source/arch/powerpc/include/asm/kvm_host.h#L37). So may be `u32` is a good idea.

andreeaflorescu:
@gabhijit I think u64 would be better so we match the KVM the `KVM_CREATE_VCPU` takes a ulong as a parameter, and so it seems the safest way to ensure we're compatible with other architectures or future enhancements.

gabhijit:
Okay. Should I submit a PR with this? Also, I think it'll be a good idea to add a test case where a number is a valid `u64` number but greater than `MAX_KVM_CPU_ID` and make sure that we get `EINVAL` errno set.

andreeaflorescu:
@gabhijit yap, sounds good.

andreeaflorescu:
Fixed by https://github.com/rust-vmm/kvm-ioctls/commit/4f61bba6236afff7beaece632e921a5428ff680d

PR #131, opened 2021-01-07
vm: `create_vcpu` takes `u64` argument #123

gabhijit:
Changed the passed parameter type from `u8` to `u64`, which the
underlying `ioctl` uses.

Added a missing cap `MaxVcpuId` (It is defined in kvm-bindings for both
arm* and x86* but was missing inside cap.rs. This was required by a test
case.)

Added a test case to assert failure if we pass parameter higher than the
`max_vcpu_id`.

Signed-off-by: Abhijit Gadgil <gabhijit@iitbombay.org>

### DIFF
--- a/src/ioctls/vm.rs
+++ b/src/ioctls/vm.rs
@@ -1021,7 +1021,7 @@ impl VmFd {
     /// let vcpu = vm.create_vcpu(0);
     /// ```
     ///
-    pub fn create_vcpu(&self, id: u8) -> Result<VcpuFd> {
+    pub fn create_vcpu(&self, id: u64) -> Result<VcpuFd> {
         // Safe because we know that vm is a VM fd and we verify the return result.
         #[allow(clippy::cast_lossless)]
         let vcpu_fd = unsafe { ioctl_with_val(&self.vm, KVM_CREATE_VCPU(), id as c_ulong) };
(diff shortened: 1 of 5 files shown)


gabhijit:
Pretty much all the build failures have the following failures - 

```~~~ Preparing plugins
[90m$[0m cd /var/lib/buildkite-agent/builds/ip-172-31-101-33-1/rust-vmm/kvm-ioctls-ci
[31m🚨 Error: Plugins have been disabled on this agent with `--no-local-hooks`[0m
^^^ +++
```
Looks like some CI issue?

andreeaflorescu:
> Pretty much all the build failures have the following failures -
> 
> ```
> �[90m$�[0m cd /var/lib/buildkite-agent/builds/ip-172-31-101-33-1/rust-vmm/kvm-ioctls-ci
> �[31m🚨 Error: Plugins have been disabled on this agent with `--no-local-hooks`�[0m
> ^^^ +++
> ```
> 
> Looks like some CI issue?

Yap, thanks for reporting it. It is now fixed, I re-triggered the CI, and it looks like all tests pass now.

src/ioctls/vm.rs
```diff
@@ -1799,6 +1799,38 @@ mod tests {
...
+        let mut max_vcpus = vm.check_extension_int(Cap::MaxVcpuId);
+        if max_vcpus == 0 {
+            max_vcpus = vm.check_extension_int(Cap::MaxVcpus);
+        }
+        if max_vcpus == 0 {
+            max_vcpus = vm.check_extension_int(Cap::NrVcpus);
+        }
+        if max_vcpus == 0 {
+            max_vcpus = 4
+        }
```
gabhijit: I think, this should be made a separate private function something like `vm.get_max_vcpu_id`? This is perhaps useful outside this test case as well.

andreeaflorescu: yes, I remember there was such a code somewhere, but I can't seem to remember where.

andreeaflorescu: Ok, I got it now. You can use the `get_max_vcpus` function that already exists in the Kvm interface.

gabhijit: `get_max_vcpus` and `get_max_vcpu_id` are different ioctls as per the [documentation here](https://www.kernel.org/doc/html/latest/virt/kvm/api.html#kvm-create-vcpu) . For instance when I checked on `x86_64` on my laptop, `max_vcpus` was 288 but `max_vcpu_id` was 1023. So I believe a function like `get_max_vcpu_id` should be added to `kvm` interface? 

and then that can be used in this test case?

andreeaflorescu: Yap, that sounds like a good idea.

gabhijit:
The build failure appears due to some `coverage` issue not sure how to fix it?

andreeaflorescu:
> The build failure appears due to some `coverage` issue not sure how to fix it?

Oh yeah, we did not fix the coverage error message. You need to update the [coverage configuration file](https://github.com/rust-vmm/kvm-ioctls/blob/master/coverage_config_x86_64.json) with the new coverage score.

gabhijit:
> > The build failure appears due to some `coverage` issue not sure how to fix it?
> 
> Oh yeah, we did not fix the coverage error message. You need to update the [coverage configuration file](https://github.com/rust-vmm/kvm-ioctls/blob/master/coverage_config_x86_64.json) with the new coverage score.

Updated. The build passes now.

src/ioctls/vm.rs
```diff
@@ -1799,6 +1799,23 @@ mod tests {
         assert!(vm.set_gsi_routing(&irq_routing).is_ok());
     }
 
+    #[test]
+    fn create_vcpu_different_cpuids() {
```
andreeaflorescu: nit: Can you rename this to `create_vcpu_different_ids`? We're using `cpuid` for something else.

andreeaflorescu (dismissed):

src/ioctls/vm.rs
gabhijit: done!

andreeaflorescu (approved):

alexandruag (approved):

gabhijit:
I believe this issue can be closed, right?

PR #139, opened 2021-02-24
update usage of FAM struct

### LINKED ISSUE #138: Fix kvm-ioctls build after new vmm-sys-util update
There were some breaking interface changes in vmm-sys-util v0.8.0 in the handling of FAM struct.
These need to be addressed for the build to pass.

Example output of failing tests: https://buildkite.com/rust-vmm/kvm-ioctls-ci/builds/602

andreeaflorescu:
This patch is required because of the latest vmm-sys-util version.

Fixes: #138 

FAM Struct now returns a result on new. Check the result before
initializing the relevant structures.

PartialEq is no longer implemented for FAM structs where the inner type
does not implement PartialEq as well. As a consequence, we cannot test
for equality between Cpuid values.

### DIFF
--- a/src/ioctls/system.rs
+++ b/src/ioctls/system.rs
@@ -255,7 +255,7 @@ impl Kvm {
             return Err(errno::Error::new(libc::ENOMEM));
         }
 
-        let mut cpuid = CpuId::new(num_entries);
+        let mut cpuid = CpuId::new(num_entries).map_err(|_| errno::Error::new(libc::ENOMEM))?;
 
         let ret = unsafe {
             // ioctl is unsafe. The kernel is trusted not to write beyond the bounds of the memory
@@ -344,7 +344,8 @@ impl Kvm {
     /// ```
     #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
     pub fn get_msr_index_list(&self) -> Result<MsrList> {
-        let mut msr_list = MsrList::new(KVM_MAX_MSR_ENTRIES);
+        let mut msr_list =
+            MsrList::new(KVM_MAX_MSR_ENTRIES).map_err(|_| errno::Error::new(libc::ENOMEM))?;
 
         let ret = unsafe {
             // ioctl is unsafe. The kernel is trusted not to write beyond the bounds of the memory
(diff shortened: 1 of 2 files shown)


acatangiu (dismissed):

rbradford (approved):

rbradford:
Doesn't the Cargo.toml need to be updated to reflect that the code is now using the new API?

andreeaflorescu:
> Doesn't the Cargo.toml need to be updated to reflect that the code is now using the new API?

Do you mean updating the version of the crate or the version of the vmm-sys-util dependency?

rbradford:
> > Doesn't the Cargo.toml need to be updated to reflect that the code is now using the new API?
> 
> Do you mean updating the version of the crate or the version of the vmm-sys-util dependency?

The version of the dependency on vmm-sys-util. The code here now depends on the newer behaviour.

jiangliu (approved):

andreeaflorescu:
> > > Doesn't the Cargo.toml need to be updated to reflect that the code is now using the new API?
> > 
> > 
> > Do you mean updating the version of the crate or the version of the vmm-sys-util dependency?
> 
> The version of the dependency on vmm-sys-util. The code here now depends on the newer behaviour.

I'll create a PR for a new release of kvm-ioctls, and also update that.

rbradford:
> > > > Doesn't the Cargo.toml need to be updated to reflect that the code is now using the new API?
> > > 
> > > 
> > > Do you mean updating the version of the crate or the version of the vmm-sys-util dependency?
> > 
> > 
> > The version of the dependency on vmm-sys-util. The code here now depends on the newer behaviour.
> 
> I'll create a PR for a new release of kvm-ioctls, and also update that.

That's cool, but in general it's probably better to try and update the Cargo.toml dependency at the same time as introducing the use of the changed API otherwise builds will break (since updating to a newer version of kvm-ioctls by git won't necessarily bring in the newer required vmm-sys-util depending on the Cargo.lock file and how it was updated.)

andreeaflorescu:
> > I'll create a PR for a new release of kvm-ioctls, and also update that.
> 
> That's cool, but in general it's probably better to try and update the Cargo.toml dependency at the same time as introducing the use of the changed API otherwise builds will break (since updating to a newer version of kvm-ioctls by git won't necessarily bring in the newer required vmm-sys-util depending on the Cargo.lock file and how it was updated.)

Sounds good to me! It was an oversight. Generally we are not expecting all commits from master to be ready to be consumed though (at least not in production use cases).

PR #143, opened 2021-03-04
Expose supplying a type when creating the VM

rbradford:
The KVM_CREATE_VM ioctl can take a parameter that is a platform and
architecture specific VM type. This can be used for SEV or TDX VMs on
x86_64 platforms of for conveying the IPA size on aarch64 platforms.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ license = "Apache-2.0 OR MIT"
 
 [dependencies]
 libc = ">=0.2.39"
-kvm-bindings = { version = ">=0.2.0", features = ["fam-wrappers"] }
+kvm-bindings = { version = "=0.3.0", features = ["fam-wrappers"] }
 vmm-sys-util = ">=0.8.0"
 
 [dev-dependencies]
--- a/src/ioctls/system.rs
+++ b/src/ioctls/system.rs
@@ -382,17 +382,7 @@ impl Kvm {
     /// ```
     ///
     pub fn create_vm(&self) -> Result<VmFd> {
-        // Safe because we know `self.kvm` is a real KVM fd as this module is the only one that
-        // create Kvm objects.
-        let ret = unsafe { ioctl(&self.kvm, KVM_CREATE_VM()) };
-        if ret >= 0 {
-            // Safe because we verify the value of ret and we are the owners of the fd.
-            let vm_file = unsafe { File::from_raw_fd(ret) };
-            let run_mmap_size = self.get_vcpu_mmap_size()?;
-            Ok(new_vmfd(vm_file, run_mmap_size))
-        } else {
-            Err(errno::Error::last())
-        }
+        self.create_vm_with_type(0) // Create using default VM type
     }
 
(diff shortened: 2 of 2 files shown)


rbradford:
@andreeaflorescu It's failing because it's pulling in two copies of vmm-sys-util. Is that because we need to pin to a newer kvm-bindings?

rbradford:
This is with local Cargo.lock deleted:

```
rob@artemis:~/src/kvm-ioctls (2021-02-23-kvm-create-type)$ cargo tree
kvm-ioctls v0.7.0 (/home/rob/src/kvm-ioctls)
├── kvm-bindings v0.3.1
│   └── vmm-sys-util v0.7.0
│       ├── bitflags v1.2.1
│       └── libc v0.2.87
├── libc v0.2.87
└── vmm-sys-util v0.8.0
    ├── bitflags v1.2.1
    └── libc v0.2.87
[dev-dependencies]
└── byteorder v1.4.2
```

sameo (dismissed):

jiangliu (dismissed):

lauralt:
kvm-bindings v0.3.1 is using vmm-sys-util v0.7.0, while kvm-ioctls  -> vmm-sys-util v.0.8.0 (details [here](https://github.com/rust-vmm/kvm-bindings/pull/31) and in the issue linked there). Looks like we should pin the kvm-bindings to =0.3.0.

rbradford:
> kvm-bindings v0.3.1 is using vmm-sys-util v0.7.0, while kvm-ioctls -> vmm-sys-util v.0.8.0 (details [here](https://github.com/rust-vmm/kvm-bindings/pull/31) and in the issue linked there). Looks like we should pin the kvm-bindings to =0.3.0.

@lauralt, I tested that, and it does work but only because the 3.0 version of kvm-bindings is more flexible in the version numbering. However that feels wrong that newer kvm-ioctls (0.7.x) forces a dependency on an older version of kvm-bindings whereas the older version of kvm-ioctls (the 0.6.x) version forces a dependency on a newer version of kvm-bindings. :exploding_head:

lauralt:
> > kvm-bindings v0.3.1 is using vmm-sys-util v0.7.0, while kvm-ioctls -> vmm-sys-util v.0.8.0 (details [here](https://github.com/rust-vmm/kvm-bindings/pull/31) and in the issue linked there). Looks like we should pin the kvm-bindings to =0.3.0.
> 
> @lauralt, I tested that, and it does work but only because the 3.0 version of kvm-bindings is more flexible in the version numbering. However that feels wrong that newer kvm-ioctls (0.7.x) forces a dependency on an older version of kvm-bindings whereas the older version of kvm-ioctls (the 0.6.x) version forces a dependency on a newer version of kvm-bindings. exploding_head

Yeah, it is indeed very weird. I am not sure what would be the best path forward here. Maybe release a new kvm-bindings version (v0.4.0). :(

sameo (approved):

jiangliu (approved):

PR #145, opened 2021-03-08
Release 0.8.0

rbradford:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+# v0.8.0
+
+## Added
+- Support for specifying VM type (an opaque platform and architecture specific
+  constant) when creating a VM (`KVM_CREATE_VM` ioctl) via the
+`Kvm::create_vm_with_type` function.
+
+## Changed
+- Now depends on kvm-bindings >=0.4.0 to support use of a newer vmm-sys-utils
+  dependency.
+
 # v0.7.0
 
 ## Added
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "kvm-ioctls"
-version = "0.7.0"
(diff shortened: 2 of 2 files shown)


andreeaflorescu:
@rbradford I was thinking we can first publish kvm-bindings v0.4.0 and then relax this dependency. As you pointed out in your previous PR, it is a bit strange to use a fixed dependency that points to a previous release. What do you say?

rbradford:
> @rbradford I was thinking we can first publish kvm-bindings v0.4.0 and then relax this dependency. As you pointed out in your previous PR, it is a bit strange to use a fixed dependency that points to a previous release. What do you say?

It would still need to be `>=0.4.0` as we don't anyone to every try it with 0.3.1 as it won't work.

andreeaflorescu:
> > @rbradford I was thinking we can first publish kvm-bindings v0.4.0 and then relax this dependency. As you pointed out in your previous PR, it is a bit strange to use a fixed dependency that points to a previous release. What do you say?
> 
> It would still need to be `>=0.4.0` as we don't anyone to every try it with 0.3.1 as it won't work.

Yes. We would need to update the `Cargo.toml` of kvm-ioctls as well.

rbradford:
kvm-bindings release PR: https://github.com/rust-vmm/kvm-bindings/pull/33 this will mean that the cloud-hypervisor & firecracker forks will need to be updated to have that newer version number.

andreeaflorescu:
I've published the new version of kvm-bindings.

rbradford:
> I've published the new version of kvm-bindings.

Thanks @andreeaflorescu this PR is updated.

andreeaflorescu (approved):

lauralt (approved):

alxiord (approved):

PR #149, opened 2021-05-11
Add KVM_SET_TSC_KHZ and KVM_GET_TSC_KHZ wrappers.

sandreim:
Add wrappers needed to enable TSC scaling for CPUs that support it.

Implemented two vCPU wrappers:
- get_tsc_khz() for KVM_GET_TSC_KHZ - depends on KVM cap KVM_CAP_GET_TSC_KHZ
- set_tsc_khz() for KVM_SET_TSC_KHZ - depends on KVM cap KVM_CAP_TSC_CONTROL

### DIFF
--- a/src/ioctls/vcpu.rs
+++ b/src/ioctls/vcpu.rs
@@ -17,7 +17,7 @@ use kvm_ioctls::*;
 use vmm_sys_util::errno;
 use vmm_sys_util::ioctl::{ioctl, ioctl_with_mut_ref, ioctl_with_ref};
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
-use vmm_sys_util::ioctl::{ioctl_with_mut_ptr, ioctl_with_ptr};
+use vmm_sys_util::ioctl::{ioctl_with_mut_ptr, ioctl_with_ptr, ioctl_with_val};
 
 /// Reasons for vCPU exits.
 ///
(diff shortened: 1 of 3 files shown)


src/ioctls/vcpu.rs
```diff
@@ -2172,4 +2205,35 @@ mod tests {
             vcpu.enable_cap(&cap).unwrap();
         }
     }
+    #[cfg(target_arch = "x86_64")]
+    #[test]
+    fn test_get_tsc_khz() {
```
andreeaflorescu, line 2234: What is the reason for having this as a separate test? I was thinking we can test both the setter and getter in the same test (something like `test_tsc_khz`) because otherwise this test seems trivial, and the functionality would be covered in  `test_set_tsc_khz`.

src/kvm_ioctls.rs
```diff
@@ -1,3 +1,4 @@
+//! A safe wrapper around the kernel's KVM interface.
```
andreeaflorescu: This change does not seem to be related to the PR.

src/ioctls/vcpu.rs
```diff
@@ -1339,6 +1339,39 @@ impl VcpuFd {
...
+    ///
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    pub fn get_tsc_khz(&self) -> Result<u32> {
+        // Safe because we know that our file is a KVM fd and that the request is one of the ones
+        // defined by kernel.
+        let ret = unsafe { ioctl(self, KVM_GET_TSC_KHZ()) };
+        if ret >= 0 {
+            Ok(ret as u32)
+        } else {
+            Err(errno::Error::new(ret))
```
andreeaflorescu, line 1364: Can you also add a negative test?

src/ioctls/vcpu.rs
```diff
@@ -1339,6 +1339,39 @@ impl VcpuFd {
...
+        }
+    }
+
+    /// Sets the specified vCPU TSC frequency.
```
andreeaflorescu, line 1368: We are trying to add usage examples as well for the public interface where this is not too hard to achieve.

andreeaflorescu (commented):
Just curious, why do we need these two ioctls? What functionality are they helping with? (maybe we can write something about it in the commit message as well).

src/ioctls/vcpu.rs
```diff
@@ -2172,4 +2205,35 @@ mod tests {
...
+        let vcpu = vm.create_vcpu(0).unwrap();
+        assert!(vcpu.get_tsc_khz().unwrap() > 0);
+    }
+
+    #[cfg(target_arch = "x86_64")]
+    #[test]
+    fn test_set_tsc_khz() {
+        let kvm = Kvm::new().unwrap();
+        // Skip test if KVM/host capabilities are missing.
+        if !(kvm.check_extension(Cap::GetTscKhz) && kvm.check_extension(Cap::TscControl)) {
```
lauralt: For the hosts that are missing these capabilities we could test that `set/get` return an error.

src/ioctls/vcpu.rs
lauralt: I think it's fine here to have separate tests since on some machines (like my laptop for example) `GetTscKhz` is present and `TscControl` is missing.
We should anyway test in `test_set_tsc_khz` that `get_tsc_khz` returns the value that was just set.

src/ioctls/vcpu.rs
georgepisaltu: Unfortunately, there is no way to reproduce the negative outcome of the function in a reliable way in the test. I think we would need mock structs for this purpose.

I updated the tests to check for machines where these capabilities are not present though.

src/ioctls/vcpu.rs
georgepisaltu: Done.

src/ioctls/vcpu.rs
georgepisaltu: Done.

src/ioctls/vcpu.rs
georgepisaltu: Done.

src/kvm_ioctls.rs
georgepisaltu: Removed.

src/ioctls/vcpu.rs
andreeaflorescu: Can we test this with a faulty vcpu fd? Here is the test: https://github.com/rust-vmm/kvm-ioctls/blob/9d0fd1fbc22918fdd03c89ca888195a8ad297685/src/ioctls/vcpu.rs#L1910

georgepisaltu: I tried using that example, PTAL.

src/ioctls/vcpu.rs
```diff
@@ -2172,4 +2227,57 @@ mod tests {
...
+        use std::os::unix::io::FromRawFd;
+
+        let faulty_vcpu_fd = VcpuFd {
+            vcpu: unsafe { File::from_raw_fd(-1) },
+            kvm_run_ptr: KvmRunWrapper {
+                kvm_run_ptr: mmap_anonymous(10),
+                mmap_size: 10,
+            },
+        };
+        assert!(faulty_vcpu_fd.get_tsc_khz().is_err());
```
andreeaflorescu: Can you just add this call to the already existing test function so that we do not need to create another faulty_vcpu_fd? Sorry, I wasn't being explicit in my previous comment.

georgepisaltu: Yes, I moved it to the other test.

lauralt (approved):

andreeaflorescu (approved):

PR #151, opened 2021-06-05
Minor enhancement and bugfixes

jiangliu:
Enhance the kvm-ioctls with:
1) a new Kvm::get_disable_exits_cap() method
2) a new VcpuFd::vcpu_file() access method
3) export VmFd::new_vmfd()
4) fix bugs in Kvm::get_max_vcpus() and Kvm::get_max_vcpu_id()

### DIFF
--- a/src/ioctls/system.rs
+++ b/src/ioctls/system.rs
@@ -239,6 +239,32 @@ impl Kvm {
         }
     }
 
+    /// Returns the "disable idle exiting" capability.
+    ///
+    /// The `KVM_CAP_X86_DISABLE_EXITS` capability provides userspace with per-VM capability
+    /// to not intercept MWAIT/HLT/PAUSE/CSTATE, which means though instructions won't cause
+    /// vm-exits and helps to improve latency in some workloads.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # extern crate kvm_bindings;
+    /// # use kvm_bindings::{KVM_X86_DISABLE_EXITS_HLT, KVM_X86_DISABLE_EXITS_PAUSE};
+    /// # use kvm_ioctls::Kvm;
+    /// let kvm = Kvm::new().unwrap();
+    /// assert!(kvm.get_disable_exits() & KVM_X86_DISABLE_EXITS_HLT != 0);
+    /// assert!(kvm.get_disable_exits() & KVM_X86_DISABLE_EXITS_PAUSE != 0);
+    /// ```
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    pub fn get_disable_exits(&self) -> u32 {
+        let x = self.check_extension_int(Cap::DisableExits);
+        if x > 0 {
+            x as u32
+        } else {
+            0
+        }
+    }
+
     #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
     fn get_cpuid(&self, kind: u64, num_entries: usize) -> Result<CpuId> {
         if num_entries > KVM_MAX_CPUID_ENTRIES {
@@ -542,7 +568,9 @@ impl FromRawFd for Kvm {
 mod tests {
     use super::*;
     #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
-    use kvm_bindings::KVM_MAX_CPUID_ENTRIES;
+    use kvm_bindings::{
+        KVM_MAX_CPUID_ENTRIES, KVM_X86_DISABLE_EXITS_HLT, KVM_X86_DISABLE_EXITS_PAUSE,
+    };
     use libc::{fcntl, FD_CLOEXEC, F_GETFD};
     #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
     use vmm_sys_util::fam::FamStruct;
@@ -596,6 +624,18 @@ mod tests {
         // Memory related getters
         assert!(kvm.get_vcpu_mmap_size().unwrap() > 0);
         assert!(kvm.get_nr_memslots() >= 32);
+    }
+
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    #[test]
+    fn test_kvm_get_disable_exits() {
+        let kvm = Kvm::new().unwrap();
+        let disable_exits = kvm.get_disable_exits();
+
+        if disable_exits != 0 {
+            assert!(disable_exits & KVM_X86_DISABLE_EXITS_HLT != 0);
+            assert!(disable_exits & KVM_X86_DISABLE_EXITS_PAUSE != 0);
+        }
     }
 
     #[test]
--- a/src/ioctls/system.rs
+++ b/src/ioctls/system.rs
@@ -213,9 +213,12 @@ impl Kvm {
     /// assert!(kvm.get_max_vcpus() > 0);
     /// ```
     pub fn get_max_vcpus(&self) -> usize {
-        match self.check_extension_int(Cap::MaxVcpus) {
-            0 => self.get_nr_vcpus(),
-            x => x as usize,
+        let v = self.check_extension_int(Cap::MaxVcpus);
+
+        if v <= 0 {
+            self.get_nr_vcpus()
+        } else {
+            v as usize
         }
     }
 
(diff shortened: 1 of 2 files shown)


jiangliu:
The first patch is temporary workaround for https://github.com/rust-vmm/kvm-ioctls/issues/152

src/ioctls/system.rs
```diff
@@ -248,6 +248,26 @@ impl Kvm {
         }
     }
 
+    /// Gets the `KVM_CAP_X86_DISABLE_EXITS` capability.
```
andreeaflorescu: nit: can we rephrase this to "Returns the disable capabilities", and then add more details (similarly to how we have the documentation of `get_max_vcpu_id`) about this calling `KVM_CAP_X86_DISABLE_EXITS`?

src/ioctls/system.rs
```diff
@@ -248,6 +248,26 @@ impl Kvm {
...
+    /// # Example
+    ///
+    /// ```
+    /// # use kvm_ioctls::Kvm;
+    /// let kvm = Kvm::new().unwrap();
+    /// assert!(kvm.get_disable_exits_cap() >= 0);
+    /// ```
+    ///
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    pub fn get_disable_exits_cap(&self) -> usize {
```
andreeaflorescu: Can we rename this to `get_disable_exists` instead so we can have a similar naming convention as the other functions?

src/ioctls/system.rs
```diff
@@ -248,6 +248,26 @@ impl Kvm {
...
+        let x = self.check_extension_int(Cap::DisableExits);
+        if x > 0 {
+            x as usize
```
andreeaflorescu: Can we return an `u32` instead? This is how `KVM_X86_DISABLE_EXITS_*` are defined in kvm-bindings and it makes the check against those values easier.

src/ioctls/system.rs
```diff
@@ -248,6 +248,26 @@ impl Kvm {
...
     }
 
+    /// Gets the `KVM_CAP_X86_DISABLE_EXITS` capability.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use kvm_ioctls::Kvm;
+    /// let kvm = Kvm::new().unwrap();
+    /// assert!(kvm.get_disable_exits_cap() >= 0);
```
andreeaflorescu: Can we have a more useful example here? For example, we can import from kvm-bindings the `KVM_X86_DISABLE_EXITS_HLT` and check if that is disabled. If it is disabled we can print a message.

The point of the examples should be to help the user correctly use the API. If we import `KVM_X86_DISABLE_EXITS_HLT` and check if that bit is set, this makes the example closer to how the API is going to be used in production use cases, and thus makes it more helpful.

src/ioctls/system.rs
```diff
@@ -615,6 +635,13 @@ mod tests {
         assert!(kvm.get_nr_memslots() >= 32);
     }
 
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    #[test]
+    fn test_kvm_get_disable_exits() {
+        let kvm = Kvm::new().unwrap();
+        let _disable_exits = kvm.get_disable_exits_cap();
```
andreeaflorescu: Can we check this value?

src/lib.rs
```diff
@@ -222,7 +222,7 @@ pub use cap::Cap;
 pub use ioctls::device::DeviceFd;
 pub use ioctls::system::Kvm;
 pub use ioctls::vcpu::{VcpuExit, VcpuFd};
-pub use ioctls::vm::{IoEventAddress, NoDatamatch, VmFd};
+pub use ioctls::vm::{new_vmfd, IoEventAddress, NoDatamatch, VmFd};
```
andreeaflorescu: The `new_vmfd` was intentionally not exported. Can you check the comment above the function? If we do export it, then the comment would need to be updated.

Why do we need to export this?

src/ioctls/vcpu.rs
```diff
@@ -1394,6 +1394,11 @@ impl VcpuFd {
             Ok(())
         }
     }
+
+    /// Get reference to the underlying vcpu file object.
+    pub fn vcpu_file(&self) -> &File {
```
andreeaflorescu: What is the use case for having the underlying file public? Can you add a comment to the function?

jiangliu: it's used to support cpu hotplug because a limitation of kvm.

src/lib.rs
jiangliu: After refactoring our code, it's not needed anymore:)

lauralt:
The CI machine was updated, so I think the first commit is not required anymore.

jiangliu:
> The CI machine was updated, so I think the first commit is not required anymore.

Done

acatangiu (dismissed):

src/ioctls/vcpu.rs
```diff
@@ -1378,6 +1378,11 @@ impl VcpuFd {
             Ok(())
         }
     }
+
+    /// Get reference to the underlying vcpu file object and then get access to underlying RawFd.
```
andreeaflorescu: This comment does not look accurate. It just returns a reference to the underlying vCPU file. Can you update this comment to something like:
```
\\\ Get a reference to the underlying vCPU file.
\\\
\\\ This can be used in the context of CPU hotplug for ..... todo details here why is this actually needed.

Comment (andreeaflorescu) [reply src/ioctls/vcpu.rs]: nit: the commit message is not accurate as we're not exporting `new_vmfd`.

File: src/ioctls/vcpu.rs (hunk 7 above)
Comment (jiangliu) [reply src/ioctls/vcpu.rs]: We have found other way to solve the problem, so I will drop this patch.

File: src/ioctls/vcpu.rs (hunk 8 above)
Comment (jiangliu) [reply src/ioctls/vcpu.rs]: dropped the patch, it's not needed anymore:)

File: src/ioctls/system.rs [hunk 9]
```diff
@@ -233,9 +236,38 @@ impl Kvm {
...
+            self.get_max_vcpus()
+        } else {
+            v as usize
+        }
+    }
+
+    /// Returns the "disable idle exiting" capability.
+    ///
+    /// The `KVM_CAP_X86_DISABLE_EXITS` capability provides userspace with per-VM capability
+    /// to not intercept MWAIT/HLT/PAUSE/CSTATE, which means though instructions won't cause
```
acatangiu, line 251:
```suggestion
    /// to not intercept MWAIT/HLT/PAUSE/CSTATE, which means such instructions won't cause
```
or `these`

acatangiu (approved):

src/ioctls/system.rs
```diff
@@ -598,6 +626,18 @@ mod tests {
...
 
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    #[test]
+    fn test_kvm_get_disable_exits() {
+        let kvm = Kvm::new().unwrap();
+        let disable_exits = kvm.get_disable_exits();
+
+        if disable_exits != 0 {
+            assert!(disable_exits & KVM_X86_DISABLE_EXITS_HLT != 0);
+            assert!(disable_exits & KVM_X86_DISABLE_EXITS_PAUSE != 0);
```
lauralt, line 643: Running this test on my laptop fails, should we check here for some capability before doing the asserts? Maybe this way we can also have a test case for `get_disable_exits` returning 0.

src/ioctls/system.rs
```diff
@@ -239,6 +239,32 @@ impl Kvm {
...
+    /// assert!(kvm.get_disable_exits() & KVM_X86_DISABLE_EXITS_PAUSE != 0);
+    /// ```
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    pub fn get_disable_exits(&self) -> u32 {
+        let x = self.check_extension_int(Cap::DisableExits);
+        if x > 0 {
+            x as u32
+        } else {
+            0
+        }
```
JonathanWoollett-Light: This can be slightly simplified.
```suggestion
        u32::try_from(x).unwrap_or(0)
```

andreeaflorescu:
@jiangliu are you still interested in getting some of these changes merged?

stefano-garzarella:
@jiangliu are you still interested in getting this PR merged? If yes, could you please re-open it in https://github.com/rust-vmm/rust-vmm/. We moved `kvm` crates there.

PR #163, opened 2021-07-21
Return kvm_debug_exit_arch with VcpuExit::Debug

### LINKED ISSUE #162: Expose `kvm_run::__bindgen_anon_1::debug`
The KVM docs say, “If the exit_reason is KVM_EXIT_DEBUG, then a vcpu is processing a debug event for which architecture specific information is returned” ([The Definitive KVM (Kernel-based Virtual Machine) API Documentation — The Linux Kernel documentation](https://www.kernel.org/doc/html/latest/virt/kvm/api.html#the-kvm-run-structure)).

I'd like to access that field on `KVM_EXIT_DEBUG`. While I could remap the `kvm_run` structure myself, it would be nicer to use the already mapped `VcpuFd::kvm_run_ptr` instead. We could add `VcpuFd::get_kvm_run(&self) -> &mut kvm_run` to allow broad access to `kvm_run` or return it from `VcpuFd::run()` with `VcpuExit::Debug(&'a kvm_bindings::kvm_run__bindgen_ty_1__bindgen_ty_5)`.

What do you think?

mkroening:
Fixes #162.

### DIFF
--- a/src/ioctls/vcpu.rs
+++ b/src/ioctls/vcpu.rs
@@ -47,7 +47,9 @@ pub enum VcpuExit<'a> {
     /// Corresponds to KVM_EXIT_HYPERCALL.
     Hypercall,
     /// Corresponds to KVM_EXIT_DEBUG.
-    Debug,
+    ///
+    /// Provides architecture specific information for the debug event.
+    Debug(kvm_debug_exit_arch),
     /// Corresponds to KVM_EXIT_HLT.
     Hlt,
     /// Corresponds to KVM_EXIT_IRQ_WINDOW_OPEN.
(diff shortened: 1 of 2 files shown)


andreeaflorescu:
@mkroening the coverage score needs to be updated. Otherwise, LGTM.

mkroening:
Done. Thanks for the quick reply!

andreeaflorescu (dismissed):

src/ioctls/vcpu.rs
```diff
@@ -1943,6 +1950,16 @@ mod tests {
...
                     assert_eq!(vcpu_regs.rip, expected_rips[instr_idx]);
                     instr_idx += 1;
+                    assert_eq!(
+                        debug,
+                        kvm_debug_exit_arch {
+                            exception: 1,
+                            pad: 0,
+                            pc: vcpu_regs.rip,
+                            dr6: 0b11111111111111110100111111110000,
+                            dr7: 1 << 10,
```
lauralt: What's the meaning of these? I just ran the tests on my local machine and this one fails with:
```rust
---- ioctls::vcpu::tests::test_run_code stdout ----
thread 'ioctls::vcpu::tests::test_run_code' panicked at 'assertion failed: `(left == right)`
  left: `kvm_debug_exit_arch { exception: 1, pad: 0, pc: 4099, dr6: 4294856688, dr7: 1024 }`,
 right: `kvm_debug_exit_arch { exception: 1, pad: 0, pc: 4099, dr6: 4294922224, dr7: 1024 }`'
```

mkroening: Good catch! I just pinned these to the values on my machine. On your machine, the RTM (restricted transactional memory) flag (bit 16) in DR6 is not set in this case. “This bit is always 1 if
the processor does not support RTM.” (Intel Manual Volume 3B, 17.2.3)

I'll rework the test case to not check this bit and also not check any reserved bits.

mkroening: I adjusted the test. Could you check again, please? :)

lauralt: Thanks!

lauralt (approved):

andreeaflorescu (approved):

PR #178, opened 2021-10-07
Add support for `KVM_MEMORY_ENCRYPT_{OP,REG_REGION,UNREG_REGION}`

rvolosatovs:
Add support for 3 `ioctls`:
- [KVM_MEMORY_ENCRYPT_OP](https://github.com/torvalds/linux/blob/5af4055fa8133662831ae2fb6e188e8f6c172688/Documentation/virt/kvm/api.rst#4110-kvm_memory_encrypt_op)
- [KVM_MEMORY_ENCRYPT_REG_REGION](https://github.com/torvalds/linux/blob/5af4055fa8133662831ae2fb6e188e8f6c172688/Documentation/virt/kvm/api.rst#4111-kvm_memory_encrypt_reg_region) (SEV-specific)
- [KVM_MEMORY_ENCRYPT_UNREG_REGION](https://github.com/torvalds/linux/blob/5af4055fa8133662831ae2fb6e188e8f6c172688/Documentation/virt/kvm/api.rst#4112-kvm_memory_encrypt_unreg_region) (SEV-specific)

cc @haraldh

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,10 @@
+// Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+fn main() {
+    // Define a `has_sev` attribute, which is used for conditional
+    // execution of SEV-specific tests and examples.
+    if std::path::Path::new("/dev/sev").exists() {
+        println!("cargo:rustc-cfg=has_sev");
+    }
+}
--- a/src/ioctls/vm.rs
+++ b/src/ioctls/vm.rs
@@ -20,6 +20,8 @@ use ioctls::{KvmRunWrapper, Result};
 use kvm_ioctls::*;
 use vmm_sys_util::errno;
 use vmm_sys_util::eventfd::EventFd;
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+use vmm_sys_util::ioctl::ioctl_with_mut_ptr;
 use vmm_sys_util::ioctl::{ioctl, ioctl_with_mut_ref, ioctl_with_ref, ioctl_with_val};
 
 /// An address either in programmable I/O space or in memory mapped I/O space.
@@ -1237,6 +1239,91 @@ impl VmFd {
     pub fn check_extension(&self, c: Cap) -> bool {
         self.check_extension_int(c) > 0
     }
+
+    /// Issues platform-specific memory encryption commands to manage encrypted VMs if
+    /// the platform supports creating those encrypted VMs.
+    ///
+    /// Currently, this ioctl is used for issuing Secure Encrypted Virtualization
+    /// (SEV) commands on AMD Processors.
+    ///
+    /// See the documentation for `KVM_MEMORY_ENCRYPT_OP` in the
+    /// [KVM API doc](https://www.kernel.org/doc/Documentation/virtual/kvm/api.txt).
+    ///
+    /// For SEV-specific functionality, prefer safe wrapper:
+    /// - [`encrypt_op_sev`](Self::encrypt_op_sev)
+    ///
+    /// # Safety
+    ///
+    /// This function is unsafe because there is no guarantee `T` is valid in this context, how
+    /// much data kernel will read from memory and where it will write data on error.
+    ///
+    /// # Arguments
+    ///
+    /// * `op` - an opaque platform specific structure.
+    ///
+    /// # Example
+    #[cfg_attr(has_sev, doc = "```rust")]
+    #[cfg_attr(not(has_sev), doc = "```rust,no_run")]
+    /// # extern crate kvm_ioctls;
+    /// # extern crate kvm_bindings;
+    /// use kvm_bindings::bindings::kvm_sev_cmd;
+    /// # use kvm_ioctls::Kvm;
+    ///
+    /// let kvm = Kvm::new().unwrap();
+    /// let vm = kvm.create_vm().unwrap();
+    ///
+    /// // Initialize the SEV platform context.
+    /// let mut init: kvm_sev_cmd = Default::default();
+    /// unsafe { vm.encrypt_op(&mut init).unwrap() };
+    /// ```
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    pub unsafe fn encrypt_op<T>(&self, op: *mut T) -> Result<()> {
+        let ret = ioctl_with_mut_ptr(self, KVM_MEMORY_ENCRYPT_OP(), op);
+        if ret == 0 {
+            Ok(())
+        } else {
+            Err(errno::Error::last())
+        }
+    }
+
+    /// Issue common lifecycle events of SEV guests, such as launching, running, snapshotting,
+    /// migrating and decommissioning via `KVM_MEMORY_ENCRYPT_OP` ioctl.
+    ///
+    /// Kernel documentation states that this ioctl can be used for testing whether SEV is enabled
+    /// by sending `NULL`. To do that, pass [`std::ptr::null_mut`](std::ptr::null_mut) to [`encrypt_op`](Self::encrypt_op).
+    ///
+    /// See the documentation for Secure Encrypted Virtualization (SEV).
+    ///
+    /// # Arguments
+    ///
+    /// * `op` - SEV-specific structure. For details check the
+    ///         [Secure Encrypted Virtualization (SEV) doc](https://www.kernel.org/doc/Documentation/virtual/kvm/amd-memory-encryption.rst).
+    ///
+    /// # Example
+    #[cfg_attr(has_sev, doc = "```rust")]
+    #[cfg_attr(not(has_sev), doc = "```rust,no_run")]
+    /// # extern crate kvm_ioctls;
+    /// # extern crate kvm_bindings;
+    /// # use std::{os::raw::c_void, ptr::null_mut};
+    /// use kvm_bindings::bindings::kvm_sev_cmd;
+    /// # use kvm_ioctls::Kvm;
+    ///
+    /// let kvm = Kvm::new().unwrap();
+    /// let vm = kvm.create_vm().unwrap();
+    ///
+    /// // Check whether SEV is enabled, optional.
+    /// assert!(unsafe { vm.encrypt_op(null_mut() as *mut c_void) }.is_ok());
+    ///
+    /// // Initialize the SEV platform context.
+    /// let mut init: kvm_sev_cmd = Default::default();
+    /// vm.encrypt_op_sev(&mut init).unwrap();
+    /// ```
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    pub fn encrypt_op_sev(&self, op: &mut kvm_sev_cmd) -> Result<()> {
+        // Safe because we know that kernel will only read the correct amount of memory from our pointer
+        // and we know where it will write it (op.error).
+        unsafe { self.encrypt_op(op) }
+    }
 }
 
 /// Helper function to create a new `VmFd`.
@@ -1803,5 +1890,16 @@ mod tests {
         let kvm = Kvm::new().unwrap();
         let vm = kvm.create_vm().unwrap();
         assert!(vm.check_extension(Cap::MpState));
+    }
+
+    #[test]
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    #[cfg_attr(not(has_sev), ignore)]
+    fn test_encrypt_op_sev() {
+        let kvm = Kvm::new().unwrap();
+        let vm = kvm.create_vm().unwrap();
+
+        let mut init: kvm_sev_cmd = Default::default();
+        assert!(vm.encrypt_op_sev(&mut init).is_ok());
     }
 }
--- a/src/ioctls/vm.rs
+++ b/src/ioctls/vm.rs
@@ -1399,6 +1399,83 @@ impl VmFd {
             Err(errno::Error::last())
         }
     }
+
+    /// Unregister a guest memory region registered with
+    /// [`register_enc_memory_region`](Self::register_enc_memory_region).
+    ///
+    /// It is used in the SEV-enabled guest.
+    ///
+    /// See the documentation for `KVM_MEMORY_ENCRYPT_UNREG_REGION` in the
+    /// [KVM API doc](https://www.kernel.org/doc/Documentation/virtual/kvm/api.txt).
+    ///
+    /// # Arguments
+    ///
+    /// * `memory_region` - Guest physical memory region.
+    ///
+    /// # Example
+    #[cfg_attr(has_sev, doc = "```rust")]
+    #[cfg_attr(not(has_sev), doc = "```rust,no_run")]
+    /// # extern crate kvm_bindings;
+    /// # extern crate kvm_ioctls;
+    /// # extern crate libc;
+    /// # use std::{fs::OpenOptions, ptr::null_mut};
+    /// # use std::os::unix::io::AsRawFd;
+    /// use kvm_bindings::bindings::{kvm_enc_region, kvm_sev_cmd, kvm_sev_launch_start, sev_cmd_id_KVM_SEV_LAUNCH_START};
+    /// # use kvm_ioctls::Kvm;
+    /// use libc;
+    ///
+    /// let kvm = Kvm::new().unwrap();
+    /// let vm = kvm.create_vm().unwrap();
+    /// let sev = OpenOptions::new()
+    ///     .read(true)
+    ///     .write(true)
+    ///     .open("/dev/sev")
+    ///     .unwrap();
+    ///
+    /// // Initialize the SEV platform context.
+    /// let mut init: kvm_sev_cmd = Default::default();
+    /// assert!(vm.encrypt_op_sev(&mut init).is_ok());
+    ///
+    /// // Create the memory encryption context.
+    /// let start_data: kvm_sev_launch_start = Default::default();
+    /// let mut start = kvm_sev_cmd {
+    ///     id: sev_cmd_id_KVM_SEV_LAUNCH_START,
+    ///     data: &start_data as *const kvm_sev_launch_start as _,
+    ///     sev_fd: sev.as_raw_fd() as _,
+    ///     ..Default::default()
+    /// };
+    /// assert!(vm.encrypt_op_sev(&mut start).is_ok());
+    ///
+    /// let addr = unsafe {
+    ///     libc::mmap(
+    ///         null_mut(),
+    ///         4096,
+    ///         libc::PROT_READ | libc::PROT_WRITE,
+    ///         libc::MAP_PRIVATE | libc::MAP_ANONYMOUS,
+    ///         -1,
+    ///         0,
+    ///     )
+    /// };
+    /// assert_ne!(addr, libc::MAP_FAILED);
+    ///
+    /// let memory_region = kvm_enc_region {
+    ///     addr: addr as _,
+    ///     size: 4096,
+    /// };
+    /// vm.register_enc_memory_region(&memory_region).unwrap();
+    /// vm.unregister_enc_memory_region(&memory_region).unwrap();
+    /// ```
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    pub fn unregister_enc_memory_region(&self, memory_region: &kvm_enc_region) -> Result<()> {
+        // Safe because we know that our file is a VM fd, we know the kernel will only read the
+        // correct amount of memory from our pointer, and we verify the return result.
+        let ret = unsafe { ioctl_with_ref(self, KVM_MEMORY_ENCRYPT_UNREG_REGION(), memory_region) };
+        if ret == 0 {
+            Ok(())
+        } else {
+            Err(errno::Error::last())
+        }
+    }
 }
 
 /// Helper function to create a new `VmFd`.
--- a/src/ioctls/vm.rs
+++ b/src/ioctls/vm.rs
@@ -1557,6 +1557,9 @@ mod tests {
     use super::*;
     use Kvm;
 
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    use std::{fs::OpenOptions, ptr::null_mut};
+
     use libc::EFD_NONBLOCK;
 
     #[test]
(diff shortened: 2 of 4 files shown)


rvolosatovs:
```
>           assert False, "Coverage drops by {:.2f}%. Please add unit tests for " \
                          "the uncovered lines.".format(diff)
E           AssertionError: Coverage drops by 2.00%. Please add unit tests for the uncovered lines.
E           assert False
```
This is tricky, since the functionality can only run on SEV-enabled hosts. I guess it would be possible to add a feature flag and guard this functionality and respective tests, although not sure if that's really necessary at this point.

rvolosatovs:
~Encountered some issues while testing, marking as draft until I'm sure this works as expected.~ that turned out to be initialization issue, SEV has to be enabled first

rvolosatovs:
Made names shorter, while staying consistent with e.g. https://docs.rs/kvm-ioctls/0.10.0/kvm_ioctls/struct.VmFd.html#method.set_user_memory_region:
```diff
diff --git a/src/ioctls/vm.rs b/src/ioctls/vm.rs
index ea97cbc..d7faf6d 100644
--- a/src/ioctls/vm.rs
+++ b/src/ioctls/vm.rs
@@ -1283,11 +1283,11 @@ impl VmFd {
     ///     addr: 0x10000 as u64,
     ///     size: 0x10000 as u64,
     /// };
-    /// vm.register_encrypted_memory_region(&memory_region).unwrap();
+    /// vm.register_enc_memory_region(&memory_region).unwrap();
     /// ```
     ///
     #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
-    pub fn register_encrypted_memory_region(&self, memory_region: &kvm_enc_region) -> Result<()> {
+    pub fn register_enc_memory_region(&self, memory_region: &kvm_enc_region) -> Result<()> {
         // Safe because we know that our file is a VM fd, we know the kernel will only read the
         // correct amount of memory from our pointer, and we verify the return result.
         let ret = unsafe { ioctl_with_ref(self, KVM_MEMORY_ENCRYPT_REG_REGION(), memory_region) };
@@ -1321,11 +1321,11 @@ impl VmFd {
     ///     addr: 0x10000 as u64,
     ///     size: 0x10000 as u64,
     /// };
-    /// vm.unregister_encrypted_memory_region(&memory_region).unwrap();
+    /// vm.unregister_enc_memory_region(&memory_region).unwrap();
     /// ```
     ///
     #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
-    pub fn unregister_encrypted_memory_region(&self, memory_region: &kvm_enc_region) -> Result<()> {
+    pub fn unregister_enc_memory_region(&self, memory_region: &kvm_enc_region) -> Result<()> {
         // Safe because we know that our file is a VM fd, we know the kernel will only read the
         // correct amount of memory from our pointer, and we verify the return result.
         let ret = unsafe { ioctl_with_ref(self, KVM_MEMORY_ENCRYPT_UNREG_REGION(), memory_region) };
```

src/ioctls/vm.rs
```diff
@@ -1259,6 +1259,82 @@ impl VmFd {
...
+    /// # extern crate kvm_bindings;
+    /// use kvm_bindings::bindings::kvm_enc_region;
+    /// use kvm_ioctls::Kvm;
+    /// let kvm = Kvm::new().unwrap();
+    /// let vm = kvm.create_vm().unwrap();
+    /// let memory_region = kvm_enc_region {
+    ///     addr: 0x10000 as u64,
+    ///     size: 0x10000 as u64,
+    /// };
+    /// vm.register_enc_memory_region(&memory_region).unwrap();
```
rvolosatovs, line 1389: I believe this would not directly work in most cases, since SEV first has to be enabled. (@haraldh)
Not sure if the whole startup sequence belongs here either.
Should I add a comment about this *in the example*? Not sure what's best to do here.

lauralt: Is it possible to verify in the test that SEV is enabled?
Yes, it would be a good idea to add a comment in the test, and also explicitly mention in the functions' documentation about the SEV requirement (i.e. the `It is used in the SEV-enabled guest.` part).

src/ioctls/vm.rs
```diff
@@ -1259,6 +1259,82 @@ impl VmFd {
...
+    ///
+    /// ```rust,no_run
+    /// # extern crate kvm_ioctls;
+    /// # extern crate kvm_bindings;
+    /// use kvm_bindings::bindings::kvm_enc_region;
+    /// use kvm_ioctls::Kvm;
+    /// let kvm = Kvm::new().unwrap();
+    /// let vm = kvm.create_vm().unwrap();
+    /// let memory_region = kvm_enc_region {
+    ///     addr: 0x10000 as u64,
```
lauralt: Is the `as u64` needed?

src/kvm_ioctls.rs
```diff
@@ -96,6 +96,10 @@ ioctl_iow_nr!(KVM_SET_PIT2, KVMIO, 0xa0, kvm_pit_state2);
 /* KVM_MEMORY_ENCRYPT_OP. Takes opaque platform dependent type: i.e. TDX or SEV */
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
 ioctl_iowr_nr!(KVM_MEMORY_ENCRYPT_OP, KVMIO, 0xba, std::os::raw::c_ulong);
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
```
lauralt, line 100: Maybe mention here as well about the SEV thing.

src/ioctls/vm.rs
```diff
@@ -1829,4 +1915,56 @@ mod tests {
...
+        let addr = unsafe {
+            libc::mmap(
+                0 as _,
+                128,
+                PROT_READ | PROT_WRITE,
+                MAP_PRIVATE | MAP_ANONYMOUS,
+                -1,
+                0,
+            )
+        } as u64;
```
rvolosatovs: I guess we should also check for error here, will add.

src/ioctls/vm.rs
```diff
@@ -1911,4 +1918,62 @@ mod tests {
...
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    fn test_register_unregister_enc_memory_region() {
+        if !Path::new("/dev/sev").exists() {
+            // This functionality is SEV-specific, skip test if SEV device does not exist.
+            return;
+        }
+        let sev_fd: u32 =
+            unsafe { open("/dev/sev\0".as_ptr() as *const c_char, O_RDWR | O_CLOEXEC) }
+                .try_into()
+                .unwrap();
```
andreeaflorescu: Why not open it as a `File` and then call `as_raw_fd()` on it?

src/ioctls/vm.rs
```diff
@@ -1911,4 +1918,62 @@ mod tests {
...
+            ..Default::default()
+        };
+        let mut start = kvm_sev_cmd {
+            id: sev_cmd_id_KVM_SEV_LAUNCH_START,
+            data: &start_data as *const kvm_sev_launch_start as _,
+            sev_fd,
+            ..Default::default()
+        };
+        assert_eq!(
+            unsafe { ioctl_with_mut_ref(&vm, KVM_MEMORY_ENCRYPT_OP(), &mut start) },
```
andreeaflorescu: Why don't we add a wrapper for this ioctl as well?

src/ioctls/vm.rs
```diff
@@ -1829,4 +1918,62 @@ mod tests {
...
+        let vm = kvm.create_vm().unwrap();
+
+        let mut init: kvm_sev_cmd = Default::default();
+        assert_eq!(
+            unsafe { ioctl_with_mut_ref(&vm, KVM_MEMORY_ENCRYPT_OP(), &mut init) },
+            0
+        );
+        assert_eq!(init.error, 0);
+
+        let start_data = kvm_sev_launch_start {
```
andreeaflorescu: Can you add some comments in the test as well to explain what test case you're writing here? Some questions that pop to mind: what is policy `0b1`? what are we actually checking with `init.error`?

src/ioctls/vm.rs
rvolosatovs: For simplicity, since we only need a `u32` and this way we don't need to deal with lifetimes and Rust closing the file descriptor after being passed to C (and working around that). I directly followed what's done for KVM internally.

rvolosatovs:
- Added support for `KVM_ENCRYPT_OP` (https://github.com/rust-vmm/kvm-ioctls/pull/178#discussion_r728006479). Note, that this is related to #111, which provides a type-safe SEV-specific implementation (although I think we still need to expose a generic `&mut T` interface anyway, since the functionality is not actually SEV-specific).
- Simplified test launch sequence:
	- The particular bit I used was required for SNP according to the documentation, but is not required for the test purposes and worked on SEV-SNP architecture without issues (https://github.com/rust-vmm/kvm-ioctls/pull/178#discussion_r728009870)
	- Removed redundant error checking (`error` field is only set when return value is negative, which we check in the test)
	- Rely on stdlib `File` (https://github.com/rust-vmm/kvm-ioctls/pull/178#discussion_r728005750)
- Added fully-functional examples
- Added more documentation
- Did some overall polishing

rvolosatovs:
Rebased, aligned mmaped region sizes with page size and updated formatting according to https://github.com/rust-vmm/kvm-ioctls/commit/b80cbcf0ba027cc3a7919d2b54e865531d4226f2 (https://github.com/rust-vmm/kvm-ioctls/pull/179)

rvolosatovs:
And updated comment formatting according to CI expectations

src/ioctls/vm.rs
```diff
@@ -1237,6 +1237,201 @@ impl VmFd {
...
+    ///
+    /// let kvm = Kvm::new().unwrap();
+    /// let vm = kvm.create_vm().unwrap();
+    ///
+    /// // Initialize the SEV platform context.
+    /// let mut init: kvm_sev_cmd = Default::default();
+    /// unsafe { vm.encrypt_op(&mut init).unwrap() };
+    /// ```
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    pub unsafe fn encrypt_op<T>(&self, op: &mut T) -> Result<()> {
```
andreeaflorescu: I think we can make this a bit safer in Rust by instead of having one wrapper over `KVM_MEMORY_ENCRYPT_OP` we just define 2:
- one that receives no parameter, and for which we send NULL to as `op`. We can call this `is_sev_enabled` or something like that.
- one that receives as a parameter the `kvm_sev_cmd`

What do you say?

rvolosatovs: I added `encrypt_opt_sev` and `is_sev_enabled`, however I kept the unsafe, generic `encrypt_op` public, because AFAIK, contrary to the kernel documentation this particular ioctl is used for functionality not limited to SEV.
@haraldh maybe you could add some context here?

rvolosatovs:
Updated the PR:
- Added safe SEV-specific wrappers around `encrypt_op` as suggested in https://github.com/rust-vmm/kvm-ioctls/pull/178#discussion_r729002538
- Introduced `has_sev` build attribute, which is used to guard the SEV-specific tests and ignore if the device is not present. That gives a way clearer test output (the ignored tests now indicate they are ignored) and lets us run the doc examples on SEV-enabled hosts

rvolosatovs:
Rebased on latest `main`

rvolosatovs:
@lauralt @andreeaflorescu is there anything I can do to help moving this forward?

andreeaflorescu:
I'll take another look, I did not review the change regarding the build.rs. You'll need to also update the coverage score so that the all the tests pass.

rvolosatovs:
> You'll need to also update the coverage score so that the all the tests pass.

It looks like there's an issue with CI (perhaps due to the config attribute?), since all the added functionality should be covered with tests at this point and there are also examples for all functionality, which are tested on supported systems

build.rs
```diff
@@ -0,0 +1,5 @@
+fn main() {
+    if std::path::Path::new("/dev/sev").exists() {
```
jiangliu, line 7: Does this means we must build on a hardware platform with sev capability?
Does it work by simple define a feature flag?

jiangliu: Seems it won't work with current rust-vmm-dev toolset because it compiles all crates within a container.

rvolosatovs: No, this is only required to avoid executing tests and examples on platforms that do not support SEV (so only relevant at runtime)

src/ioctls/vm.rs
```diff
@@ -1237,6 +1241,262 @@ impl VmFd {
...
+    /// ```
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    pub unsafe fn encrypt_op<T>(&self, op: *mut T) -> Result<()> {
+        let ret = ioctl_with_mut_ptr(self, KVM_MEMORY_ENCRYPT_OP(), op);
+        if ret == 0 {
+            Ok(())
+        } else {
+            Err(errno::Error::last())
+        }
+    }
```
jiangliu, line 1287: Could the same interface used for TDX?

src/ioctls/vm.rs
```diff
@@ -1237,6 +1241,262 @@ impl VmFd {
...
+    /// let vm = kvm.create_vm().unwrap();
+    ///
+    /// if vm.is_sev_enabled().is_ok() {
+    ///     // SEV is enabled.
+    /// }
+    /// ```
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    pub fn is_sev_enabled(&self) -> Result<()> {
+        // Safe because we know that kernel will only read the correct amount of memory from our pointer.
+        unsafe { self.encrypt_op(null_mut() as *mut c_void) }
```
jiangliu: Any linux doc about this KVM_MEMORY_ENCRYPT_OP to detect SEV cap by using a null ptr? According to linux kernel code, sev is the only provider for memory encryption. But the IOCTL interface seems generic enough, what will happen when enabling TDX?

src/ioctls/vm.rs
haraldh: `is_sev_enabled()` would need more checks than for the empty `KVM_MEMORY_ENCRYPT_OP` to succeed.

Also as later mentioned in this PR, other technologies can and will be using `KVM_MEMORY_ENCRYPT_OP`.

haraldh: See https://github.com/intel/tdx/blob/kvm/Documentation/virt/kvm/intel-tdx.rst

jiangliu: Below doc really helps to understand the implementation, so may worth quoting.
"The main ioctl to access SEV is KVM_MEMORY_ENCRYPT_OP. If the argument to KVM_MEMORY_ENCRYPT_OP is NULL, the ioctl returns 0 if SEV is enabled and ENOTTY` if it is disabled (on some older versions of Linux, the ioctl runs normally even with a NULL argument, and therefore will likely return ``EFAULT). If non-NULL, the argument to KVM_MEMORY_ENCRYPT_OP must be a struct kvm_sev_cmd"

src/ioctls/vm.rs
rvolosatovs: Yes, see https://github.com/rust-vmm/kvm-ioctls/pull/178#discussion_r732541875

src/ioctls/vm.rs
rvolosatovs: > Below doc really helps to understand the implementation, so may worth quoting. "The main ioctl to access SEV is KVM_MEMORY_ENCRYPT_OP. If the argument to KVM_MEMORY_ENCRYPT_OP is NULL, the ioctl returns 0 if SEV is enabled and ENOTTY` if it is disabled (on some older versions of Linux, the ioctl runs normally even with a NULL argument, and therefore will likely return ``EFAULT). If non-NULL, the argument to KVM_MEMORY_ENCRYPT_OP must be a struct kvm_sev_cmd"

I'm not sure we should add too many SEV-specific details here, since the purpose of this particular method is to be generic and not bound to one platform.

I'll try to add something about this in the doc for SEV-specific one.

build.rs
jiangliu: thanks!

src/ioctls/vm.rs
jiangliu: thanks

lauralt:
> > You'll need to also update the coverage score so that the all the tests pass.
> 
> It looks like there's an issue with CI (perhaps due to the config attribute?), since all the added functionality should be covered with tests at this point and there are also examples for all functionality, which are tested on supported systems

kcov is counting the lines that are executed by tests on the platform on which it runs, the CI instance doesn't have sev enabled, so the sev functionality covered by your tests won't be executed when running kcov. We are not yet running the CI on AMD, so until that point we can decrease the coverage value :(.

andreeaflorescu:
> It looks like there's an issue with CI (perhaps due to the config attribute?), since all the added functionality should be covered with tests at this point and there are also examples for all functionality, which are tested on supported systems

~~The host on which the CI runs does not have SEV enabled. This means that the tests with `#[cfg_attr(not(has_sev), ignore)] ` are ignored, so they do not contribute to the coverage score. You should just update the coverage score, and we can work on having a host with SEV so we can make sure that the tests are running in the CI as well.~~

LE: Laura already replied to this.

src/ioctls/vm.rs
rvolosatovs: FTR, added in https://github.com/rust-vmm/kvm-ioctls/pull/178/commits/14fdc74d6c3bfd8c7e1543c39aac6850e7f889a4

jiangliu:
Is there plan to add more for KVM sev commands such as KVM_SEV_LAUNCH_START, KVM_SEV_LAUNCH_UPDATE_DATA?
I'm thinking whether we should introduce submodules for sev/tdx etc.

rvolosatovs:
> Is there plan to add more for KVM sev commands such as KVM_SEV_LAUNCH_START, KVM_SEV_LAUNCH_UPDATE_DATA? I'm thinking whether we should introduce submodules for sev/tdx etc.

AFAIK there was an (unsuccessful?) attempt to add SEV commands in #111.

We currently maintain most of SEV-specific functionality externally, but if there is interest here we may consider contributing those.

jiangliu:
> #111

The enarx project also makes use of the kvm-bindings/kvm-ioctls crates, so is ti feasible to move some needed data structures from enarx into kvm-ioctls? But I do have some concern about `openssl`, which is always troublesome.

rvolosatovs:
> > #111
> 
> The enarx project also makes use of the kvm-bindings/kvm-ioctls crates, so is ti feasible to move some needed data structures from enarx into kvm-ioctls? But I do have some concern about `openssl`, which is always troublesome.

We are actually deprecating [SEV crate](https://github.com/enarx/sev/) at the moment, so it could be a good chance to salvage useful parts.
`openssl` is only necessary for certificate stuff, which not necessarily has to be added here, I believe
We could file an issue about this and discuss there?

jiangliu:
> > > #111
> > 
> > 
> > The enarx project also makes use of the kvm-bindings/kvm-ioctls crates, so is ti feasible to move some needed data structures from enarx into kvm-ioctls? But I do have some concern about `openssl`, which is always troublesome.
> 
> We are actually deprecating [SEV crate](https://github.com/enarx/sev/) at the moment, so it could be a good chance to salvage useful parts. `openssl` is only necessary for certificate stuff, which not necessarily has to be added here, I believe We could file an issue about this and discuss there?

Yes, a good idea to open a new issue for discussion.

rvolosatovs:
Any update on this?

andreeaflorescu (dismissed):

build.rs
lauralt: Can we leave a comment about this here?
This file is also missing the copyright license header.

src/ioctls/vm.rs
```diff
@@ -1316,6 +1316,82 @@ impl VmFd {
...
+    ///
+    /// It is used in the SEV-enabled guest.
+    ///
+    /// See the documentation for `KVM_MEMORY_ENCRYPT_REG_REGION`.
+    ///
+    /// # Arguments
+    ///
+    /// * `memory_region` - Guest physical memory region. For details check the
+    ///                     `kvm_enc_region` structure in the
+    ///                     [KVM API doc](https://www.kernel.org/doc/Documentation/virtual/kvm/api.txt).
```
lauralt: hmm I don't see any documentation about this structure in the doc link. I guess we can just remove the entire phrase, and maybe link the KVM API for the See the documentation for `KVM_MEMORY_ENCRYPT_REG_REGION` ...

src/ioctls/vm.rs
```diff
@@ -1392,6 +1392,83 @@ impl VmFd {
             Err(errno::Error::last())
         }
     }
+
+    /// Unregister a guest memory region registered with `register_encrypted_memory_region`.
```
lauralt: nit: `register_enc_memory_region`.

src/ioctls/vm.rs
```diff
@@ -2047,4 +2050,52 @@ mod tests {
...
+            )
+        };
+        assert_ne!(addr, libc::MAP_FAILED);
+
+        let memory_region = kvm_enc_region {
+            addr: addr as _,
+            size: 4096,
+        };
+        assert!(vm.register_enc_memory_region(&memory_region).is_ok());
+        assert!(vm.unregister_enc_memory_region(&memory_region).is_ok());
```
lauralt, line 2125: Just curious, how difficult would it be to test an error case for at least one of these? If it's too complicated I guess we can extend [this test](https://github.com/rust-vmm/kvm-ioctls/blob/main/src/ioctls/vm.rs#L1612).

rvolosatovs:
- Added documentation and license header to `build.rs`
- Fixed outdated doc reference
- Ensured the doc reference format is consistent with Rust stdlib and represents actual method references
- Added test cases for invalid `KVM_MEMORY_ENCRYPT_{,UN}REG_REGION` usage (I suppose it does not make much sense to do that for `encrypt_op`, since it's platform-specific)
- Made documentation more consistent with existing methods on `VmFd`

lauralt:
>     * Added documentation and license header to `build.rs`
> 
>     * Fixed outdated doc reference
> 
>     * Ensured the doc reference format is consistent with Rust stdlib and represents actual method references
> 
>     * Added test cases for invalid `KVM_MEMORY_ENCRYPT_{,UN}REG_REGION` usage (I suppose it does not make much sense to do that for `encrypt_op`, since it's platform-specific)
> 
>     * Made documentation more consistent with existing methods on `VmFd`

Thanks!

lauralt (approved):

andreeaflorescu (approved):

PR #184, opened 2021-11-17
derive Debug for DeviceFd

andreeaflorescu:
Without this derive it is hard to write negative tests because we cannot
do unwrap_err.


lauralt (dismissed):
The coverage needs to be updated.

lauralt (approved):

jiangliu (approved):

PR #187, opened 2021-12-03
Add support for KVM_SET_IDENTITY_MAP_ADDR

sboeuf:
Adding a missing ioctl to the list of supported ioctls. It is useful
to define the address of a one-page region that is needed on Intel
hardware because of a quirk in the virtualization implementation.

### DIFF
--- a/src/ioctls/vm.rs
+++ b/src/ioctls/vm.rs
@@ -136,6 +136,34 @@ impl VmFd {
         }
     }
 
+    /// Sets the address of the one-page region in the VM's address space.
+    ///
+    /// See the documentation for `KVM_SET_IDENTITY_MAP_ADDR`.
+    ///
+    /// # Arguments
+    ///
+    /// * `address` - Physical address of a one-page region in the guest's physical address space.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # extern crate kvm_ioctls;
+    /// # use kvm_ioctls::Kvm;
+    /// let kvm = Kvm::new().unwrap();
+    /// let vm = kvm.create_vm().unwrap();
+    /// vm.set_identity_map_address(0xfffb_c000).unwrap();
+    /// ```
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    pub fn set_identity_map_address(&self, address: u64) -> Result<()> {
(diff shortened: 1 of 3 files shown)


src/ioctls/vm.rs
```diff
@@ -1584,6 +1612,14 @@ mod tests {
         assert!(vm.set_tss_address(0xfffb_d000).is_ok());
     }
 
+    #[test]
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    fn test_set_identity_map_address() {
+        let kvm = Kvm::new().unwrap();
+        let vm = kvm.create_vm().unwrap();
+        assert!(vm.set_identity_map_address(0xfffb_c000).is_ok());
```
rbradford: I see the unit test is failing. Maybe you need to check for the capability in case the test system kernel is too old?

lauralt: Is it possible to test an error case as well, for example this thing from the documentation: `Fails if any VCPU has already been created.`?

sboeuf: Done!

sboeuf: Yes I added this check, but the failure was about an error in the implementation of the ioctl. It's all fixed now.

src/ioctls/vm.rs
```diff
@@ -136,6 +136,34 @@ impl VmFd {
         }
     }
 
+    /// Sets the address of the one-page region in the VM's address space.
+    ///
+    /// See the documentation for `KVM_SET_IDENTITY_MAP_ADDR`.
+    ///
+    /// # Arguments
+    ///
+    /// * `offset` - Physical address of a one-page region in the guest's physical address space.
```
lauralt:
```suggestion
    /// * `address` - Physical address of a one-page region in the guest's physical address space.
```

lauralt (dismissed):

src/ioctls/vm.rs
sboeuf: Ah yes thanks!

sboeuf:
@rbradford @lauralt I think the PR's ready now :)

lauralt (approved):

rbradford (approved):

PR #189, opened 2022-01-12
Add SET and HAS attr interfaces for vcpu

jongwu:
From [1], vcpu also has ioctl interfaces of KVM_SET_DEVICE_ATTR and
KVM_HAS_DEVICE_ATTR like deivce does, thus should be added to vcpu struct.

For now, I just see it is only used on arm64 in Cloud Hypervisor, so just add
these APIs for arm64.

Signed-off-by: Jianyong Wu <jianyong.wu@arm.com>
[1] https://www.kernel.org/doc/Documentation/virtual/kvm/api.txt

### DIFF
--- a/src/ioctls/vcpu.rs
+++ b/src/ioctls/vcpu.rs
@@ -132,6 +132,86 @@ impl VcpuFd {
         Ok(regs)
     }
 
+    /// Sets a specified piece of cpu configuration and/or state.
+    ///
+    /// See the documentation for `KVM_SET_DEVICE_ATTR` in
+    /// [KVM API doc](https://www.kernel.org/doc/Documentation/virtual/kvm/api.txt)
+    /// # Arguments
+    ///
+    /// * `device_attr` - The cpu attribute to be set.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # extern crate kvm_ioctls;
+    /// # extern crate kvm_bindings;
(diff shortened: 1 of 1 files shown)


src/ioctls/vcpu.rs
```diff
@@ -2256,4 +2336,27 @@ mod tests {
...
+            flags: 0,
+        };
+
+        let mut kvi: kvm_bindings::kvm_vcpu_init = kvm_bindings::kvm_vcpu_init::default();
+        vm.get_preferred_target(&mut kvi)
+            .expect("Cannot get preferred target");
+        kvi.features[0] |= 1 << kvm_bindings::KVM_ARM_VCPU_PSCI_0_2 | 1 << KVM_ARM_VCPU_PMU_V3;
+        assert!(vcpu.vcpu_init(&kvi).is_ok());
+        assert!(vcpu.has_device_attr(&dist_attr).is_ok());
+        assert!(vcpu.set_device_attr(&dist_attr).is_ok());
```
lauralt, line 2362: Can you add a test for an error case as well, similar to what is done in [test_create_device](https://github.com/rust-vmm/kvm-ioctls/blob/main/src/ioctls/device.rs#L271)?

jongwu: thanks @lauralt -, done for that.

lauralt (approved):

andreeaflorescu (approved):

PR #195, opened 2022-02-14
ioctls: Don't panic on unsupported exit reason

sboeuf:
In case the underlying KVM version is more recent than what is known
and supported by the kvm-ioctls crate, and in case KVM reports an
unknown reason for a VM exit, we don't want the vCPU run() to panic.

A more elegant way of handling such situation is by propagating the exit
reason up to the consumer's crate as it might know what to do with it.

### DIFF
--- a/src/ioctls/vcpu.rs
+++ b/src/ioctls/vcpu.rs
@@ -93,6 +93,10 @@ pub enum VcpuExit<'a> {
     IoapicEoi(u8 /* vector */),
     /// Corresponds to KVM_EXIT_HYPERV.
     Hyperv,
+    /// Corresponds to an exit reason that is unknown from the current version
+    /// of the kvm-ioctls crate. Let the consumer decide about what to do with
+    /// it.
+    Unsupported(u32),
 }
 
 /// Wrapper over KVM vCPU ioctls.
@@ -1405,7 +1409,7 @@ impl VcpuFd {
                     Ok(VcpuExit::IoapicEoi(eoi.vector))
                 }
                 KVM_EXIT_HYPERV => Ok(VcpuExit::Hyperv),
-                r => panic!("unknown kvm exit reason: {}", r),
+                r => Ok(VcpuExit::Unsupported(r)),
             }
         } else {
             Err(errno::Error::last())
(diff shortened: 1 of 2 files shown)


sboeuf:
In our case, this is particularly useful to support `KVM_EXIT_TDX` which is not yet part of the upstream Linux kernel.

src/ioctls/vcpu.rs
```diff
@@ -93,6 +93,10 @@ pub enum VcpuExit<'a> {
     IoapicEoi(u8 /* vector */),
     /// Corresponds to KVM_EXIT_HYPERV.
     Hyperv,
+    /// Corresponds to an exit reason that is unknown from the current version
+    /// of the kvm-ioctls crate. Let the consumer decide about what to do with
+    /// it.
+    Future(u32),
```
andreeaflorescu: Can we have it as `Unknown` instead?

sboeuf: I wanted to use `Unknown`, but there's an actual KVM exit reason called `KVM_EXIT_UNKNOWN` :(

Do you have any other idea?

andreeaflorescu: Right, I forgot about that one. We can name it `Unsupported`? Returning this as `Unsupported` is not going to bring much information though, it is just a stepping stone I assume until we add the actual variants defined here?

jiangliu: Really funny, a known unknown and an unknown unknown:)

andreeaflorescu: > Really funny, a known unknown and an unknown unknown:)

Exactly :laughing:

sboeuf: Yes I think `Unsupported` would be appropriate here.

sboeuf: >it is just a stepping stone I assume until we add the actual variants defined here?

Yes exactly, but we have to wait for these things to land in upstream Linux kernel before we can update the `kvm-bindings` and `kvm-ioctls`.

andreeaflorescu (approved):

jiangliu (approved):

PR #196, opened 2022-02-16
ioctls: vcpu: Expose kvm_run structure

sboeuf:
Returning a mutable reference to the kvm_run structure is needed for
unsupported VM exit reasons as it allows the consumer of the kvm-ioctls
crate to handle them correctly.

### DIFF
--- a/src/ioctls/vcpu.rs
+++ b/src/ioctls/vcpu.rs
@@ -1416,6 +1416,11 @@ impl VcpuFd {
         }
     }
 
+    /// Returns a mutable reference to the kvm_run structure
+    pub fn get_kvm_run(&mut self) -> &mut kvm_run {
+        self.kvm_run_ptr.as_mut_ref()
+    }
+
     /// Sets the `immediate_exit` flag on the `kvm_run` struct associated with this vCPU to `val`.
     pub fn set_kvm_immediate_exit(&self, val: u8) {
         let kvm_run = self.kvm_run_ptr.as_mut_ref();
@@ -2476,7 +2481,16 @@ mod tests {
     }
 
     #[test]
-    fn set_kvm_immediate_exit() {
+    fn test_get_kvm_run() {
+        let kvm = Kvm::new().unwrap();
+        let vm = kvm.create_vm().unwrap();
+        let mut vcpu = vm.create_vcpu(0).unwrap();
+        vcpu.kvm_run_ptr.as_mut_ref().immediate_exit = 1;
+        assert_eq!(vcpu.get_kvm_run().immediate_exit, 1);
+    }
+
+    #[test]
+    fn test_set_kvm_immediate_exit() {
         let kvm = Kvm::new().unwrap();
         let vm = kvm.create_vm().unwrap();
         let vcpu = vm.create_vcpu(0).unwrap();


lauralt (commented):
Looks good to me, can you also fix clippy?

sboeuf:
> Looks good to me, can you also fix clippy?

I fixed the clippy issue by allowing this behavior. My point is, I understand to "properly" fix it I should be using `&mut self` rather than `&self`, but this has huge consequences on the consumer side, meaning it will have to manage a mutable Vcpu. That's complicating a lot of things, so I'd rather workaround this clippy issue.

src/ioctls/vcpu.rs
```diff
@@ -1416,6 +1416,12 @@ impl VcpuFd {
         }
     }
 
+    /// Returns a mutable reference to the kvm_run structure
+    #[allow(clippy::mut_from_ref)]
```
jiangliu: Should this be marked as unsafe?

src/ioctls/vcpu.rs
```diff
@@ -1416,6 +1416,12 @@ impl VcpuFd {
...
+    pub fn get_kvm_run(&self) -> &mut kvm_run {
```
jiangliu: Should we mark it as unsafe? Seems it breaks the borrow checker.

dgreid: KvmRunWrapper uses the same pattern unfortunately. This clippy lint is legit. Minting mut refs from const refs is unsafe because you can create aliased mutable refs.

sboeuf: I have to admit, it'd be kind of bad to force this at the crate level. Better to make sure the crate doesn't allow for this kind of unsafe behavior.

sboeuf: > Should we mark it as unsafe? Seems it breaks the borrow checker.

How did you see it was breaking the borrow checker?

jiangliu: We could get multiple mutable reference to the underlying `kvm_run` object:)

sboeuf: But now that I updated the code to:
```rust
     pub fn get_kvm_run(&mut self) -> &mut kvm_run { 
```
the borrow checker should be able to do its job without having to specify `unsafe`, right?

jiangliu: I think so:)

sboeuf: Ok cool well then the PR is ready!

src/ioctls/vcpu.rs
```diff
@@ -2475,6 +2480,15 @@ mod tests {
         assert!(vcpu.get_reg_list(&mut reg_list).is_ok());
     }
 
+    #[test]
+    fn get_kvm_run() {
```
lauralt: nit: I think we convention is to add `test_` at the beginning of the name.

lauralt (dismissed):

src/ioctls/vcpu.rs
sboeuf: Fixed :)

lauralt (approved):

jiangliu (approved):

PR #199, opened 2022-06-23
ioctls: vm: Add register_irqfd_with_resample

NunoDasNeves:
Signed-off-by: Nuno Das Neves <nudasnev@microsoft.com>

### Summary of the PR

Extend the support for KVM_IRQFD ioctl so that irqfd + resamplefd can be registered.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/ioctls/vm.rs
+++ b/src/ioctls/vm.rs
@@ -896,6 +896,65 @@ impl VmFd {
         }
     }
 
+    /// Registers an event that will, when signaled, assert the `gsi` IRQ.
+    /// If the irqchip is resampled by the guest, the IRQ is de-asserted,
+    /// and `resamplefd` is notified.
+    ///
+    /// # Arguments
+    ///
+    /// * `fd` - `EventFd` to be signaled.
+    /// * `resamplefd` - `EventFd`to be notified on resample.
+    /// * `gsi` - IRQ to be triggered.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # extern crate kvm_ioctls;
+    /// # extern crate libc;
+    /// # extern crate vmm_sys_util;
+    /// # use kvm_ioctls::Kvm;
+    /// # use libc::EFD_NONBLOCK;
+    /// # use vmm_sys_util::eventfd::EventFd;
+    /// let kvm = Kvm::new().unwrap();
+    /// let vm = kvm.create_vm().unwrap();
+    /// let evtfd = EventFd::new(EFD_NONBLOCK).unwrap();
+    /// let resamplefd = EventFd::new(EFD_NONBLOCK).unwrap();
+    /// #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    /// {
+    ///     vm.create_irq_chip().unwrap();
+    ///     vm.register_irqfd_with_resample(&evtfd, &resamplefd, 0)
+    ///         .unwrap();
+    /// }
+    /// ```
+    #[cfg(any(
+        target_arch = "x86",
+        target_arch = "x86_64",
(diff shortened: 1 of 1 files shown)


src/ioctls/vm.rs
```diff
@@ -880,11 +881,26 @@ impl VmFd {
...
     ))]
-    pub fn register_irqfd(&self, fd: &EventFd, gsi: u32) -> Result<()> {
-        let irqfd = kvm_irqfd {
-            fd: fd.as_raw_fd() as u32,
-            gsi,
-            ..Default::default()
+    pub fn register_irqfd(
+        &self,
+        fd: &EventFd,
+        resample_fd: Option<&EventFd>,
```
lauralt: I think it would make the code more readable if we would define a new method, e.g. `register_irqfd_with_resample` instead of adding an optional parameter here. Then you should also add in the `Arguments` section of that method the new parameter and some info about it.
Even if using `Option` is the preferred way here, we can still remove some code duplication by initializing a mutable default kvm_irqfd, and then mutating the resample related fields when we have a `Some`.

src/ioctls/vm.rs
```diff
@@ -880,12 +886,21 @@ impl VmFd {
...
+        gsi: u32,
+    ) -> Result<()> {
+        let mut irqfd = kvm_irqfd {
             fd: fd.as_raw_fd() as u32,
             gsi,
             ..Default::default()
         };
+        if let Some(resample_eventfd) = resamplefd {
+            irqfd.resamplefd = resample_eventfd.as_raw_fd() as u32;
+            irqfd.flags = KVM_IRQFD_FLAG_RESAMPLE;
```
liuw: This should be `|=` here. If the default value is not zero then the code here destroys all other flags.

src/ioctls/vm.rs
NunoDasNeves: Thanks for your review -
I updated the `Arguments` and added a description of how the behavior is modified if resamplefd is given.
I changed the `kvm_irqfd` to be mutable and modify when there is `Some` resamplefd.

Regarding  `Option` vs adding another function e.g. `register_irqfd_and_resamplefd`:
- Adding optional resamplefd makes this function more directly match the `KVM_IRQFD` ioctl
- It's not much harder to read/use - users of kvm-ioctls are probably familiar with the kvm api anyway

On the other hand,
- There is already a separate function to unregister `fd`. Maybe it is a bit clearer and more idiomatic to create a new function for this new case/behavior.
- Changing the function signature of `register_irqfd` adds a little work for downstream

So, I'm a bit on the fence about creating a new function...

src/ioctls/vm.rs
NunoDasNeves: Fair enough... the default is zero, but I see your point.

src/ioctls/vm.rs
andreeaflorescu: Since the kvm-ioctls already does not map 1:1 with the KVM API, we could add the function as suggested by @lauralt because then we would also not break the API for existing consumers of this crate. Carrying a `None` in your function calls when you're not using resampling makes the code a bit harder to read as well because you don't immediately know what the `None` refers to.

Our goal with this crate is to make it easy to use for all users, not only the ones that are already familiar with KVM. That's why we also have extensive documentation and code examples.

NunoDasNeves: Thank you for the additional feedback!
I've updated this to be a new function `register_irqfd_with_resample`, as @lauralt originally suggested.

lauralt (approved):

liuw (approved):

andreeaflorescu (approved):

PR #200, opened 2022-06-28
vcpu: Fix Arm target_arch gate for set_guest_debug

MrXinWang:
### Summary of the PR

KVM_SET_GUEST_DEBUG supports arm64 architecture, but the correct
target_arch gate name in Rust should be aarch64 instead of arm64.
This commit does the fix for set_guest_debug() so that other
crates that depend on the kvm-ioctls crate can be benefited.

Signed-off-by: Henry Wang <Henry.Wang@arm.com>

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/ioctls/vcpu.rs
+++ b/src/ioctls/vcpu.rs
@@ -1157,17 +1157,15 @@ impl VcpuFd {
     /// let vm = kvm.create_vm().unwrap();
     /// let vcpu = vm.create_vcpu(0).unwrap();
     ///
-    /// #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    /// #[cfg(any(target_arch = "x86", target_arch = "x86_64", target_arch = "aarch64"))]
     /// {
     ///     let debug_struct = kvm_guest_debug {
     ///         // Configure the vcpu so that a KVM_DEBUG_EXIT would be generated
     ///         // when encountering a software breakpoint during execution
     ///         control: KVM_GUESTDBG_ENABLE | KVM_GUESTDBG_USE_SW_BP,
     ///         pad: 0,
-    ///         // Reset all x86-specific debug registers
-    ///         arch: kvm_guest_debug_arch {
-    ///             debugreg: [0, 0, 0, 0, 0, 0, 0, 0],
-    ///         },
+    ///         // Reset all arch-specific debug registers
+    ///         arch: Default::default(),
     ///     };
     ///
     ///     vcpu.set_guest_debug(&debug_struct).unwrap();
@@ -1176,7 +1174,7 @@ impl VcpuFd {
     #[cfg(any(
         target_arch = "x86",
         target_arch = "x86_64",
-        target_arch = "arm64",
+        target_arch = "aarch64",
         target_arch = "s390",
         target_arch = "ppc"
     ))]


src/ioctls/vcpu.rs
```diff
@@ -1157,26 +1157,26 @@ impl VcpuFd {
...
-    ///         // Configure the vcpu so that a KVM_DEBUG_EXIT would be generated
-    ///         // when encountering a software breakpoint during execution
-    ///         control: KVM_GUESTDBG_ENABLE | KVM_GUESTDBG_USE_SW_BP,
-    ///         pad: 0,
-    ///         // Reset all x86-specific debug registers
-    ///         arch: kvm_guest_debug_arch {
-    ///             debugreg: [0, 0, 0, 0, 0, 0, 0, 0],
-    ///         },
-    ///     };
+    /// let debug_struct = kvm_guest_debug {
```
liuw: You also enable this for all other architectures such as s390 and ppc, which may not be what you intended.

MrXinWang: Good point, thanks for the review. I updated this PR and I think the latest implementation excluded the s390 and ppc.

liuw (dismissed):
LGTM

MrXinWang:
Hi @andreeaflorescu @lauralt, any comments for this one? To add a little bit more context, we found this issue when trying to enable the GDB debuggability for the guest VM.

andreeaflorescu (dismissed):

src/ioctls/vcpu.rs
```diff
@@ -1157,17 +1157,20 @@ impl VcpuFd {
...
     ///         // when encountering a software breakpoint during execution
     ///         control: KVM_GUESTDBG_ENABLE | KVM_GUESTDBG_USE_SW_BP,
     ///         pad: 0,
     ///         // Reset all x86-specific debug registers
+    ///         #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
     ///         arch: kvm_guest_debug_arch {
     ///             debugreg: [0, 0, 0, 0, 0, 0, 0, 0],
     ///         },
+    ///         #[cfg(target_arch = "aarch64")]
+    ///         arch: Default::default(),
```
lauralt, line 1168: Just curious, can we use the Default impl on both x86 and aarch64 arches?

lauralt (dismissed):

src/ioctls/vcpu.rs
MrXinWang: Hi @lauralt, to be honest I don't know :P the x86 code is the existing one so I just chose not to touch them. But if we can use the Default impl, I think the code will be much simpler as a lot of #cfg can be removed.

I am happy to change if the original author @baciumar agrees :)

lauralt: Hmm, I'm not quite sure how responsive @baciumar is now, at that time I think he was doing an internship.
To me it looks like we can safely use Default on both arches.

MrXinWang: Yep, I just looked the definition of both structures on arm (https://github.com/rust-vmm/kvm-bindings/blob/main/src/arm64/bindings.rs#L1435) and x86 (https://github.com/rust-vmm/kvm-bindings/blob/main/src/x86/bindings.rs#L3011). I agree that you are correct.

Just did the change and updated the PR. PTAL again.

lauralt (approved):

src/ioctls/vcpu.rs
lauralt: Thanks!

andreeaflorescu (approved):

PR #201, opened 2022-08-12
ioctls: Use generic type for Vcpu::{set_one_reg,get_one_reg}

rbradford:
There are some registers that are specified as 128-bit wide so assuming
that they are all 64-bit is broken.

### DIFF
--- a/src/ioctls/vcpu.rs
+++ b/src/ioctls/vcpu.rs
@@ -19,6 +19,12 @@ use vmm_sys_util::ioctl::{ioctl, ioctl_with_mut_ref, ioctl_with_ref};
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
 use vmm_sys_util::ioctl::{ioctl_with_mut_ptr, ioctl_with_ptr, ioctl_with_val};
 
+pub trait ValidRegisterType: Sized {}
+impl ValidRegisterType for u16 {}
+impl ValidRegisterType for u32 {}
+impl ValidRegisterType for u64 {}
+impl ValidRegisterType for u128 {}
+
 /// Reasons for vCPU exits.
 ///
 /// The exit reasons are mapped to the `KVM_EXIT_*` defines in the
@@ -1196,8 +1202,8 @@ impl VcpuFd {
     /// * `reg_id` - ID of the register for which we are setting the value.
     /// * `data` - value for the specified register.
     #[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
-    pub fn set_one_reg(&self, reg_id: u64, data: u64) -> Result<()> {
-        let data_ref = &data as *const u64;
+    pub fn set_one_reg<T: Sized>(&self, reg_id: u64, data: T) -> Result<()> {
+        let data_ref = &data as *const T;
         let onereg = kvm_one_reg {
             id: reg_id,
             addr: data_ref as u64,
@@ -1220,18 +1226,20 @@ impl VcpuFd {
     ///
     /// * `reg_id` - ID of the register.
     #[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
-    pub fn get_one_reg(&self, reg_id: u64) -> Result<u64> {
-        let mut reg_value = 0;
+    pub fn get_one_reg<T: ValidRegisterType>(&self, reg_id: u64) -> Result<T> {
+        let mut reg_value = std::mem::MaybeUninit::<T>::uninit();
         let mut onereg = kvm_one_reg {
             id: reg_id,
-            addr: &mut reg_value as *mut u64 as u64,
+            addr: reg_value.as_mut_ptr() as u64,
         };
 
         let ret = unsafe { ioctl_with_mut_ref(self, KVM_GET_ONE_REG(), &mut onereg) };
         if ret < 0 {
             return Err(errno::Error::last());
         }
-        Ok(reg_value)
+
+        // SAFETY: Value has been initialised by the kernel
+        Ok(unsafe { reg_value.assume_init() })
     }
 
     /// Notify the guest about the vCPU being paused.
@@ -2454,7 +2462,7 @@ mod tests {
             .expect("Failed to set pstate register");
 
         assert_eq!(
-            vcpu.get_one_reg(PSTATE_REG_ID)
+            vcpu.get_one_reg::<u64>(PSTATE_REG_ID)
                 .expect("Failed to get pstate register"),
             PSTATE_FAULT_BITS_64
         );


liuw (dismissed):
I think this is better than changing `u64` to `u128`. This also means Cloud Hypervisor itself can drop a lot of sketchy casting code.

sboeuf (dismissed):

src/ioctls/vcpu.rs
```diff
@@ -1220,11 +1220,12 @@ impl VcpuFd {
     ///
     /// * `reg_id` - ID of the register.
     #[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
-    pub fn get_one_reg(&self, reg_id: u64) -> Result<u64> {
-        let mut reg_value = 0;
+    pub fn get_one_reg<T: Sized>(&self, reg_id: u64) -> Result<T> {
+        // SAFETY: Zeroing the value for the kernel to write to
+        let mut reg_value: T = unsafe { std::mem::zeroed() };
```
lauralt: What happens if the user is passing a reference type as T? I don't think this will be safe anymore.

liuw: You get a runtime error which is easy to catch. https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=66745c63561329a210caf1cc09936fa8

lauralt: Then it means we should mark the function unsafe and update its documentation.

We should also add some tests exercising the new API, e.g. call the functions with u128 instead of u64.

lauralt: I'm not sure if we should just add a `Panics` section instead of marking the whole function unsafe, but according to [mem::zeroed()](https://doc.rust-lang.org/std/mem/fn.zeroed.html) documentation, we get undefined behavior when we use references (it doesn't explicitly say that we panic), so I think it would be safer to mark the function as unsafe.

liuw: I just attempted to write a negative trait bound to rule out `Deref` but that's not allowed in Rust yet.

Marking it unsafe is of course fine. Alternatively, the return type can be made `u128` so that it stays a safe function. And then perhaps have a few checks in the function to only accept below 128-bit register types.

The original implementation is already unsound (safe function can violate memory safety) so any attempt is one step towards the right direction.

rbradford: The alternative is to use `MaybeUninit`:

```rust
    pub fn get_one_reg<T: Sized + Copy>(&self, reg_id: u64) -> Result<T> {
        let mut reg_value = std::mem::MaybeUninit::<T>::uninit();
        let mut onereg = kvm_one_reg {
            id: reg_id,
            // SAFETY: Passing a temporary value on the stack that is sized
            addr: unsafe { reg_value.as_mut_ptr() as u64 },
        };

        let ret = unsafe { ioctl_with_mut_ref(self, KVM_GET_ONE_REG(), &mut onereg) };
        if ret < 0 {
            return Err(errno::Error::last());
        }

        // SAFETY: Value has been initialised by the kernel
        Ok(unsafe { reg_value.assume_init() })
    }
```

rbradford: Also, we could check the size of the register (as it's embedded in `reg_id`) matches the size of the datatype.

liuw: On second thought, I don't think using `MaybeUninit` is going to cut it.

Rust requires reference to be valid at all times. If a caller passes a reference type, it is going to get back a seemingly valid reference, but that's not the case.

liuw: We can limit the types by having our own marker trait.

```
trait Unsigned {}

impl Unsigned for u16 {}
impl Unsigned for u32 {}
impl Unsigned for u64 {}
impl Unsigned for u128 {}

fn f<T: Sized + Unsigned>() -> Result<T, ()> {
    let x = std::mem::MaybeUninit::<T>::uninit();

    Err(())
}

fn main() {
    let _ = f::<u32>();
    let _ = f::<&u32>(); // Rejected
}
```

rbradford: @liuw Nice solution. I've always been disappointed/surprised that there isn't a trait for the basic primitive number types. I've updated this PR with your idea. WDYT?

liuw: Fine by me of course.

lauralt: Ill have a better look at the solution tomorrow, but just wondering, if we limit T to ValidRegisterType, isn't it safe to go back to the `mem::zeroed()` solution?

lauralt: Also if the trait is public, then the user can still implement it for unsafe types, or am I missing something?

rbradford: @lauralt I think `MaybeUninit` is actually better - this is what it was designed for. Using `std::mem::zeroed` is also wasteful as we don't need to zero the memory being handed to the kernel as it will overwrite it anyway.

I would love it not to be a public trait but it's required to be as it is exposed in the public interface. If all this is too much faff then we can go with the `u128` approach.

lauralt: We would probably need to mark the trait as unsafe.. It might be better to use an enum instead, but yeah it becomes overly complicated in my opinion anyway. I think I'm in favor of going with u128, since we are breaking the API either way, and will need turbofish code with the current proposal, as opposed to casting from u64 to u128.

src/ioctls/vcpu.rs
```diff
@@ -19,6 +19,12 @@ use vmm_sys_util::ioctl::{ioctl, ioctl_with_mut_ref, ioctl_with_ref};
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
 use vmm_sys_util::ioctl::{ioctl_with_mut_ptr, ioctl_with_ptr, ioctl_with_val};
 
+pub trait ValidRegisterType: Sized {}
```
liuw, line 22: Does this need to be `unsafe`? The users are declaring their type meets all the properties required for being used as a register and using it with `get_one_reg` does not violate Rust's invariants. The compiler itself cannot check this.

src/ioctls/vcpu.rs
```diff
@@ -1220,18 +1226,20 @@ impl VcpuFd {
     ///
     /// * `reg_id` - ID of the register.
     #[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
-    pub fn get_one_reg(&self, reg_id: u64) -> Result<u64> {
-        let mut reg_value = 0;
+    pub fn get_one_reg<T: ValidRegisterType>(&self, reg_id: u64) -> Result<T> {
+        let mut reg_value = std::mem::MaybeUninit::<T>::uninit();
```
liuw, line 1230: I think we still need to check the size of `T` here to make sure the function does not cause memory safety issues.

src/ioctls/vcpu.rs
andreeaflorescu: Marking this as unsafe will follow the same principle that we already apply to vm-memory: https://github.com/rust-vmm/vm-memory/blob/3ca7388bea433b05632ced19c82711f4d7d88a18/src/bytes.rs#L23.

src/ioctls/vcpu.rs
liuw: I missed this yesterday. Is the decision to go with u128?

Cloud Hypervisor is broken right now with a memory safety issue on ARM, so I would like to get this fixed before the next release, otherwise the release will need to declare ARM save / restore is broken.

rbradford: The problem with the `u128` approach is that it assumes that there won't be `u256` size registers in the near future. The KVM header defines register sizes up to 2048.

src/ioctls/vcpu.rs
liuw: Good to hear.

src/ioctls/vcpu.rs
liuw: My preference is still using generic. But then if there is no clear way forward, I think we can do the `u128` change to fix this issue at hand and ponder for a longer period of time whether using generic makes sense. I just cannot unsee that bug once I saw it. :-)

andreeaflorescu: Do you have a reference to the possible register values? I think we can go ahead with generics, but we need to make sure to not introduce unsafe behavior.

liuw: The exhaustive list only exists in KVM's header.

Not sure what you want to do with it. If you want to check all possible registers, that list will change in the future if more registers are made available.

andreeaflorescu: I want to look at the current registers. And potentially add a link as a comment to the code as well so people know where to find them. Can you paste a link to the lines in the KVM header that you are referring?

liuw: I was referring to `arch/arm64/include/uapi/asm/kvm.h`. Other relevant files are `arch/arm64/kvm/guest.c` and `sys_regs.c`.

andreeaflorescu: Here I see that the registers are defined as maximum 64 bytes: https://elixir.bootlin.com/linux/latest/source/arch/arm64/include/uapi/asm/kvm.h#L127. Can you use elixir an paste a link please? I do not have so much experience with arm and I think it would help with the review.

liuw: It includes other headers too. In the included `ptrace.h`, there is this https://elixir.bootlin.com/linux/latest/source/arch/arm64/include/uapi/asm/ptrace.h#L96.

rbradford: @andreeaflorescu It's the `fp_regs.vregs` that are u128s: https://elixir.bootlin.com/linux/latest/source/arch/arm64/kvm/guest.c#L92

andreeaflorescu: Thanks @rbradford. That's what I was looking for!

src/ioctls/vcpu.rs
andreeaflorescu: That's a good idea. Alternatively we can initialize this value with 0s because if the user requests a register with size larger that the size that the kernel actually writes, we will end up with garbage data in the result.

src/ioctls/vcpu.rs
liuw: Anything more to discuss? @andreeaflorescu what's your opinion here?

andreeaflorescu: We can use both the trait or u128, I don't have any strong preference. For the trait I think we need to mark it as unsafe as we did for BytedValue in vm-memory. Also, to avoid undefined behavior we need to make sure to zero out the bytes that are not set by KVM.

rbradford:
Superceded by #206

PR #202, opened 2022-08-18
Add `KVM_CAP_GUEST_DEBUG_HW_BPS/WPS`

michael2012z:
### Summary of the PR

The API `KVM_CAP_GUEST_DEBUG_HW_BPS` and `KVM_CAP_GUEST_DEBUG_HW_WPS`
are required to determine the number of supported guest debug registers.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/ioctls/system.rs
+++ b/src/ioctls/system.rs
@@ -121,7 +121,22 @@ impl Kvm {
     /// Wrapper over `KVM_CHECK_EXTENSION`.
     ///
     /// Returns 0 if the capability is not available and a positive integer otherwise.
-    fn check_extension_int(&self, c: Cap) -> i32 {
+    /// See the documentation for `KVM_CHECK_EXTENSION`.
+    ///
+    /// # Arguments
+    ///
+    /// * `c` - KVM capability to check.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use kvm_ioctls::Kvm;
+    /// use kvm_ioctls::Cap;
+    ///
+    /// let kvm = Kvm::new().unwrap();
+    /// assert!(kvm.check_extension_int(Cap::MaxVcpuId) > 0);
+    /// ```
+    pub fn check_extension_int(&self, c: Cap) -> i32 {
         // Safe because we know that our file is a KVM fd and that the extension is one of the ones
         // defined by kernel.
         unsafe { ioctl_with_val(self, KVM_CHECK_EXTENSION(), c as c_ulong) }
--- a/src/ioctls/system.rs
+++ b/src/ioctls/system.rs
@@ -97,9 +97,25 @@ impl Kvm {
     /// AArch64 specific call to get the host Intermediate Physical Address space limit.
     ///
     /// Returns 0 if the capability is not available and an integer >= 32 otherwise.
-    #[cfg(any(target_arch = "aarch64"))]
+    #[cfg(target_arch = "aarch64")]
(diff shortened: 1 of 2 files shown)


src/ioctls/system.rs
```diff
@@ -584,6 +600,14 @@ mod tests {
         }
     }
 
+    #[test]
+    #[cfg(target_arch = "aarch64")]
+    fn test_guest_debug_hw_capacity() {
+        let kvm = Kvm::new().unwrap();
+        assert!(kvm.get_guest_debug_hw_bps() >= 0);
+        assert!(kvm.get_guest_debug_hw_wps() >= 0);
```
michael2012z, line 626: To discuss:

I am not sure about the rule to check the result. The returned value vary on different machines. ">=0" seems not very meaningful.

And good idea?

src/ioctls/system.rs
```diff
@@ -97,11 +97,27 @@ impl Kvm {
...
+    #[cfg(target_arch = "aarch64")]
+    pub fn get_guest_debug_hw_bps(&self) -> i32 {
+        self.check_extension_int(Cap::DebugHwBps)
+    }
+
+    /// AArch64 specific call to get the number of supported hardware watchpoints.
+    ///
+    /// Returns 0 if the capability is not available and a positive integer otherwise.
+    #[cfg(target_arch = "aarch64")]
+    pub fn get_guest_debug_hw_wps(&self) -> i32 {
```
michael2012z, line 117: If there is increasing requirements to call the APIs that return a integer, is it good to expose `check_extension_int` itself?

src/ioctls/system.rs
```diff
@@ -97,11 +97,27 @@ impl Kvm {
...
     /// Returns 0 if the capability is not available and an integer >= 32 otherwise.
-    #[cfg(any(target_arch = "aarch64"))]
+    #[cfg(target_arch = "aarch64")]
     pub fn get_host_ipa_limit(&self) -> i32 {
         self.check_extension_int(Cap::ArmVmIPASize)
     }
 
+    /// AArch64 specific call to get the number of supported hardware breakpoints.
+    ///
+    /// Returns 0 if the capability is not available and a positive integer otherwise.
```
andreeaflorescu, line 107: Does the integer bear any meaning? Can we return a bool in case it doesn't have any useful information?

rbradford: The integer is the number of breakpoints.

src/ioctls/system.rs
andreeaflorescu: I think it's okay to check it as it is now in the test because we don't know what hosts we're running on. Can you add a comment about why we're doing the check this way? (i.e. different values on different platforms).

src/ioctls/system.rs
andreeaflorescu: We can mark it as public as well, I don't have any concern with that. That's a good point!

src/ioctls/system.rs
michael2012z: The integer is useful in supporting the debug of guest kernel. 
On x86, the max number of breakpoints is 4, while on arm the number is not fixed but is indicated by this API.

src/ioctls/system.rs
michael2012z: I added a commit for that.

andreeaflorescu (approved):

acatangiu (approved):
Looks good to me!

PR #213, opened 2023-01-12
ioctls: Add open_with_cloexec_at and new_with_path

Lencerf:
### Summary of the PR

These two functions are the same as `open_with_cloexec` and `new` except they allow users to specify the path to the KVM device file.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/ioctls/system.rs
+++ b/src/ioctls/system.rs
@@ -5,6 +5,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the THIRD-PARTY file.
 use libc::{open, O_CLOEXEC, O_RDWR};
+use std::ffi::CStr;
 use std::fs::File;
 use std::os::raw::{c_char, c_ulong};
 use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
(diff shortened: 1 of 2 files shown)


rminnich (dismissed):
very nice, this is going to be very helpful.

andreeaflorescu:
We have a problem with the CI, and we need to update it before we can merge this. The PR looks good to me, and I'll approve it once I get to fix the CI issue (later today).

src/ioctls/system.rs
```diff
@@ -45,6 +46,31 @@ impl Kvm {
...
+    /// let kvm = Kvm::new_with_path(&kvm_path).unwrap();
+    /// ```
+    #[allow(clippy::new_ret_no_self)]
+    pub fn new_with_path<P>(kvm_path: P) -> Result<Self>
+    where
+        P: AsRef<CStr>,
+    {
+        // Open `kvm_path` using `O_CLOEXEC` flag.
+        let fd = Self::open_with_cloexec_at(kvm_path, true)?;
+        // Safe because we verify that the fd is valid in `open_with_cloexec_at` and we own the fd.
```
andreeaflorescu: You will need to add the "SAFETY:" suffix before the safety comment to make the CI happy. Otherwise this clippy check will fail: https://rust-lang.github.io/rust-clippy/master/#missing_safety_doc.

andreeaflorescu:
@Lencerf I fixed the CI problems I was mentioning, can you please do a rebase and fix the way the unsafe comments are written so that they start with the prefix "SAFETY:"?

src/ioctls/system.rs
Lencerf: Done. Thank you very much, now All checks have passed.

Lencerf:
Ops, I clicked re-request review and github removed all other reviewers...😕

src/ioctls/system.rs
```diff
@@ -583,6 +640,12 @@ mod tests {
         Kvm::new().unwrap();
     }
 
+    #[test]
+    fn test_kvm_new_with_path() {
+        let kvm_path = std::ffi::CString::new("/dev/kvm").unwrap();
```
acatangiu: Why not test the same `cstr` build as the func?
```suggestion
        let kvm_path = unsafe { CStr::from_bytes_with_nul_unchecked(b"/dev/kvm\0") };
```

acatangiu (dismissed):

src/ioctls/system.rs
Lencerf: Done.

andreeaflorescu (approved):

acatangiu (approved):

PR #219, opened 2023-03-22
feat: add MSR-related system ioctls

zulinx86:
### Summary of the PR

Adds two system ioctls for MSR: `KVM_GET_MSR_FEATURE_INDEX_LIST` and `KVM_GET_MSRS`.

### Requirements

These ioctls are required to read features that are available for the VM, stored in MSRs.
This is similar to `KVM_GET_SUPPORTED_CPUID`.

requirements:

  description line.
  test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Upcoming Release
+
+## Added
+- [[#219](https://github.com/rust-vmm/kvm-ioctls/pull/219)] Support for
+  `KVM_GET_MSR_FEATURE_INDEX_LIST` and `KVM_GET_MSRS` system ioctls.
+
 # v0.13.0
 
 ## Added
--- a/src/ioctls/system.rs
+++ b/src/ioctls/system.rs
@@ -17,7 +17,7 @@ use crate::kvm_ioctls::*;
 #[cfg(any(target_arch = "aarch64"))]
 use kvm_bindings::KVM_VM_TYPE_ARM_IPA_SIZE_MASK;
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# Upcoming Release
+
+## Added
+- [[#219](https://github.com/rust-vmm/kvm-ioctls/pull/219)] Add `KVM_GET_MSR_FEATURE_INDEX_LIST` and `KVM_GET_MSRS` system ioctls.
```
andreeaflorescu: nit: can you keep the line under 80 chars so that it follows the same style as the other ones in this file?

src/ioctls/system.rs
```diff
@@ -441,6 +441,81 @@ impl Kvm {
...
+            )
+        };
+        if ret < 0 {
+            return Err(errno::Error::last());
+        }
+
+        Ok(msr_list)
+    }
+
+    /// X86 specific call to read the values of MSR-based features that are available for the VM.
```
andreeaflorescu, line 479: I would like to have a comment here that explains the difference between this ioctl and the one in vcpu. For me it was confusing at first cause we already have a get_msrs ioctl wrapper 😆. We could write something like: "as opposed to the Vcpu::get_msrs, this call returns all the MSR supported by the system."

CHANGELOG.md
zulinx86: Good catch. Thanks!

src/ioctls/system.rs
zulinx86: It must be better! Thanks.

andreeaflorescu (approved):

jiangliu (approved):

PR #223, opened 2023-05-15
feat(aarch64): generic get/set_one_reg

ShadowCurse:
### Summary of the PR
Linux kernel defines arm registers up to 2048 bits wide. To support all different sized the PR changes `get/set_one_reg` methods to accept a byte slice that either contains register data or will have register data be written to.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
   `KVM_GET_MSR_FEATURE_INDEX_LIST` and `KVM_GET_MSRS` system ioctls.
 - [[#221](https://github.com/rust-vmm/kvm-ioctls/pull/221)] Add
   `Cap::ArmPmuV3`.
+- [[#223](https://github.com/rust-vmm/kvm-ioctls/pull/223)] aarch64:
+  Updated `get/set_one_reg` to support different registers sizes through
+  byte slices.
 
 # v0.13.0
 
--- a/src/ioctls/vcpu.rs
+++ b/src/ioctls/vcpu.rs
@@ -19,6 +19,12 @@ use vmm_sys_util::ioctl::{ioctl, ioctl_with_mut_ref, ioctl_with_ref};
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
 use vmm_sys_util::ioctl::{ioctl_with_mut_ptr, ioctl_with_ptr, ioctl_with_val};
 
+/// Helper method to obtain the size of the register through its id
+#[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
+fn reg_size(reg_id: u64) -> usize {
+    2_usize.pow(((reg_id & KVM_REG_SIZE_MASK) >> KVM_REG_SIZE_SHIFT) as u32)
+}
+
 /// Reasons for vCPU exits.
 ///
 /// The exit reasons are mapped to the `KVM_EXIT_*` defines in the
@@ -1162,13 +1168,21 @@ impl VcpuFd {
     /// # Arguments
     ///
     /// * `reg_id` - ID of the register for which we are setting the value.
-    /// * `data` - value for the specified register.
+    /// * `data` - byte slice where the register value will be written to.
+    ///
+    /// # Note
+    ///
+    /// `data` should be equal or bigger then the register size
+    /// oterwise function will return EINVAL error
     #[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
-    pub fn set_one_reg(&self, reg_id: u64, data: u128) -> Result<()> {
-        let data_ptr = &data as *const _;
+    pub fn set_one_reg(&self, reg_id: u64, data: &[u8]) -> Result<usize> {
+        let reg_size = reg_size(reg_id);
+        if data.len() < reg_size {
+            return Err(errno::Error::new(libc::EINVAL));
+        }
         let onereg = kvm_one_reg {
             id: reg_id,
-            addr: data_ptr as u64,
+            addr: data.as_ptr() as u64,
         };
         // SAFETY: This is safe because we allocated the struct and we know the kernel will read
         // exactly the size of the struct.
@@ -1176,10 +1190,10 @@ impl VcpuFd {
         if ret < 0 {
             return Err(errno::Error::last());
         }
-        Ok(())
+        Ok(reg_size)
     }
 
-    /// Returns the value of the specified vCPU register.
+    /// Writes the value of the specified vCPU register into provided buffer.
     ///
     /// The id of the register is encoded as specified in the kernel documentation
     /// for `KVM_GET_ONE_REG`.
@@ -1187,12 +1201,20 @@ impl VcpuFd {
     /// # Arguments
     ///
     /// * `reg_id` - ID of the register.
+    /// * `data` - byte slice where the register value will be written to.
+    /// # Note
+    ///
+    /// `data` should be equal or bigger then the register size
+    /// oterwise function will return EINVAL error
     #[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
-    pub fn get_one_reg(&self, reg_id: u64) -> Result<u128> {
-        let mut reg_value = 0;
+    pub fn get_one_reg(&self, reg_id: u64, data: &mut [u8]) -> Result<usize> {
+        let reg_size = reg_size(reg_id);
+        if data.len() < reg_size {
+            return Err(errno::Error::new(libc::EINVAL));
+        }
         let mut onereg = kvm_one_reg {
             id: reg_id,
-            addr: &mut reg_value as *mut _ as u64,
+            addr: data.as_ptr() as u64,
         };
         // SAFETY: This is safe because we allocated the struct and we know the kernel will read
         // exactly the size of the struct.
@@ -1200,7 +1222,7 @@ impl VcpuFd {
         if ret < 0 {
             return Err(errno::Error::last());
         }
-        Ok(reg_value)
+        Ok(reg_size)
     }
 
     /// Notify the guest about the vCPU being paused.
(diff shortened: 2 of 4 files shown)


src/ioctls/vcpu.rs
```diff
@@ -1188,8 +1197,8 @@ impl VcpuFd {
     ///
     /// * `reg_id` - ID of the register.
     #[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
-    pub fn get_one_reg(&self, reg_id: u64) -> Result<u128> {
-        let mut reg_value = 0;
+    pub fn get_one_reg<T: ArmRegister>(&self, reg_id: u64) -> Result<T> {
+        let mut reg_value = std::mem::MaybeUninit::<T>::uninit();
```
roypat: Mh, I think with this implementation the function needs to be marked `unsafe`: If I call, say, `get_one_reg::<u32>` for a register that is 64 bit wide, then we'll pass a point to a 32 bit wide value to the kernel, to which the kernel will happily write 64 bits, causing stack corruption. 

An alternative might be using https://github.com/torvalds/linux/blob/f1fcbaa18b28dec10281551dfe6ed3a3ed80e3d6/include/uapi/linux/kvm.h#L1364-L1374 to do a width-validation before doing the ioctl (e.g. `if std::mem::size_of::<T>() != register_width {return error}`). That shouldn't have too much of a performance impact (should be a single instruction or so) - although maybe we'll still want to also provide a `get_one_reg_unsafe` that doesnt do the check? Although I am also really not sure how KVM would know about what the size of the SVE registers happens to be at compile time, given that it can be different for each ARM CPU :/

src/ioctls/vcpu.rs
```diff
@@ -19,6 +19,15 @@ use vmm_sys_util::ioctl::{ioctl, ioctl_with_mut_ref, ioctl_with_ref};
...
 use vmm_sys_util::ioctl::{ioctl_with_mut_ptr, ioctl_with_ptr, ioctl_with_val};
 
+/// Guard trait for types that can hold arm register values
+/// SAFETY: Type that implements this trait must ensure that it is valid to initialize it
+/// by writing to it throught the pointer and it is big enough to hold register value
+pub unsafe trait ArmRegister: Sized {}
+
+unsafe impl ArmRegister for u32 {}
+unsafe impl ArmRegister for u64 {}
+unsafe impl ArmRegister for u128 {}
```
roypat: Do we also want to implement this for `[u128; N]` for reasonable values of `N` (e.g. `128 * N <= 2048`) or something?

roypat (changes requested):

src/ioctls/vcpu.rs
ShadowCurse: I was thinking, what if we change `get_one_reg` to accept buffer to write register to?
So it would be:
```rust
    pub fn set_one_reg(&self, reg_id: u64, data: &[u8]) -> Result<usize> {
        let reg_size = 2_usize.pow(((reg_id & KVM_REG_SIZE_MASK) >> KVM_REG_SIZE_SHIFT) as u32);
        if data.len() < reg_size {
            return Err(errno::Error::new(libc::EINVAL));
        }
        let onereg = kvm_one_reg {
            id: reg_id,
            addr: data as *const _ as u64,
        };
        // SAFETY: This is safe because we allocated the struct and we know the kernel will read
        // exactly the size of the struct.
        let ret = unsafe { ioctl_with_ref(self, KVM_SET_ONE_REG(), &onereg) };
        if ret < 0 {
            return Err(errno::Error::last());
        }
        Ok(reg_size)
    }

    pub fn get_one_reg(&self, reg_id: u64, data: &mut [u8]) -> Result<usize> {
        let reg_size = 2_usize.pow(((reg_id & KVM_REG_SIZE_MASK) >> KVM_REG_SIZE_SHIFT) as u32);
        if data.len() < reg_size {
            return Err(errno::Error::new(libc::EINVAL));
        }
        let mut onereg = kvm_one_reg {
            id: reg_id,
            addr: data as *mut _ as u64,
        };
        // SAFETY: This is safe because we allocated the struct and we know the kernel will read
        // exactly the size of the struct.
        let ret = unsafe { ioctl_with_mut_ref(self, KVM_GET_ONE_REG(), &mut onereg) };
        if ret < 0 {
            return Err(errno::Error::last());
        }
        // SAFETY: Value is initialized by the kernel
        Ok(reg_size)
    }
```

roypat: I like this approach a lot!

src/ioctls/vcpu.rs
JonathanWoollett-Light: This would implement it for types we don't want like `u384` since the max is fairly constricted could it be
```
unsafe impl ArmRegister for [u8;1] {} // 8
unsafe impl ArmRegister for [u8;2] {} // 16
unsafe impl ArmRegister for [u8;4] {} // 32
unsafe impl ArmRegister for [u8;8] {} // 64
unsafe impl ArmRegister for [u8;16] {} // 128
unsafe impl ArmRegister for [u8;32] {} // 256
unsafe impl ArmRegister for [u8;64] {} // 512
unsafe impl ArmRegister for [u8;128] {} // 1024
unsafe impl ArmRegister for [u8;256] {} // 2048
```

ShadowCurse: @JonathanWoollett-Light I have already switched to `buffer` approach from above.

src/ioctls/vcpu.rs
ShadowCurse: Ok converted to this new approach.

roypat (dismissed):
To me this looks future proof and safe, meaning we won't run into weird issues with too-wide ARM registers again. 

It should definitely get a CHANGELOG entry, since its a breaking change, though. Also, I'm not 100% sure the `to_le_bytes` in the tests work for big endian architectures that rust-vmm formally supports. Someone with more experience there should give that a look over.

JonathanWoollett-Light (dismissed):
Lgtm

src/ioctls/vcpu.rs
```diff
@@ -19,6 +19,12 @@ use vmm_sys_util::ioctl::{ioctl, ioctl_with_mut_ref, ioctl_with_ref};
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
 use vmm_sys_util::ioctl::{ioctl_with_mut_ptr, ioctl_with_ptr, ioctl_with_val};
 
+/// Helper method to obtain the size of the register throught its id
```
andreeaflorescu: typo: throught/through.

src/ioctls/vcpu.rs
```diff
@@ -1162,45 +1168,61 @@ impl VcpuFd {
...
+    /// * `data` - byte slice where the register value will be written to.
+    ///
+    /// # Note
+    ///
+    /// `data` should be equal or bigger then the register size
+    /// oterwise function will return EINVAL error
     #[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
-    pub fn set_one_reg(&self, reg_id: u64, data: u128) -> Result<()> {
-        let data_ptr = &data as *const _;
+    pub fn set_one_reg(&self, reg_id: u64, data: &[u8]) -> Result<usize> {
```
andreeaflorescu, line 1178: I think we need a couple of more tests for these changes, as follows:
- Error cases, where we pass `data` that is smaller than the required size for both set & get functions
- set/get a register that is 256 bytes, since this is the reason that we added this change

andreeaflorescu (commented):
LGTM, but needs more testing.

src/ioctls/vcpu.rs
roypat: Mh, for the second point, what ARM CPUs do we have in the rust-vmm CI? Afaik only Gravitron 3 (correct me if I'm wrong here @ShadowCurse) have any 256 bit registers.

andreeaflorescu: We are currently running on C6g.metal, but we can easily switch to C7g.metal which is Graviton 3. The test though doesn't need to fail in case the registers are not there because we should be able to run them on different hardware as well.

src/ioctls/vcpu.rs
ShadowCurse: fixed

src/ioctls/vcpu.rs
ShadowCurse: @andreeaflorescu I did some testing regarding register width on c6g and c7g and found out that on both system the max register length available through KVM is 128 bits. (with default kvm_vcpu_init used in tests here) 
I added error cases to `test_get/set_one_reg` tests with already present 64 bit register.

andreeaflorescu: Hmm, but then I am confused, why do we need this patch? Is this not supported by KVM?

ShadowCurse: If we want to use SVE registers to test 256 or bigger sizes we need a call to `KVM_ARM_VCPU_FINALIZE` to be implemented.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
   `KVM_GET_MSR_FEATURE_INDEX_LIST` and `KVM_GET_MSRS` system ioctls.
 - [[#221](https://github.com/rust-vmm/kvm-ioctls/pull/221)] Add
   `Cap::ArmPmuV3`.
+- [[#223](https://github.com/rust-vmm/kvm-ioctls/pull/223)] feat(aarch64):
+  generic `get/set_one_reg`
```
andreeaflorescu: Can you be more explicit with the change? What is important for customers reading this is not that get/set is generic, but what is the functionality that it now supports: i.e. support for getting and setting registers that are not of a fixed size.

ShadowCurse: Updated.

andreeaflorescu (approved):

alxiord (approved):

PR #226, opened 2023-06-15
Add capabilities for pointer authentication

bchalios:
### Summary of the PR

ARMv8.3 introduces Pointer Authentication Code (PAC) as a mechanism to protect processes from attacks like buffer overflows. 

This PR adds the capabilities for pointer authentication in Aarch64, `KVM_CAP_ARM_PTRAUTH_GENERIC` and `KVM_CAP_ARM_PTRAUTH_ADDRESS`. Also, add a unit test for checking the existence of capabilities and try to enable the feature in the vcpu initialization if present.

### Requirements

requirements:

- [ ] All commits in this PR are signed (with `git commit -s`), and the commit
  description line.
  test.

### DIFF
--- a/src/cap.rs
+++ b/src/cap.rs
@@ -150,4 +150,8 @@ pub enum Cap {
     GetMsrFeatures = KVM_CAP_GET_MSR_FEATURES,
     #[cfg(target_arch = "aarch64")]
     ArmSve = KVM_CAP_ARM_SVE,
+    #[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
+    ArmPtrAuthAddress = KVM_CAP_ARM_PTRAUTH_ADDRESS,
+    #[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
+    ArmPtrAuthGeneric = KVM_CAP_ARM_PTRAUTH_GENERIC,
 }
--- a/src/cap.rs
+++ b/src/cap.rs
@@ -20,6 +20,7 @@ use kvm_bindings::*;
 // We are allowing docs to be missing here because this enum is a wrapper
 // over auto-generated code.
 #[allow(missing_docs)]
+#[derive(PartialEq, Eq)]
 pub enum Cap {
     Irqchip = KVM_CAP_IRQCHIP,
     Hlt = KVM_CAP_HLT,
--- a/src/ioctls/vcpu.rs
+++ b/src/ioctls/vcpu.rs
@@ -2785,4 +2785,23 @@ mod tests {
         assert!(vcpu.has_device_attr(&dist_attr).is_ok());
         assert!(vcpu.set_device_attr(&dist_attr).is_ok());
     }
+
+    #[test]
+    #[cfg(target_arch = "aarch64")]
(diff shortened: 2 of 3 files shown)


src/ioctls/vcpu.rs
```diff
@@ -2729,4 +2729,23 @@ mod tests {
...
+        let vm = kvm.create_vm().unwrap();
+        let vcpu = vm.create_vcpu(0).unwrap();
+
+        let mut kvi: kvm_bindings::kvm_vcpu_init = kvm_bindings::kvm_vcpu_init::default();
+        vm.get_preferred_target(&mut kvi)
+            .expect("Cannot get preferred target");
+        if kvm.check_extension(Cap::ArmPtrAuthAddress) {
+            kvi.features[0] |= 1 << kvm_bindings::KVM_ARM_VCPU_PTRAUTH_ADDRESS;
+        }
+        if kvm.check_extension(Cap::ArmPtrAuthAddress) {
```
roypat: I think you meant

```suggestion
        if kvm.check_extension(Cap::ArmPtrAuthGeneric) {
```

otherwise the if's can be merged.

src/cap.rs
```diff
@@ -148,4 +148,6 @@ pub enum Cap {
     DebugHwBps = KVM_CAP_GUEST_DEBUG_HW_BPS,
     DebugHwWps = KVM_CAP_GUEST_DEBUG_HW_WPS,
     GetMsrFeatures = KVM_CAP_GET_MSR_FEATURES,
+    ArmPtrAuthAddress = KVM_CAP_ARM_PTRAUTH_ADDRESS,
```
kalyazin, line 155: just to doublecheck, don't we need the aarch64 guard here?

src/ioctls/vcpu.rs
bchalios: oh, you 're right. I did!

src/cap.rs
bchalios: I was confused too at the beginning. But these values are present in the [x86 bindings](https://github.com/rust-vmm/kvm-bindings/blob/aa57a1b5299d03b4814ece77f61d45fb04e2238a/src/x86_64/bindings.rs#L598) as well.

src/ioctls/vcpu.rs
bchalios: fixed!

src/cap.rs
jiangliu: Only arm64 code in linux kernel actually uses this constant.
I feel it's a limitation of the way we generate x86 bindings, it shouldn't be used on x86 at all.

bchalios: I don't think we're doing something wrong with how we generate bindings. KVM defines this regardless of architecture: https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/kvm.h#L1136

bchalios: Either way, I updated the PR to allow the definition only in ARM platforms.

roypat (dismissed):

roypat (approved):

JonathanWoollett-Light (approved):

PR #230, opened 2023-07-17
feat: additional method to query kvm extension

ShadowCurse:
### Summary of the PR

Added method to query kvm extension using raw integer instead of an enum value.

### Requirements

requirements:

- [ ] All commits in this PR are signed (with `git commit -s`), and the commit
  description line.
  test.

### DIFF
--- a/src/ioctls/system.rs
+++ b/src/ioctls/system.rs
@@ -183,6 +183,31 @@ impl Kvm {
     ///
     /// # Arguments
     ///
+    /// * `c` - KVM capability to check in a form of a raw integer.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use kvm_ioctls::Kvm;
+    /// # use std::os::raw::c_ulong;
+    /// use kvm_ioctls::Cap;
+    ///
+    /// let kvm = Kvm::new().unwrap();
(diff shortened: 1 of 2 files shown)


JonathanWoollett-Light (dismissed):

src/ioctls/system.rs
```diff
@@ -176,6 +176,31 @@ impl Kvm {
...
+    /// use kvm_ioctls::Cap;
+    ///
+    /// let kvm = Kvm::new().unwrap();
+    /// assert!(kvm.check_extension_raw(Cap::MaxVcpuId as c_ulong) > 0);
+    /// ```
+    pub fn check_extension_raw(&self, c: c_ulong) -> i32 {
+        // SAFETY: Safe because we know that our file is a KVM fd.
+        // If `c` is not a known kernel extension, kernel will return 0.
+        unsafe { ioctl_with_val(self, KVM_CHECK_EXTENSION(), c) }
+    }
```
roypat, line 202: Can we also add a test for the "kernel will return 0" case if we pass in nonsense?

ShadowCurse: Added

roypat (approved):

JonathanWoollett-Light (approved):

PR #244, opened 2023-11-21
Add support for coalesced MMIO (`KVM_CAP_COALESCED_MMIO` /  `KVM_CAP_COALESCED_PIO`)

00xc:
### Summary of the PR

Add support for coalesced MMIO. This performance feature allows guest writes to port and memory space to not trigger VM exits. Instead, the kernel will write an entry into a shared ring buffer for each access, which userspace must consume. The ring buffer is mapped at a certain offset in the vcpu's file descriptor.

In order to enable this capability, introduce the `KvmCoalescedIoRing` struct, which will act as a safe wrapper around the raw mapping of the ring buffer. Since users may not want to use coalesced MMIO, or it might not be available, store it as an `Option` in the `VcpuFd` struct.

Add two tests as well for the newly added code.

The public API consists of:
* A couple of capabilities:
    - `Cap::CoalescedMmio` (`KVM_CAP_COALESCED_MMIO`)
    - `Cap::CoalescedPio` (`KVM_CAP_COALESCED_PIO`).
* Methods to register and unregister addresses that will use this feature:
    - `VmFd::register_coalesced_mmio()` (`KVM_REGISTER_COALESCED_MMIO` ioctl).
    - `VmFd::unregister_coalesced_mmio()` (`KVM_UNREGISTER_COALESCED_MMIO` ioctl).
* A shared ring buffer (mapped with `VcpuFd::map_coalesced_mmio_ring()`).
* A safe method to read from the ring (`VcpuFd::read_coalesced_mmio()`).

### Requirements

requirements:

- [X] All commits in this PR are signed (with `git commit -s`), and the commit
  description line.
- [X] All added/changed functionality has a corresponding unit/integration
  test.
- [X] All added/changed public-facing functionality has entries in the "Upcoming 
- [X] Any newly added `unsafe` code is properly documented.

### DIFF
--- a/src/ioctls/mod.rs
+++ b/src/ioctls/mod.rs
@@ -5,10 +5,13 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the THIRD-PARTY file.
 
+use std::mem::size_of;
 use std::os::unix::io::AsRawFd;
 use std::ptr::null_mut;
 
-use kvm_bindings::kvm_run;
+use kvm_bindings::{
+    kvm_coalesced_mmio, kvm_coalesced_mmio_ring, kvm_run, KVM_COALESCED_MMIO_PAGE_OFFSET,
+};
 use vmm_sys_util::errno;
 
 /// Wrappers over KVM device ioctls.
@@ -25,6 +28,100 @@ pub mod vm;
 /// This typedef is generally used to avoid writing out errno::Error directly and
 /// is otherwise a direct mapping to Result.
 pub type Result<T> = std::result::Result<T, errno::Error>;
+
+/// A wrapper around the coalesced MMIO ring page.
+#[derive(Debug)]
+pub(crate) struct KvmCoalescedIoRing {
+    addr: *mut kvm_coalesced_mmio_ring,
+    page_size: usize,
+}
+
+impl KvmCoalescedIoRing {
+    /// Maps the coalesced MMIO ring from the vCPU file descriptor.
+    pub(crate) fn mmap_from_fd<F: AsRawFd>(fd: &F) -> Result<Self> {
+        // SAFETY: We trust the sysconf libc function and we're calling it
+        // with a correct parameter.
+        let page_size = match unsafe { libc::sysconf(libc::_SC_PAGESIZE) } {
+            -1 => return Err(errno::Error::last()),
+            ps => ps as usize,
+        };
+
+        let offset = KVM_COALESCED_MMIO_PAGE_OFFSET * page_size as u32;
+        // SAFETY: KVM guarantees that there is a page at offset
+        // KVM_COALESCED_MMIO_PAGE_OFFSET * PAGE_SIZE if the appropriate
+        // capability is available. If it is not, the call will simply
+        // fail.
+        let addr = unsafe {
+            libc::mmap(
+                null_mut(),
+                page_size,
+                libc::PROT_READ | libc::PROT_WRITE,
+                libc::MAP_SHARED,
+                fd.as_raw_fd(),
+                offset.into(),
+            )
+        };
+        if addr == libc::MAP_FAILED {
+            return Err(errno::Error::last());
+        }
+        Ok(Self {
+            addr: addr.cast(),
+            page_size,
+        })
+    }
+
+    /// Compute the size of the MMIO ring.
+    /// Taken from [include/uapi/linux/kvm.h](https://elixir.bootlin.com/linux/v6.6/source/include/uapi/linux/kvm.h#L562)
+    const fn ring_max(&self) -> usize {
+        (self.page_size - size_of::<kvm_coalesced_mmio_ring>()) / size_of::<kvm_coalesced_mmio>()
+    }
+
+    /// Gets a mutable reference to the ring
+    fn ring_mut(&mut self) -> &mut kvm_coalesced_mmio_ring {
+        // SAFETY: We have a `&mut self` and the pointer is private, so this
+        // access is exclusive.
+        unsafe { &mut *self.addr }
+    }
+
+    /// Reads a single entry from the MMIO ring.
+    ///
+    /// # Returns
+    ///
+    /// An entry from the MMIO ring buffer, or [`None`] if the ring is empty.
+    pub(crate) fn read_entry(&mut self) -> Option<kvm_coalesced_mmio> {
+        let ring_max = self.ring_max();
+
+        let ring = self.ring_mut();
+        if ring.first == ring.last {
+            return None;
+        }
+
+        let entries = ring.coalesced_mmio.as_ptr();
+        // SAFETY: `ring.first` is an `u32` coming from mapped memory filled
+        // by the kernel, so we trust it. `entries` is a pointer coming from
+        // mmap(), so pointer arithmetic cannot overflow. We have a `&mut self`,
+        // so nobody else has access to the contents of the pointer.
+        let elem = unsafe { entries.add(ring.first as usize).read() };
+        ring.first = (ring.first + 1) % ring_max as u32;
+
+        Some(elem)
+    }
+}
+
+impl Drop for KvmCoalescedIoRing {
+    fn drop(&mut self) {
+        // SAFETY: This is safe because we mmap the page ourselves, and nobody
+        // else is holding a reference to it.
+        unsafe {
+            libc::munmap(self.addr.cast(), self.page_size);
+        }
+    }
+}
+
+// SAFETY: See safety comments about [`KvmRunWrapper`].
+unsafe impl Send for KvmCoalescedIoRing {}
+// SAFETY: See safety comments about [`KvmRunWrapper`].
+unsafe impl Sync for KvmCoalescedIoRing {}
 
 /// Safe wrapper over the `kvm_run` struct.
 ///
--- a/src/ioctls/vcpu.rs
+++ b/src/ioctls/vcpu.rs
@@ -10,7 +10,7 @@ use libc::EINVAL;
 use std::fs::File;
 use std::os::unix::io::{AsRawFd, RawFd};
 
-use crate::ioctls::{KvmRunWrapper, Result};
+use crate::ioctls::{KvmCoalescedIoRing, KvmRunWrapper, Result};
 use crate::kvm_ioctls::*;
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
 use kvm_bindings::{
@@ -169,6 +169,8 @@ pub enum VcpuExit<'a> {
 pub struct VcpuFd {
     vcpu: File,
     kvm_run_ptr: KvmRunWrapper,
+    /// A pointer to the coalesced MMIO page
+    coalesced_mmio_ring: Option<KvmCoalescedIoRing>,
 }
 
 /// KVM Sync Registers used to tell KVM which registers to sync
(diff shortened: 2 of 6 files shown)


src/ioctls/mod.rs
```diff
@@ -26,6 +29,100 @@ pub mod vm;
...
+                libc::MAP_SHARED,
+                fd.as_raw_fd(),
+                offset.into(),
+            )
+        };
+        if addr == libc::MAP_FAILED {
+            return Err(errno::Error::last());
+        }
+        Ok(Self {
+            addr: addr.cast::<kvm_coalesced_mmio_ring>(),
```
JonathanWoollett-Light: Nit: I think it can infer the type of this cast.
```suggestion
            addr: addr.cast(),
```

src/ioctls/mod.rs
```diff
@@ -26,6 +29,100 @@ pub mod vm;
...
+        Some(elem)
+    }
+}
+
+impl Drop for KvmCoalescedIoRing {
+    fn drop(&mut self) {
+        // SAFETY: This is safe because we mmap the page ourselves, and nobody
+        // else is holding a reference to it.
+        unsafe {
+            libc::munmap(self.addr.cast::<libc::c_void>(), self.page_size);
```
JonathanWoollett-Light: Nit: I think it can infer the type of this cast.
```suggestion
            libc::munmap(self.addr.cast(), self.page_size);
```

src/ioctls/vcpu.rs
```diff
@@ -1743,6 +1745,56 @@ impl VcpuFd {
...
+    /// }
+    /// ```
+    pub fn map_coalesced_mmio_ring(&mut self) -> Result<()> {
+        if self.coalesced_mmio_ring.is_some() {
+            return Ok(());
+        }
+        let ring = KvmCoalescedIoRing::mmap_from_fd(&self.vcpu)?;
+        self.coalesced_mmio_ring = Some(ring);
+        Ok(())
+    }
```
JonathanWoollett-Light:
```suggestion
    pub fn map_coalesced_mmio_ring(&mut self) -> Result<()> {
        if self.coalesced_mmio_ring.is_none() {
            let ring = KvmCoalescedIoRing::mmap_from_fd(&self.vcpu)?;
            self.coalesced_mmio_ring = Some(ring);
        }
        Ok(())
    }
```

src/ioctls/vcpu.rs
```diff
@@ -1743,6 +1745,56 @@ impl VcpuFd {
...
+    /// * An error if [`map_coalesced_mmio_ring()`](VcpuFd::map_coalesced_mmio_ring)
+    ///   was not called beforehand.
+    /// * [`Ok<None>`] if the ring is empty.
+    /// * [`Ok<Some<kvm_coalesced_mmio>>`] if an entry was successfully read.
+    pub fn coalesced_mmio_read(&mut self) -> Result<Option<kvm_coalesced_mmio>> {
+        let Some(ring) = self.coalesced_mmio_ring.as_mut() else {
+            return Err(errno::Error::new(libc::EIO));
+        };
+        Ok(ring.read_entry())
+    }
```
JonathanWoollett-Light:
```suggestion
    pub fn coalesced_mmio_read(&mut self) -> Result<Option<kvm_coalesced_mmio>> {
        let ring = self.coalesced_mmio_ring.ok_or(errno::Error::new(libc::EIO))?;
        Ok(ring.read_entry())
    }
```
You can also do
```rust
    pub fn coalesced_mmio_read(&mut self) -> Result<Option<kvm_coalesced_mmio>> {
        self.coalesced_mmio_ring
            .ok_or(errno::Error::new(libc::EIO))
            .map(|ring|ring.read_entry())
    }
```

src/ioctls/vcpu.rs
```diff
@@ -2826,4 +2883,199 @@ mod tests {
...
+        const DATA: u64 = 0x39;
+        const SIZE: u32 = 1;
+
+        #[rustfmt::skip]
+        let code = [
+            0xe6, 0x2c,   // out 0x2c, al
+            0xf4,         // hlt
+            0xe6, 0x2c,   // out 0x2c, al
+            0xf4,         // hlt
+        ];
```
JonathanWoollett-Light, line 3151: Can we use constants for the mnemonics?

E.g.
```suggestion
        use iced_x86::Mnemonic::{Hlt,Out};
        
        #[rustfmt::skip]
        let code = [
            Out as u8, 0x2c,
            Hlt as u8,
            Out as u8, 0x2c,
            Hlt as u8,
        ];
```

src/ioctls/vcpu.rs
```diff
@@ -2826,4 +2883,199 @@ mod tests {
...
+        let vm = kvm.create_vm().unwrap();
+        if !vm.check_extension(Cap::CoalescedPio) {
+            return;
+        }
+
+        // Prepare guest memory
+        let mem_size = 0x4000;
+        let load_addr = mmap_anonymous(mem_size);
+        let guest_addr: u64 = 0x1000;
+        let slot: u32 = 0;
```
JonathanWoollett-Light:
```suggestion
        let slot = 0;
```

src/ioctls/vcpu.rs
```diff
@@ -2826,4 +2883,199 @@ mod tests {
...
+        let code = [
+            0x66, 0x31, 0xFF,        // xor di,di
+            0x66, 0xBF, 0x24, 0x01,  // mov di, 0x124
+            0x67, 0x66, 0x89, 0x05,  // mov WORD PTR [di], ax
+            0xF4,                    // hlt
+            0x66, 0x31, 0xFF,        // xor di,di
+            0x66, 0xBF, 0x24, 0x01,  // mov di, 0x124
+            0x67, 0x66, 0x89, 0x05,  // mov WORD PTR [di], ax
+            0xF4,                    // hlt
+        ];
```
JonathanWoollett-Light, line 3248: Can we use constants for the mnemonics?

E.g.
```suggestion
        use iced_x86::Mnemonic::{Xor, Mov, Hlt};
        
        #[rustfmt::skip]
        let code = [
            Xor as u8, 0x31, 0xFF,        // xor di,di
            Mov as u8, 0xBF, 0x24, 0x01,  // mov di, 0x124
            Mov as u8, 0x66, 0x89, 0x05,  // mov WORD PTR [di], ax
            Hlt as u8,                    // hlt
            Xor as u8, 0x31, 0xFF,        // xor di,di
            Mov as u8, 0xBF, 0x24, 0x01,  // mov di, 0x124
            Mov as u8, 0x66, 0x89, 0x05,  // mov WORD PTR [di], ax
            Hlt as u8,                    // hlt
        ];
```

00xc:
@JonathanWoollett-Light addressed your suggestions in the newer version, except the ones about using function mnemonics. Regarding that, isn't that change out of scope for this PR? It would add a new dependency and we would still have a bunch of tests which use raw bytes as assembly code. I'd rather work on that change in a separate PR.

00xc:
By the way the code can also properly be assembled with that crate's `CodeAssembler`, e.g.

```rust
let mut asm = CodeAssembler::new(16).unwrap();
asm.out(0x2c, al).unwrap();
asm.hlt().unwrap();
asm.out(0x2c, al).unwrap();
asm.hlt().unwrap();
let code = asm.assemble(base_addr).unwrap();
```

src/ioctls/mod.rs
```diff
@@ -26,6 +29,100 @@ pub mod vm;
...
+    const fn ring_max(&self) -> usize {
+        (self.page_size - size_of::<kvm_coalesced_mmio_ring>()) / size_of::<kvm_coalesced_mmio>()
+    }
+
+    /// Gets a mutable reference to the ring
+    fn ring_mut(&mut self) -> &mut kvm_coalesced_mmio_ring {
+        // SAFETY: We have a `&mut self` and the pointer is private, so this
+        // access is exclusive.
+        unsafe { &mut *self.addr }
+    }
```
roypat, line 84: Just to be sure, since access to this only happens on a KVM_EXIT path, KVM will also not concurrently modify the mmap'd memory region while we hold this &mut?

00xc: As far as I know KVM will not append entries to the ring buffer while the vCPU is stopped. That would make an implementation very difficult, even in C, as the ring buffer is not atomic.

roypat: Sounds good, thanks for clarifying!

src/ioctls/vcpu.rs
```diff
@@ -2826,4 +2882,199 @@ mod tests {
...
+            0xe6, 0x2c,   // out 0x2c, al
+            0xf4,         // hlt
+            0xe6, 0x2c,   // out 0x2c, al
+            0xf4,         // hlt
+        ];
+
+        let kvm = Kvm::new().unwrap();
+        let vm = kvm.create_vm().unwrap();
+        if !vm.check_extension(Cap::CoalescedPio) {
+            return;
```
roypat: Can we replace this `return` with a `panic`, to make sure we're actually running the test in CI?

src/ioctls/vcpu.rs
```diff
@@ -2826,4 +2882,199 @@ mod tests {
...
+            0x66, 0x31, 0xFF,        // xor di,di
+            0x66, 0xBF, 0x24, 0x01,  // mov di, 0x124
+            0x67, 0x66, 0x89, 0x05,  // mov WORD PTR [di], ax
+            0xF4,                    // hlt
+        ];
+
+        let kvm = Kvm::new().unwrap();
+        let vm = kvm.create_vm().unwrap();
+        if !vm.check_extension(Cap::CoalescedMmio) {
+            return;
```
roypat: Same here

JonathanWoollett-Light:
> @JonathanWoollett-Light addressed your suggestions in the newer version, except the ones about using function mnemonics. Regarding that, isn't that change out of scope for this PR? It would add a new dependency and we would still have a bunch of tests which use raw bytes as assembly code. I'd rather work on that change in a separate PR.

Yeah this makes sense.

src/ioctls/mod.rs
00xc: Just to elaborate:
1. qemu seems to work under this assumption (see [`kvm_flush_coalesced_mmio_buffer()`](https://github.com/qemu/qemu/blob/8fa379170c2a12476021f5f50d6cf3f672e79e7b/accel/kvm/kvm-all.c#L2663))
2. This feature is not very well documented in KVM, but briefly looking at the kernel code, KVM simply avoids exiting to user space when it gets a VM exit with a write to an address registered for coalesced I/O. In other words, updates to the ring buffer do not happen on a vCPU that has exited to userspace, as there are no further VM exits.

src/ioctls/vcpu.rs
00xc: Fixed in the new version.

src/ioctls/vcpu.rs
00xc: Fixed in the new version.

roypat (dismissed):

00xc:
Rebased on main and fixed conflicts

roypat (dismissed):

00xc:
Gentle ping for reviewers :)

JonathanWoollett-Light:
@00xc Can you please resolve the merge conflicts, after this LGTM.

00xc:
Rebased on main and fixed conflicts

roypat (approved):

JonathanWoollett-Light (approved):

PR #253, opened 2024-01-24
ioctls/vm: Change kvm_create_device to immutable reference

tylerfanelli:
### Summary of the PR

The `device` argument to `VmFd.create_device()` is not modified. With this, the argument does not need to be a mutable reference. This PR changes the `device` argument to an immutable reference and updates the tests + docs to reflect this change.

Although this is public facing, I wasn't sure if this required an entry in CHANGELOG.md. If this is approved and someone feels that it should be an entry, please let me know.

### DIFF
--- a/src/ioctls/vm.rs
+++ b/src/ioctls/vm.rs
@@ -1188,7 +1188,7 @@ impl VmFd {
     ///     }
     /// });
     /// ```
-    pub fn create_device(&self, device: &mut kvm_create_device) -> Result<DeviceFd> {
+    pub fn create_device(&self, device: &kvm_create_device) -> Result<DeviceFd> {
         // SAFETY: Safe because we are calling this with the VM fd and we trust the kernel.
         let ret = unsafe { ioctl_with_ref(self, KVM_CREATE_DEVICE(), device) };
         if ret == 0 {
@@ -1655,11 +1655,11 @@ pub(crate) fn create_gic_device(vm: &VmFd, flags: u32) -> DeviceFd {
         fd: 0,
         flags,
     };
-    match vm.create_device(&mut gic_device) {
+    match vm.create_device(&gic_device) {
         Ok(fd) => fd,
         Err(_) => {
             gic_device.type_ = kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V2;
-            vm.create_device(&mut gic_device)
+            vm.create_device(&gic_device)
                 .expect("Cannot create KVM vGIC device")
         }
     }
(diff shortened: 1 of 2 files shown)


src/ioctls/vm.rs
```diff
@@ -1177,18 +1177,18 @@ impl VmFd {
...
     ///     {
     ///         device.type_ = kvm_device_type_KVM_DEV_TYPE_ARM_VGIC_V2;
-    ///         vm.create_device(&mut device)
+    ///         vm.create_device(&device)
     ///             .expect("Cannot create vGIC device")
     ///     }
     /// });
     /// ```
-    pub fn create_device(&self, device: &mut kvm_create_device) -> Result<DeviceFd> {
+    pub fn create_device(&self, device: &kvm_create_device) -> Result<DeviceFd> {
```
roypat, line 1191: Mh, maybe I'm missing some context here, but wouldnt the below be a better signature for this function?

```suggestion
    pub fn create_device(&mut self, device: kvm_create_device) -> Result<DeviceFd> {
```

We take `self` by mutable reference, because we're modifying the virtual machine that `self` represents, and we take ownership of the `kvm_create_device` object, because in the success case we return ownership over a `File` created from the `fd` inside the `kvm_create_device`, so we "transfer" ownership here

roypat (commented):
Thank you for your contribution! Could you have a look at the comment below? (Also, yeah, this would require a changelog entry under `Changed` :)

src/ioctls/vm.rs
roypat: Ah, sorry, I did indeed misunderstand what is going on here. The `KVM_CREATE_DEVICE` ioctl writes to the `fd` field of the `kvm_create_device` structure, so that should indeed be passed by mutable reference (since it's getting mutated). The error in the code is that it uses `ioctl_with_ref` instead of `ioctl_with_mut_ref`.

tylerfanelli: So if I'm understanding correctly, instead of modifying `kvm_create_device` to be a immutable reference, the actual change is that it should call `ioctl_with_mut_ref` (since the ioctl is what will change data in `kvm_create_device`).

Is that accurate? This change was submitted because `clippy` was complaining about the mutable reference. I was missing the underlying context that `ioctl_with_mut_ref` should be called.

roypat: Yep, that's right :)
There actually seem to be a few more instances of this (e.g. where `ioctl_with_ref` is used for an ioctl that actually mutates its input), we should fix those too (although we can do that outside of this PR if you want)

tylerfanelli: I can close this PR, and will submit some changes on another. Thanks for clearing this up.

PR #255, opened 2024-02-13
Fix soundness issues via `KvmRunWrapper::as_mut_ref()`

### LINKED ISSUE #248: `VcpuFd::set_kvm_immediate_exit` may lead to data races/undefined behavior
Currently the `VcpuFd` structure implements both `Send` and `Sync` so it's safe to share across threads, but this method enables mutation of shared memory. The signature for this method takes `&self` which means it can be invoked concurrently on multiple threads. Its current implementation:

```rust
    pub fn set_kvm_immediate_exit(&self, val: u8) {
        let kvm_run = self.kvm_run_ptr.as_mut_ref();
        kvm_run.immediate_exit = val;
    }
```

currently can expose two issues in Rust's memory model. The first is that the [`as_mut_ref`](https://github.com/rust-vmm/kvm-ioctls/blob/c61285025ab6f388fa8f23551426c8f0778e7cd9/src/ioctls/mod.rs#L77-L86) function cannot be sound because it's promoting a safe shared reference to a safe mutable reference. Mutable references guarantee exclusivitity but that's not the case here with multiple threads. The second issue is that the write here is a non-atomic write as opposed to an atomic write, which means that multiple threads can do non-atomic writes causing a data race.

I found this when poking around this crate (thanks for providing it!). I'm still very new to this space and am trying to get my bearings. I was planning on using this to set this flag from another thread to assist with interrupting a VM, which is why I was curious how it worked internally given its signature.

I think that the fix here is relatively simple, just restructuring things internally. In theory `as_mut_ref()` should be removed entirely (that method signature can't ever be sound in Rust) and instead code can use `ptr::addr_of` to convert from `*mut kvm_run` to `*mut u8` for the flag here. That could then be casted to `AtomicU8` and the store done through that.

00xc:
### Summary of the PR

As detailed in #248, and since `KvmRunWrapper` implements `Send` and `Sync`, `as_mut_ref()` can cause undefined behavior, as two threads can acquire a mutable reference to the `kvm_run` struct via an immutable reference to `KvmRunWrapper`.

Fix this by making `KvmRunWrapper::as_mut_ref()` take `&mut self`, which also gets rid of a clippy warning suppression, and update the callers. This results in potentially breaking changes in the public interface, as several `VcpuFd` methods now take `&mut self` as well.

Fixes: #248 

### Requirements

requirements:

  test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ## Changed
 
+- [[#255](https://github.com/rust-vmm/kvm-ioctls/issues/255)]: Fixed a
+  soundness issue when accessing the `kvm_run` struct. `VcpuFd::run()` and
+  `VcpuFd::set_kvm_immediate_exit()` now take `&mut self` as a consequence.
+
 # v0.16.0
 
 ## Added
--- a/src/ioctls/mod.rs
+++ b/src/ioctls/mod.rs
@@ -183,6 +183,14 @@ impl KvmRunWrapper {
     }
 }
 
+impl AsRef<kvm_run> for KvmRunWrapper {
+    fn as_ref(&self) -> &kvm_run {
+        // SAFETY: Safe because we know we mapped enough memory to hold the kvm_run struct because
+        // the kernel told us how large it was.
+        unsafe { &*(self.kvm_run_ptr as *const kvm_run) }
+    }
+}
+
 impl Drop for KvmRunWrapper {
     fn drop(&mut self) {
         // SAFETY: This is safe because we mmap the area at kvm_run_ptr ourselves,
--- a/src/ioctls/mod.rs
+++ b/src/ioctls/mod.rs
@@ -172,8 +172,7 @@ impl KvmRunWrapper {
     }
 
     /// Returns a mutable reference to `kvm_run`.
-    #[allow(clippy::mut_from_ref)]
-    pub fn as_mut_ref(&self) -> &mut kvm_run {
+    pub fn as_mut_ref(&mut self) -> &mut kvm_run {
         #[allow(clippy::cast_ptr_alignment)]
         // SAFETY: Safe because we know we mapped enough memory to hold the kvm_run struct because
         // the kernel told us how large it was.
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 ## Changed
 
+- [[#255](https://github.com/rust-vmm/kvm-ioctls/issues/255)]: Fixed a
+  soundness issue when accessing the `kvm_run` struct.
+
 # v0.16.0
```
roypat: Could we list all the functions that had their signature changed here?

00xc: Done

roypat (approved):
Thanks for fixing this :)

00xc:
Ping for the remaining reviewers.

JonathanWoollett-Light:
This is a breaking change so it would be good to check if anyone has any other concerns, otherwise LGTM. I've messaged in [the rust-vmm public slack](https://join.slack.com/t/rust-vmm/shared_invite/enQtODAxMzA2ODIyMTc2LWRhYjIwZmQ0YzUxODJlMTRhZWU2ZDBjYmJiNzBmOWVmYjg4MjY5YWRjYjM0YzQ5YzgyMTBmYzNlMjMzYmZlODU), if there are no concerns after some time I will approve.

andreeaflorescu (approved):

00xc:
Perhaps there should be a security advisory for this. Producing undefined behavior is pretty simple:

```rust
let vcpu = ...;

// Suppose this is a VcpuExit that contains an immutable
// reference to kvm_run, e.g. IoOut, MmioWrite
let exit = vcpu.run().unwrap();

// Suppose this is a VcpuExit of the same type as above.
// This modifies kvm_run again.
vcpu.run().unwrap();

// The value `exit` immutably references has been modified,
// yet the compiler does not complain because `run()` takes
// `&self`. Using `exit` here is UB.
println!("{:?}", exit);
```

Another way to think about it is that mutable accesses *must* invalidate all previous immutable references.

JonathanWoollett-Light:
@00xc if you think there is a security vulnerability please report it via GItHub security vulnerability reporting. https://github.com/rust-vmm/kvm-ioctls/security/advisories

PR #266, opened 2024-06-04
Add test for vcpu_init with an invalid target

### LINKED ISSUE #265: Assert failing when `MIDR_EL1` is `0x413fd0c1`
The assert at https://github.com/rust-vmm/kvm-ioctls/blob/0ddf7ef6f6079bf217cb297278b7ba35e3e70116/src/ioctls/vcpu.rs#L2732 is supposing to fail when setting target to `0`. However, when MIDR_EL1 is `0x413fd0c1`, the test does not fail since `0` corresponds to `KVM_ARM_TARGET_AEM_V8`, which is a valid target. If the test is supposed to always fail, shall not the initial value be replaced with an invalid value like `KVM_ARM_NUM_TARGETS`? (see arch/arm64/kvm/guest.c::kvm_target_cpu())

Thanks.

MatiasVara:
### Summary of the PR

To test vcpu_init, replace the zeroed value with the invalid value KVM_ARM_NUM_TARGETS. Put the test in a separate test from the test_get_preferred_target test. This fixes issue #265. 

### Requirements

requirements:

  test.

### DIFF
--- a/src/ioctls/vcpu.rs
+++ b/src/ioctls/vcpu.rs
@@ -2646,6 +2646,23 @@ mod tests {
 
     #[test]
     #[cfg(target_arch = "aarch64")]
+    fn test_faulty_vcpu_target_aarch64() {
+        let kvm = Kvm::new().unwrap();
+        let vm = kvm.create_vm().unwrap();
+        let vcpu = vm.create_vcpu(0).unwrap();
+
+        // KVM defines valid targets as 0 to KVM_ARM_NUM_TARGETS-1, so pick a big raw number
+        // greater than that as target to be invalid
+        let kvi: kvm_bindings::kvm_vcpu_init = kvm_bindings::kvm_vcpu_init {
+            target: 300,
+            ..Default::default()
+        };
+
+        assert!(vcpu.vcpu_init(&kvi).is_err());
+    }
+
+    #[test]
+    #[cfg(target_arch = "aarch64")]
     fn test_faulty_vcpu_fd_aarch64() {
         use std::os::unix::io::FromRawFd;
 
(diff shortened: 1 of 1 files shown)


src/ioctls/vcpu.rs
```diff
@@ -2644,6 +2644,22 @@ mod tests {
...
+    fn test_faulty_vcpu_target_aarch64() {
+        let kvm = Kvm::new().unwrap();
+        let vm = kvm.create_vm().unwrap();
+        let vcpu = vm.create_vcpu(0).unwrap();
+
+        let mut kvi: kvm_bindings::kvm_vcpu_init = kvm_bindings::kvm_vcpu_init {
+            target: KVM_ARM_NUM_TARGETS,
+            ..Default::default()
+        };
+        kvi.target = KVM_ARM_NUM_TARGETS;
```
ShadowCurse: I think `kvi.target = KVM_ARM_NUM_TARGETS;` is not needed here (also `mut` is not needed).
Additionally can you add a comment about `KVM_ARM_NUM_TARGETS` representing an incorrect target? Smth like: `"KVM defines valid targets from 0 to 5 and KVM_ARM_NUM_TARGETS has a value 6"`
Or maybe just use some big raw value here instead of a defined constant.

MatiasVara: > I think `kvi.target = KVM_ARM_NUM_TARGETS;` is not needed here (also `mut` is not needed). 

Apologies, I forgot to remove that line.

> Additionally can you add a comment about `KVM_ARM_NUM_TARGETS` representing an incorrect target? Smth like: `"KVM defines valid targets from 0 to 5 and KVM_ARM_NUM_TARGETS has a value 6"` Or maybe just use some big raw value here instead of a defined constant.

Yes, I can.

ShadowCurse (approved):

roypat (approved):

PR #267, opened 2024-06-26
expose fields for KVM_EXIT_HYPERCALL

Freax13:
This is a **breaking change**.

### Summary of the PR

This is needed for userspace to handle the `KVM_HC_MAP_GPA_RANGE` hypercall.

### Requirements

requirements:

- [X] All commits in this PR have Signed-Off-By trailers (with
- [X] All added/changed functionality has a corresponding unit/integration
  test.
- [X] Any newly added `unsafe` code is properly documented.

### DIFF
--- a/src/cap.rs
+++ b/src/cap.rs
@@ -167,5 +167,6 @@ pub enum Cap {
     ArmPtrAuthGeneric = KVM_CAP_ARM_PTRAUTH_GENERIC,
     #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
     X86UserSpaceMsr = KVM_CAP_X86_USER_SPACE_MSR,
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
     ExitHypercall = KVM_CAP_EXIT_HYPERCALL,
 }
--- a/src/cap.rs
+++ b/src/cap.rs
@@ -167,4 +167,5 @@ pub enum Cap {
     ArmPtrAuthGeneric = KVM_CAP_ARM_PTRAUTH_GENERIC,
     #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
     X86UserSpaceMsr = KVM_CAP_X86_USER_SPACE_MSR,
+    ExitHypercall = KVM_CAP_EXIT_HYPERCALL,
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Added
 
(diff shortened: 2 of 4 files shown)


src/cap.rs
```diff
@@ -167,4 +167,5 @@ pub enum Cap {
     ArmPtrAuthGeneric = KVM_CAP_ARM_PTRAUTH_GENERIC,
     #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
     X86UserSpaceMsr = KVM_CAP_X86_USER_SPACE_MSR,
+    ExitHypercall = KVM_CAP_EXIT_HYPERCALL,
```
00xc, line 171: I think this is only defined for x86_64 and arm64. From [kvm-bindings](https://github.com/rust-vmm/kvm-bindings):

```
$ git grep KVM_CAP_EXIT_HYPERCALL
src/arm64/bindings.rs:pub const KVM_CAP_EXIT_HYPERCALL: u32 = 201;
src/x86_64/bindings.rs:pub const KVM_CAP_EXIT_HYPERCALL: u32 = 201;
```

00xc: Interestingly, the docs say this is only available for x86(-64):
https://docs.kernel.org/virt/kvm/api.html#kvm-cap-exit-hypercall

Freax13: Isn't that because `kvm-bindings` only supports x86_64 and arm64? I can't find definitions for any `KVM_CAP_*` outside of those two files.

AFAICT [`KVM_CAP_EXIT_HYPERCALL`](https://github.com/torvalds/linux/blob/55027e689933ba2e64f3d245fb1ff185b3e7fc81/include/uapi/linux/kvm.h#L885) is defined in a non-architecture-specific header.

That said KVM currently only implements `KVM_CAP_EXIT_HYPERCALL` on x86_64, so I'd also be fine with restricting it to that platform.

Freax13: > Interestingly, the docs say this is only available for x86(-64): https://docs.kernel.org/virt/kvm/api.html#kvm-cap-exit-hypercall

If that's what to official docs say, that seems like a good reason to restrict it to x86_64 :)

00xc: > Isn't that because `kvm-bindings` only supports x86_64 and arm64? I can't find definitions for any `KVM_CAP_*` outside of those two files.

Ah yes, that could be it.

00xc (approved):

ShadowCurse (approved):

roypat (approved):

PR #270, opened 2024-08-01
Propagate MemoryFault exit reason in KVM_RUN

kalyazin:
### Summary of the PR

- Add [`KVM_CAP_MEMORY_FAULT_INFO`](https://docs.kernel.org/virt/kvm/api.html#kvm-cap-memory-fault-info) variant to `Cap`
- Propagate `KVM_EXIT_MEMORY_FAULT` exit reason in [`KVM_RUN`](https://docs.kernel.org/virt/kvm/api.html#kvm-run).

### Requirements

requirements:

- ~~[ ] All added/changed functionality has a corresponding unit/integration
  test.~~
- ~~[ ] Any newly added `unsafe` code is properly documented.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 - [[#267](https://github.com/rust-vmm/kvm-ioctls/pull/267)]: Added `HypercallExit` field to `VcpuExit::Hypercall` and added `ExitHypercall` to `Cap`.
+- [[#270](https://github.com/rust-vmm/kvm-ioctls/pull/270)]: Added `MemoryFaultInfo` to `Cap` and propagated `MemoryFault` exit reason in `KVM_RUN`.
 
 ### Changed
 
--- a/src/ioctls/vcpu.rs
+++ b/src/ioctls/vcpu.rs
@@ -166,6 +166,15 @@ pub enum VcpuExit<'a> {
     X86Rdmsr(ReadMsrExit<'a>),
     /// Corresponds to KVM_EXIT_X86_WRMSR.
     X86Wrmsr(WriteMsrExit<'a>),
+    /// Corresponds to KVM_EXIT_MEMORY_FAULT.
+    MemoryFault {
+        /// flags
+        flags: u64,
+        /// gpa
+        gpa: u64,
+        /// size
+        size: u64,
+    },
     /// Corresponds to an exit reason that is unknown from the current version
     /// of the kvm-ioctls crate. Let the consumer decide about what to do with
     /// it.
@@ -1569,7 +1578,30 @@ impl VcpuFd {
                 r => Ok(VcpuExit::Unsupported(r)),
             }
         } else {
-            Err(errno::Error::last())
+            let errno = errno::Error::last();
+            let run = self.kvm_run_ptr.as_mut_ref();
+            // From https://docs.kernel.org/virt/kvm/api.html#kvm-run :
+            //
+            // KVM_EXIT_MEMORY_FAULT is unique among all KVM exit reasons in that it accompanies
+            // a return code of ‘-1’, not ‘0’! errno will always be set to EFAULT or EHWPOISON
+            // when KVM exits with KVM_EXIT_MEMORY_FAULT, userspace should assume kvm_run.exit_reason
+            // is stale/undefined for all other error numbers.
+            if ret == -1
+                && (errno == errno::Error::new(libc::EFAULT)
+                    || errno == errno::Error::new(libc::EHWPOISON))
+                && run.exit_reason == KVM_EXIT_MEMORY_FAULT
+            {
+                // SAFETY: Safe because the exit_reason (which comes from the kernel) told us
+                // which union field to use.
+                let fault = unsafe { &mut run.__bindgen_anon_1.memory_fault };
+                Ok(VcpuExit::MemoryFault {
+                    flags: fault.flags,
+                    gpa: fault.gpa,
+                    size: fault.size,
+                })
+            } else {
+                Err(errno)
+            }
         }
     }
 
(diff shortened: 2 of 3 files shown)


src/ioctls/vcpu.rs
```diff
@@ -166,6 +166,12 @@ pub enum VcpuExit<'a> {
     X86Rdmsr(ReadMsrExit<'a>),
     /// Corresponds to KVM_EXIT_X86_WRMSR.
     X86Wrmsr(WriteMsrExit<'a>),
+    /// Corresponds to KVM_EXIT_MEMORY_FAULT.
+    MemoryFault(
+        u64, /* flags */
+        u64, /* gpa */
+        u64, /* size */
+    ),
```
roypat: Let's use named fields here 

```suggestion
    MemoryFault {
        flags: u64,
        gpa: u64,
        size: u64,
    },
```

src/ioctls/vcpu.rs
```diff
@@ -1568,6 +1574,17 @@ impl VcpuFd {
...
+            let run = self.kvm_run_ptr.as_mut_ref();
+            match run.exit_reason {
+                KVM_EXIT_MEMORY_FAULT => {
+                    // SAFETY: Safe because the exit_reason (which comes from the kernel) told us
+                    // which union field to use.
+                    let fault = unsafe { &mut run.__bindgen_anon_1.memory_fault };
+                    Ok(VcpuExit::MemoryFault(fault.flags, fault.gpa, fault.size))
+                }
+                _ => Err(errno::Error::last()),
+            }
```
roypat: I'd probably write this as 

```rs
else {
    let errno = errno::Error::last();
    if errno.0 == libc::EFAULT && ret == - 1 {
        let run = ...
    } else {
        Err(errno)
    }
}
```

to avoid multiple calls to `errno::Error::last()`

roypat: Also, can we add a comment here about why this exit reason is so weird (its the only one that comes with a non-zero return value after all!)

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Added
 
 - [[#267](https://github.com/rust-vmm/kvm-ioctls/pull/267)]: Added `HypercallExit` field to `VcpuExit::Hypercall` and added `ExitHypercall` to `Cap`.
+- [[#270](https://github.com/rust-vmm/kvm-ioctls/pull/267)]: Propagate `MemoryFault` exit reason in `KVM_RUN`.
```
roypat: There's an associated capability, `KVM_CAP_MEMORY_FAULT_INFO`, for this, let's include that in this PR too

roypat (commented):
Thanks Nikita :)

src/ioctls/vcpu.rs
kalyazin: done

src/ioctls/vcpu.rs
kalyazin: done

CHANGELOG.md
kalyazin: done

roypat (dismissed):

src/ioctls/vcpu.rs
```diff
@@ -1569,7 +1578,33 @@ impl VcpuFd {
...
+            // KVM_EXIT_MEMORY_FAULT is unique among all KVM exit reasons in that it accompanies
+            // a return code of ‘-1’, not ‘0’! errno will always be set to EFAULT or EHWPOISON
+            // when KVM exits with KVM_EXIT_MEMORY_FAULT, userspace should assume kvm_run.exit_reason
+            // is stale/undefined for all other error numbers.
+            if ret == -1
+                && (errno == errno::Error::new(libc::EFAULT)
+                    || errno == errno::Error::new(libc::EHWPOISON))
+            {
+                let run = self.kvm_run_ptr.as_mut_ref();
+                if run.exit_reason == KVM_EXIT_MEMORY_FAULT {
```
ShadowCurse: Maybe combine these 2 `if`?
```Rust
if ret == -1
  && (errno == errno::Error::new(libc::EFAULT) || errno == errno::Error::new(libc::EHWPOISON)) 
  && self.kvm_run_ptr.as_mut_ref().exit_reason == KVM_EXIT_MEMORY_FAULT
            {
                ...
```
This way we will not have 2 `else`s with `Err(errno)`

kalyazin: ~~in that case we'd have to dereference `self.kvm_run_ptr` twice: once for the condition and once for fault data. Do you think it's better than 2 elses?~~

kalyazin: done

src/ioctls/vcpu.rs
```diff
@@ -1569,7 +1578,30 @@ impl VcpuFd {
...
+            // From https://docs.kernel.org/virt/kvm/api.html#kvm-run :
+            //
+            // KVM_EXIT_MEMORY_FAULT is unique among all KVM exit reasons in that it accompanies
+            // a return code of ‘-1’, not ‘0’! errno will always be set to EFAULT or EHWPOISON
+            // when KVM exits with KVM_EXIT_MEMORY_FAULT, userspace should assume kvm_run.exit_reason
+            // is stale/undefined for all other error numbers.
+            if ret == -1
+                && (errno == errno::Error::new(libc::EFAULT)
+                    || errno == errno::Error::new(libc::EHWPOISON)
+                        && run.exit_reason == KVM_EXIT_MEMORY_FAULT)
```
ShadowCurse: Shouldn't `errno ==` checks be in their own parenthesis?
E.g. 
```
if (ret == -1)
   && (errno == ... || errno == )
   && (run.exit_reason == ...)
```
because now second `errno == ` check will be `and`ed with `run.exit_reason` check before `or`d with first `errno == ` check.

kalyazin: thanks, fixed

ShadowCurse (approved):

roypat (approved):

PR #272, opened 2024-08-12
Fix IO safety runtime errors in tests

alyssais:
### Summary of the PR

If debug_assertions is enabled, [Rust 1.80.0 and later will abort the process when attempting to close a file descriptor that is already closed][1].  This means that, in tests, where an invalid file descriptor object is deliberately constructed, we have to avoid that object being dropped.  Using `into_raw_fd()` allows the object's memory to be freed (unlike `std::mem::forget()`), without attempting to close the file descriptor.

[1]: https://github.com/rust-lang/rust/commit/cb4940645775f60d74aee2e018d6c516c5aa9ed7

### Requirements

requirements:

  test.

### DIFF
--- a/src/ioctls/system.rs
+++ b/src/ioctls/system.rs
@@ -730,6 +730,7 @@ mod tests {
     #![allow(clippy::undocumented_unsafe_blocks)]
     use super::*;
     use libc::{fcntl, FD_CLOEXEC, F_GETFD};
+    use std::os::fd::IntoRawFd;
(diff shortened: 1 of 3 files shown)


roypat (approved):
Thanks for this! I'm using this as an opportunity to also update our CI toolchain to 1.80.1 (https://github.com/rust-vmm/rust-vmm-container/pull/111) so that we (a) don't regress on these again, and (b) potentially find and fix other issues before downstream users discover them :)

ShadowCurse (approved):

PR #275, opened 2024-08-16
Introduce `riscv64` ioctls

TimePrinciple:
### Summary of the PR

Add basic support for `riscv64` architecture.

### Requirements

requirements:

  test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Added
 
+- [[#275](https://github.com/rust-vmm/kvm-ioctls/pull/275)]: Introduce `riscv64` ioctls.
+
 ### Changed
 
 - [[#273](https://github.com/rust-vmm/kvm-ioctls/pull/273)]: `DeviceFd::get_device_attr` is now
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ as the code documentation.
 
 ## Supported Platforms
 
-The kvm-ioctls can be used on x86_64 and aarch64.
+The kvm-ioctls can be used on x86_64, aarch64 and riscv64 (experimental).
 
 ## Running the tests
 
@@ -28,14 +28,16 @@ For the complete list of tests, check our
 [CI pipeline](https://buildkite.com/rust-vmm/kvm-ioctls-ci).
 
 Each individual test runs in a container. To reproduce a test locally, you can
-use the dev-container on both x86 and arm64.
+use the dev-container on x86_64, arm64 and riscv64.
 
 ```bash
+# For running riscv64 tests, replace v47 with v47-riscv. This provides an
+# emulated riscv64 environment on a x86_64 host.
 docker run --device=/dev/kvm \
            -it \
            --security-opt seccomp=unconfined \
            --volume $(pwd)/kvm-ioctls:/kvm-ioctls \
-           rustvmm/dev:v16
+           rustvmm/dev:v47
 cd kvm-ioctls/
 cargo test
 ```
--- a/src/ioctls/device.rs
+++ b/src/ioctls/device.rs
@@ -1,3 +1,5 @@
+// Copyright © 2024 Institute of Software, CAS. All rights reserved.
+//
 // Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
@@ -320,6 +322,91 @@ mod tests {
         unsafe { device_fd.get_device_attr(&mut gic_attr) }.unwrap();
         // The maximum supported number of IRQs should be 128, same as the value
         // when we initialize the GIC.
+        assert_eq!(data, 128);
+    }
+
+    #[test]
+    #[cfg(target_arch = "riscv64")]
+    fn test_create_device() {
+        use crate::ioctls::vm::{create_aia_device, request_aia_init, set_supported_nr_irqs};
+        use kvm_bindings::{
+            kvm_device_attr, kvm_device_type_KVM_DEV_TYPE_FSL_MPIC_20,
+            KVM_DEV_RISCV_AIA_ADDR_APLIC, KVM_DEV_RISCV_AIA_CONFIG_SRCS,
+            KVM_DEV_RISCV_AIA_GRP_ADDR, KVM_DEV_RISCV_AIA_GRP_CONFIG,
+        };
+        use vmm_sys_util::errno::Error;
+
+        let kvm = Kvm::new().unwrap();
+        let vm = kvm.create_vm().unwrap();
+
+        let mut aia_device = kvm_bindings::kvm_create_device {
+            type_: kvm_device_type_KVM_DEV_TYPE_FSL_MPIC_20,
+            fd: 0,
+            flags: KVM_CREATE_DEVICE_TEST,
+        };
+        // This fails on riscv64 as it does not use MPIC (MultiProcessor Interrupt Controller),
+        // it uses the vAIA.
+        vm.create_device(&mut aia_device).unwrap_err();
+
+        let device_fd = create_aia_device(&vm, 0);
+
+        // AIA on riscv64 requires at least one online vCPU prior to setting
+        // device attributes. Otherwise it would fail when trying to set address
+        // of IMSIC.
+        vm.create_vcpu(0).unwrap();
+
+        // Following lines to re-construct device_fd are used to test
+        // DeviceFd::from_raw_fd() and DeviceFd::as_raw_fd().
+        let raw_fd = unsafe { libc::dup(device_fd.as_raw_fd()) };
+        assert!(raw_fd >= 0);
+        let device_fd = unsafe { DeviceFd::from_raw_fd(raw_fd) };
+
+        // Set maximum supported number of IRQs of the vAIA device to 128.
+        set_supported_nr_irqs(&device_fd, 128);
+
+        // Before request vAIA device to initialize, APLIC and IMSIC must be set
+        let aplic_addr: u64 = 0x4000;
+        device_fd
+            .set_device_attr(&kvm_device_attr {
+                group: KVM_DEV_RISCV_AIA_GRP_ADDR,
+                attr: u64::from(KVM_DEV_RISCV_AIA_ADDR_APLIC),
+                addr: &aplic_addr as *const u64 as u64,
+                flags: 0,
+            })
+            .unwrap();
+        let imsic_addr: u64 = 0x8000;
+        device_fd
+            .set_device_attr(&kvm_device_attr {
+                group: KVM_DEV_RISCV_AIA_GRP_ADDR,
+                attr: 1u64,
+                addr: &imsic_addr as *const u64 as u64,
+                flags: 0,
+            })
+            .unwrap();
+
+        // Initialize valid vAIA device.
+        request_aia_init(&device_fd);
+
+        // Test `get_device_attr`. Here we try to extract the maximum supported number of IRQs.
+        // This value should be saved in the address provided to the ioctl.
+        let mut data: u32 = 0;
+
+        let mut aia_attr = kvm_bindings::kvm_device_attr {
+            group: KVM_DEV_RISCV_AIA_GRP_CONFIG,
+            attr: u64::from(KVM_DEV_RISCV_AIA_CONFIG_SRCS),
+            addr: data as u64,
+            ..Default::default()
+        };
+
+        // Without properly providing the address to where the
+        // value will be stored, the ioctl fails with EFAULT.
+        let res = unsafe { device_fd.get_device_attr(&mut aia_attr) };
+        assert_eq!(res, Err(Error::new(libc::EFAULT)));
+
+        aia_attr.addr = &mut data as *mut u32 as u64;
+        unsafe { device_fd.get_device_attr(&mut aia_attr) }.unwrap();
+        // The maximum supported number of IRQs should be 128, same as the value
+        // when we initialize the AIA.
         assert_eq!(data, 128);
     }
 }
--- a/src/ioctls/device.rs
+++ b/src/ioctls/device.rs
@@ -41,36 +41,46 @@ impl DeviceFd {
     ///
     /// # Example
     ///
+    /// Configuring a VFIO device using `set_device_attr`. Note that VFIO
+    /// devices are not yet available on RISC-V The patch for QEMU:
+    /// https://lore.kernel.org/all/20240903201633.93182-1-dbarboza@ventanamicro.com/
+    /// and patch for linux kernel
+    /// https://github.com/ventanamicro/linux/tree/dev-upstream are both not
+    /// upstreamed. Disabling VFIO device test for RISC-V at the time being.
+    ///
     /// ```rust
     /// # extern crate kvm_ioctls;
     /// # extern crate kvm_bindings;
     /// # use kvm_ioctls::Kvm;
-    /// # use kvm_bindings::{
-    ///    kvm_device_type_KVM_DEV_TYPE_VFIO,
-    ///    KVM_DEV_VFIO_GROUP, KVM_DEV_VFIO_GROUP_ADD, KVM_CREATE_DEVICE_TEST
-    /// };
     /// let kvm = Kvm::new().unwrap();
     /// let vm = kvm.create_vm().unwrap();
     ///
-    /// let mut device = kvm_bindings::kvm_create_device {
-    ///     type_: kvm_device_type_KVM_DEV_TYPE_VFIO,
-    ///     fd: 0,
-    ///     flags: KVM_CREATE_DEVICE_TEST,
-    /// };
+    /// #[cfg(not(target_arch = "riscv64"))]
+    /// {
+    ///     # use kvm_bindings::{
+    ///     #     kvm_device_type_KVM_DEV_TYPE_VFIO,
+    ///     #     KVM_DEV_VFIO_GROUP, KVM_DEV_VFIO_GROUP_ADD, KVM_CREATE_DEVICE_TEST
+    ///     # };
+    ///     let mut device = kvm_bindings::kvm_create_device {
+    ///         type_: kvm_device_type_KVM_DEV_TYPE_VFIO,
+    ///         fd: 0,
+    ///         flags: KVM_CREATE_DEVICE_TEST,
+    ///     };
     ///
-    /// let device_fd = vm
-    ///     .create_device(&mut device)
-    ///     .expect("Cannot create KVM device");
+    ///     let device_fd = vm
+    ///         .create_device(&mut device)
+    ///         .expect("Cannot create KVM device");
     ///
-    /// let dist_attr = kvm_bindings::kvm_device_attr {
-    ///     group: KVM_DEV_VFIO_GROUP,
-    ///     attr: u64::from(KVM_DEV_VFIO_GROUP_ADD),
-    ///     addr: 0x0,
-    ///     flags: 0,
-    /// };
+    ///     let dist_attr = kvm_bindings::kvm_device_attr {
+    ///         group: KVM_DEV_VFIO_GROUP,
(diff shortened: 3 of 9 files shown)


TimePrinciple:
How strange, I have no clue why `test_pit2` suddenly fails to pass :(

TimePrinciple:
It's ready now, @roypat @ShadowCurse @andreeaflorescu  PTAL when convenient 😵‍💫

rbradford:
@TimePrinciple Looks like there are some build failures for the riscv64 checks.

TimePrinciple:
> @TimePrinciple Looks like there are some build failures for the riscv64 checks.

Ah yes, I tested locally and all passed. I'll come back to fix that "experimental" features thingy later. And it's wired those were not reported on other architectures 🤔

Sorry I didn't find you in the reviewers list 🥲

TimePrinciple:
> @TimePrinciple Looks like there are some build failures for the riscv64 checks.

@rbradford I found the cause of this problem, the rust toolchain version on riscv64 diverges from the other two. I will fix it there :(

I tried to test my branch locally, everything works fine:

屏幕截图 2024-10-04 224441

TimePrinciple:
> > @TimePrinciple Looks like there are some build failures for the riscv64 checks.
> 
> @rbradford I found the cause of this problem, the rust toolchain version on riscv64 diverges from the other two. I will fix it there :(

屏幕截图 2024-10-04 224441

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Added
 
+- [[#275](https://github.com/rust-vmm/kvm-ioctls/pull/275)]: Introduce `riscv64` ioctls.
```
dm0-, line 7: Should the README's platform list also be updated to reflect this?

TimePrinciple: Thanks for reminding! I'll put it into `READMD` too 😊

TimePrinciple:
CI is ready, PTAL when convenient @roypat @rbradford @ShadowCurse

src/ioctls/vcpu.rs
```diff
@@ -222,7 +222,7 @@ impl VcpuFd {
     /// let vcpu = vm.create_vcpu(0).unwrap();
     /// let regs = vcpu.get_regs().unwrap();
     /// ```
-    #[cfg(not(any(target_arch = "arm", target_arch = "aarch64")))]
+    #[cfg(not(any(target_arch = "arm", target_arch = "aarch64", target_arch = "riscv64")))]
     pub fn get_regs(&self) -> Result<kvm_regs> {
```
rbradford, line 228: At this point I wonder if it should just be `#[cfg(target_arch = "x86_64")]`

src/ioctls/vcpu.rs
```diff
@@ -1212,12 +1212,23 @@ impl VcpuFd {
     /// let vm = kvm.create_vm().unwrap();
     /// let vcpu = vm.create_vcpu(0).unwrap();
     ///
-    /// // KVM_GET_REG_LIST demands that the vcpus be initalized.
-    /// let mut kvi: kvm_bindings::kvm_vcpu_init = kvm_bindings::kvm_vcpu_init::default();
-    /// vm.get_preferred_target(&mut kvi).unwrap();
-    /// vcpu.vcpu_init(&kvi).expect("Cannot initialize vcpu");
+    /// // KVM_GET_REG_LIST on Aarch64 demands that the vcpus be initalized.
```
rbradford: nit: missing the 3rd 'i' in initialized

src/lib.rs
```diff
@@ -21,6 +23,7 @@
 //!
 //! - x86_64
 //! - arm64 (experimental)
```
rbradford, line 25: I think elsewhere you mark arm64 as non-experimental.

rbradford (commented):
Cool - thanks!

src/lib.rs
TimePrinciple: Centainly, that would be in the `prepare-release` PR after this is merged :)

src/ioctls/vcpu.rs
TimePrinciple: Nice catch :)

src/ioctls/vcpu.rs
TimePrinciple: Good point, I'll check if there is anything similar to this

TimePrinciple: It just occured to me, `ppc` and etc. also have this ioctl, wouldn't narrowing to `x86_64` break the builds on other architectures 🧐

rbradford: I think it's best to confine the checks only to the architectures we support with this crate.

rbradford: FWIW - your solution is fine as is.

rbradford (dismissed):

src/ioctls/vcpu.rs
TimePrinciple: > I think it's best to confine the checks only to the architectures we support with this crate.

I will purpose that in another PR :) to discuss whether we should drop `target_arch` predicates to architectures other than `x86`, `arm` and `riscv`

README.md
```diff
@@ -28,14 +28,15 @@ For the complete list of tests, check our
 [CI pipeline](https://buildkite.com/rust-vmm/kvm-ioctls-ci).
 
 Each individual test runs in a container. To reproduce a test locally, you can
-use the dev-container on both x86 and arm64.
+use the dev-container on both x86, arm64 and riscv64.
```
roypat:
```suggestion
use the dev-container on x86_64, arm64 and riscv64.
```

README.md
```diff
@@ -28,14 +28,15 @@ For the complete list of tests, check our
...
 
 ```bash
+# As of riscv64, the tag is v47-riscv
```
roypat: How about "for running riscv64 tests, replace `v47` with `v47-riscv`. This provides an emulated riscv64 environment on a x86_64 host."

src/ioctls/device.rs
```diff
@@ -45,32 +47,71 @@ impl DeviceFd {
...
-    /// };
-    ///
-    /// if (device_fd.has_device_attr(&dist_attr).is_ok()) {
-    ///     device_fd.set_device_attr(&dist_attr).unwrap();
+    /// // The VFIO device is not yet available on RISC-V. The patch for QEMU:
+    /// // https://lore.kernel.org/all/20240903201633.93182-1-dbarboza@ventanamicro.com/
+    /// // and patch for linux kernel
+    /// // https://github.com/ventanamicro/linux/tree/dev-upstream are both not
+    /// // upstreamed. Disabling VFIO device test fro RISC-V at the time being.
+    /// #[cfg(not(target_arch = "riscv64"))]
```
roypat: Mh, is there another test we can instead do on riscv for this function? Otherwise, if its completely impossible to use this function on riscv yet, this should be part of the normal doccomment for this function, and not just the example. We should also use the `#` prefixes here so that the cfg and comment don't show up on docs.rs (and then have a little note at the top of the example that briefly says what it does, e.g. "seting up a vfio device on arm64")

src/ioctls/device.rs
```diff
@@ -117,7 +158,7 @@ impl DeviceFd {
     /// let vm = kvm.create_vm().unwrap();
     ///
     /// // As on x86_64, `get_device_attr` is not necessarily needed. Therefore here
-    /// // the code example is only for AArch64.
+    /// // the code example is only for AArch64 and RISC-V 64-bit.
     /// #[cfg(any(target_arch = "aarch64"))]
```
roypat: Mh, I admittedly think that for doc comments, having the doc comments only for a single architecture is fine. I'd move the riscv64 example you add below to the unittests instead (also, I now realize that the `cfg`s are _already_ shown on doc.rs. ugh. disregard my previous comment about that, maybe I'll post a separate fixup for that later

src/ioctls/vcpu.rs
```diff
@@ -1210,16 +1214,27 @@ impl VcpuFd {
...
+    /// #[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
+    /// {
+    ///     let mut kvi: kvm_bindings::kvm_vcpu_init = kvm_bindings::kvm_vcpu_init::default();
```
roypat, line 1220: nit:

```suggestion
    ///     let mut kvi = kvm_vcpu_init::default();
```

and add an import for `kvm_bindings::kvm_vcpu_init`

src/ioctls/vcpu.rs
roypat: Still missing :D

src/ioctls/vcpu.rs
```diff
@@ -1210,16 +1214,27 @@ impl VcpuFd {
...
+    ///     vcpu.vcpu_init(&kvi).expect("Cannot initialize vcpu");
+    /// }
+    ///
+    /// let nr_regs = if cfg!(any(target_arch = "arm", target_arch = "aarch64")) {
+    ///     500
+    /// } else if cfg!(target_arch = "riscv64") {
+    ///     200
+    /// } else {
+    ///     unreachable!();
+    /// };
```
roypat: Here I also think it's fine to have the code example only for ARM, with the riscv case moved to unittests (or vice-versa). The idea of the doc example is more to give an idea of how to use the API, and since the API itself is identical on multiple architectures, I don't really think we need to have the same code twice, just with different constants

src/ioctls/vcpu.rs
```diff
@@ -2559,10 +2668,22 @@ mod tests {
...
+            target_arch = "x86_64",
+            target_arch = "arm",
+            target_arch = "aarch64"
+        ))]
         assert_eq!(
             faulty_vcpu_fd.get_vcpu_events().unwrap_err().errno(),
             badf_errno
         );
+        #[cfg(any(
+            target_arch = "x86",
```
roypat, line 2674: No need for 32bit targets in these cfgs (actually, KVM on ARM32 hosts went away in 5.7 I think: https://www.phoronix.com/news/Linux-5.7-Kill-32-bit-ARM-KVM)

roypat: Mh, I see we still have those all over the place... might also be worth a separate cleanup to just get rid of them all. I made a note for myself about this, feel free to ignore this comment for now, as its just a "fitting in with existing code" kinda thing.

src/lib.rs
```diff
@@ -93,6 +97,15 @@
 //!             0x14, /* b <this address>; shouldn't get here, but if so loop forever */
 //!         ];
 //!     }
+//!     #[cfg(target_arch = "riscv64")]
```
roypat, line 100: Ah, I see now. I guess here in this top level documentation, having something for each supported arch is fine :)

roypat (commented):
Sorry it took me a while to get to this :(

I'm trusting Rob on his assessment of the riscv bits, my comments are largely about code organization

README.md
TimePrinciple: Noted :)

README.md
TimePrinciple: Ah, yes, I'm not a native speaker, glad to have you to take care of this :)

src/ioctls/device.rs
TimePrinciple: Good point, I'll fix it

src/ioctls/vcpu.rs
TimePrinciple: Yes, I was about to raise a PR to drop unsupported architectures (or not tested) not just limited to `x86` (`s390` and etc. as well) :)

roypat: I think the s390x and ppc ones are fine, because there's folks that use them and also provide fixes if we break stuff, so it's sort of "unofficially maintained", to some capacity. The 32bit ones can definitely go though imo (ARM32 because it doesnt exist anymore, and x86 because we're already very explicit about not supporting 32 bit in other rust-vmm crates)

src/ioctls/vcpu.rs
TimePrinciple: Sorry, somehow missed, thanks for pointing out! :)

src/ioctls/device.rs
TimePrinciple: I moved the comment to normal document, and left cfg visible to tell the difference between riscv and no-riscv platform :)

src/ioctls/device.rs
TimePrinciple: Oh, I just recollected, I'm writing these doc examples because without completing them, the doc test simply wouldn't pass on `riscv64` architecture :(

src/ioctls/device.rs
TimePrinciple: > Mh, is there another test we can instead do on riscv for this function? Otherwise, if its completely impossible to use this function on riscv yet, this should be part of the normal doccomment for this function, and not just the example. We should also use the `#` prefixes here so that the cfg and comment don't show up on docs.rs (and then have a little note at the top of the example that briefly says what it does, e.g. "seting up a vfio device on arm64")

IMHO we are illustrating `set_device_attr` here, it could be used on various devices. The VFIO device previously chose for `x86_64` and `arm64` is not available yet, using an AIA device to illustrate also proves that `set_device_attr` is functioning :)

src/ioctls/device.rs
roypat: Huh, can you give an example? I don't immediately see why they wouldn't pass

src/ioctls/device.rs
roypat: Well, then we could also add separate entries for all sorts of devices that are supported on each of the architectures. For such a low-level ioctl wrapper, I don't really think we should go down that route. E.g. using the functions in this crate, you're still very much expected to consult the KVM docs, and we probably shouldn't try to duplicate those here. The examples for these imo should indicate what the general pattern of doing the ioctl should be (e.g. what structures to set up for the call, how to handle the return value), but I don't really see the value of doing that with multiple permutations of constants. And to prove that `set_device_attr` is functioning on riscv, we can add a unit test (or if you think the AIA device is a better example than the VFIO device, then we keep the AIA example here, and move the VFIO example to the unittests).

That being said, maybe it's worth to add some slightly higher-level wrappers, say, `set_aia_device_attr` or something that specifically deals with AIA devices, or `set_vfio_device_attr` that specifically deals with VFIO devices? Then those should have examples in their doc comments detailing the specifics of their device.

src/ioctls/device.rs
TimePrinciple: Yes, I'll illustrate per commit:

- 6a2d5dc: `type_` field is architecture specific, when running on riscv the field would be missing then failing
- a75b454: fails on riscv because 500 are too big
- 8fc3cda: fails because VFIO is not yet available on riscv
- ea0e5a4: this one is fine, because it does nothing on riscv
- 3649c68: `set_gsi_routing` requires AIA created in advance on riscv
- 2246f1c: `asm_code` is architecture specific

roypat: - https://github.com/rust-vmm/kvm-ioctls/commit/6a2d5dcb80b273a52a040d4918a55e69b15177dd: this one is fine, its just a single line
- https://github.com/rust-vmm/kvm-ioctls/commit/a75b454cba25cebf487b672e082601f0cd130841: Have the example only be the ARM64 variant, and do the riscv check in unittests
- https://github.com/rust-vmm/kvm-ioctls/commit/8fc3cdad44d2b74f7eb5ae21fe4a91ee6551b8d5: cfg it to not run on riscv, do something riscv related in unittests. If the QEMU patches get merged, we can drop the cfg
- https://github.com/rust-vmm/kvm-ioctls/commit/3649c684e2552872b332856630f1e5d490a1a05f: This one is again fine
- https://github.com/rust-vmm/kvm-ioctls/commit/2246f1c7965538823093fc1a328d907e45cd1ad7: also fine

I think I'm drawing the line at: single line additions because setup for ioctls is different on different architectures (fine) vs duplication of large parts of/the entire example only to pass different constants to the ioctl (let's not do this).

TimePrinciple: Noted, I'll stick to it 👌

src/ioctls/device.rs
TimePrinciple: > That being said, maybe it's worth to add some slightly higher-level wrappers, say, `set_aia_device_attr` or something that specifically deals with AIA devices, or `set_vfio_device_attr` that specifically deals with VFIO devices? Then those should have examples in their doc comments detailing the specifics of their device.

Very good point, I'll work on it :)

src/ioctls/device.rs
```diff
@@ -148,6 +159,33 @@ impl DeviceFd {
...
+    ///
+    ///     let mut data: u32 = 0;
+    ///     let mut aia_attr = kvm_bindings::kvm_device_attr::default();
+    ///     aia_attr.group = KVM_DEV_RISCV_AIA_GRP_CONFIG;
+    ///     aia_attr.attr = u64::from(KVM_DEV_RISCV_AIA_CONFIG_IDS);
+    ///     aia_attr.addr = &mut data as *mut u32 as u64;
+    ///
+    ///     // SAFETY: gic_attr.addr is safe to write to.
+    ///     unsafe { device_fd.get_device_attr(&mut aia_attr) }.unwrap();
+    /// }
```
roypat: This is still a duplicated example :(

src/ioctls/device.rs
```diff
@@ -34,6 +36,12 @@ impl DeviceFd {
 
     /// Sets a specified piece of device configuration and/or state.
     ///
+    /// The VFIO device is not yet available on RISC-V. The patch for QEMU:
+    /// https://lore.kernel.org/all/20240903201633.93182-1-dbarboza@ventanamicro.com/
+    /// and patch for linux kernel
+    /// https://github.com/ventanamicro/linux/tree/dev-upstream are both not
+    /// upstreamed. Disabling VFIO device test for RISC-V at the time being.
+    ///
```
roypat: Can you move this to line 51? I was thinking that the `Example` section should be 

```
# Example

Configuring a VFIO device using `set_device_attr`. Note that VFIO devices are not yet available on RISC-V [...]
```

TimePrinciple: Noted

src/ioctls/device.rs
TimePrinciple: 😂 Sorry I missed this, fixing right now

TimePrinciple: 😂 Sorry I missed this, fixing right now

TimePrinciple: I will drop this commit :)

TimePrinciple:
@rbradford @roypat please take another look :)

src/ioctls/device.rs
```diff
@@ -312,4 +324,89 @@ mod tests {
...
+            flags: KVM_CREATE_DEVICE_TEST,
+        };
+        // This fails on riscv64 as it does not use MPIC (MultiProcessor Interrupt Controller),
+        // it uses the vAIA.
+        vm.create_device(&mut aia_device).unwrap_err();
+
+        let device_fd = create_aia_device(&vm, 0);
+
+        // AIA on riscv64 requires at least one online vCPU prior to setting
+        // device attributes. Otherwise it would fail when trying ot set address
```
rbradford: nit: typo "ot" -> "to"

src/ioctls/vm.rs
```diff
@@ -2305,6 +2462,47 @@ mod tests {
         vm_fd.set_irq_line(0x02_00_0010, true).unwrap();
     }
 
+    #[test]
+    #[cfg(target_arch = "riscv64")]
+    #[allow(clippy::unusual_byte_groupings)]
```
rbradford: nit: I think this was copied from another test - I don't see any unusual byte grouping in this test.

rbradford (commented):
Looking really good!

src/ioctls/vm.rs
TimePrinciple: Ah yes, I reference the test on aarch64, I'll remove this :)

src/ioctls/device.rs
TimePrinciple: Noted, thanks 😊

rbradford (approved):

roypat (approved):

PR #281, opened 2024-09-26
aarch64: Reorganize `test_create_device`

TimePrinciple:
### Summary of the PR

Add `has_device_attr` check in both `set_supported_nr_irqs` and `request_gic_init`, effectively enabling reuse of `request_gic_init` in `test_create_device`.

I was going through these tests while introducing RISC-V ioctls. Some minor improvements :)

### Requirements

requirements:

  test.

### DIFF
--- a/src/ioctls/device.rs
+++ b/src/ioctls/device.rs
@@ -196,10 +196,7 @@ mod tests {
         KVM_DEV_VFIO_GROUP, KVM_DEV_VFIO_GROUP_ADD,
     };
     #[cfg(target_arch = "aarch64")]
-    use kvm_bindings::{
-        KVM_DEV_ARM_VGIC_CTRL_INIT, KVM_DEV_ARM_VGIC_GRP_CTRL, KVM_DEV_VFIO_GROUP,
-        KVM_DEV_VFIO_GROUP_ADD,
-    };
+    use kvm_bindings::{KVM_DEV_VFIO_GROUP, KVM_DEV_VFIO_GROUP_ADD};
 
     use kvm_bindings::KVM_CREATE_DEVICE_TEST;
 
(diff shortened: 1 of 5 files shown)


src/ioctls/device.rs
```diff
@@ -292,15 +289,7 @@ mod tests {
         set_supported_nr_irqs(&device_fd, 128);
 
         // Following attribute works with VGIC, they should be accepted.
```
roypat: Let's also update this comment (something like "initialize valid vgic"?)

src/ioctls/vm.rs
```diff
@@ -1942,6 +1943,7 @@ pub(crate) fn request_gic_init(vgic: &DeviceFd) {
         addr: 0,
         flags: 0,
     };
+    assert!(vgic.has_device_attr(&vgic_attr).is_ok());
```
roypat: Kinda unrelated to your PR, but could we do
```suggestion
    vgic.has_device_attr(&vgic_attr).unwrap();
```
instead? `assert((...).is_ok())` gives very unhelpful error messages, because it doesn't print out the `Err` variant (should an error happen).

TimePrinciple: A very good point, should I replace these `assert!`s with `unwrap()` within the lines this PR touched, or do it globally 🤔

roypat: If you want to, I'd really appreciate a global change, although I don't insist on it, because it can be quite tedious (for reference, the PR when I did it on Firecracker: https://github.com/firecracker-microvm/firecracker/pull/4321).

So just changing it for the function you touch here is fine!

src/ioctls/device.rs
TimePrinciple: Updated :)

.cargo/config.toml
```diff
@@ -1,3 +1,8 @@
+[build]
+rustflags = [
+  "-Wclippy::assertions_on_result_states",
+]
```
roypat: This should be a `#![warn(clippy::assertions_on_result_states)]` in `lib.rs` :)

We did it like this in Firecracker for unrelated reasons that don't apply to rust-vmm

TimePrinciple: > This should be a `#![warn(clippy::assertions_on_result_states)]` in `lib.rs` :)
> 
> We did it like this in Firecracker for unrelated reasons that don't apply to rust-vmm

Noted

roypat (approved):

rbradford (approved):
Great improvements. I especially like the assertion cleanups.

TimePrinciple:
> Great improvements. I especially like the assertion cleanups.

If you like, I will do it in cloud-hypervisor :)

PR #284, opened 2024-10-04
Added kvm guest memfd related capabilities

KeyboardNerd:
### Summary of the PR

The capabilities are required to properly setup a guest_memfd to provide better host and guest memory isolation. The memory attributes capability returns an integer with each bits representing different configs. For example, the return integer & KVM_MEMORY_ATTRIBUTE_PRIVATE > 0 means the vm is capable of setting memory pages to private.

### Requirements

requirements:

  test.


src/ioctls/vm.rs
```diff
@@ -1390,7 +1390,7 @@ impl VmFd {
     /// Wrapper over `KVM_CHECK_EXTENSION`.
     ///
     /// Returns 0 if the capability is not available and a positive integer otherwise.
-    fn check_extension_int(&self, c: Cap) -> i32 {
+    pub fn check_extension_int(&self, c: Cap) -> i32 {
```
roypat: Hey @KeyboardNerd,
thanks for you PR! what is the motivation for making this function public? Do we need the exactly int return value for any of these capabilities? :o If not, let's keep this function private, as `VmFd::check_extension` provides equivalent functionality

KeyboardNerd: cap check with [KVM_CAP_MEMORY_ATTRIBUTES](https://github.com/torvalds/linux/blob/87d6aab2389e5ce0197d8257d5f8ee965a67c4cd/include/uapi/linux/kvm.h#L930C9-L930C34) returns an integer, which may have bits representing different attributes. One of them is [KVM_MEMORY_ATTRIBUTE_PRIVATE=8](https://github.com/torvalds/linux/blob/master/include/uapi/linux/kvm.h#L1557C9-L1557C37) I'm intended to use ( [other attributes](https://lore.kernel.org/lkml/20221202061347.1070246-3-chao.p.peng@linux.intel.com/) ). So it's necessary to get the extension integer value to check specific bit in the capability.

For another reference, qemu implemented the memory check with & on the KVM_CAP_MEMORY_ATTRIBUTES return value.
[source](https://github.com/qemu/qemu/blob/2af37e791906cfda42cb9604a16d218e56994bb1/accel/kvm/kvm-all.c#L2552)

roypat: Ahhh, thanks, makes sense then! 

I do wonder if we should include a slightly less thin wrapper though. E.g in addition to making this function public, can we add a `fn supported_memory_attributes(&self) -> KvmMemoryAttributes` with some `enum KvmMemoryAttributes(i32) { Private = KVM_MEMORY_ATTRIBUTE_PRIVATE=8 }` or something? We have those sort of abstractions for `KVM_CHECK_EXTENSION` on the system fd (e.g. https://github.com/rust-vmm/kvm-ioctls/blob/e75abe5f7b0293cc2318012a941d5884cd6abd42/src/ioctls/system.rs#L158-L177).

Also, while we're talking about making this consistent with the API offered on the system fd, maybe we can also add a [`check_extension_raw`](https://github.com/rust-vmm/kvm-ioctls/blob/main/src/ioctls/system.rs#L198)? That way, people can use extension checks before they're supported by kvm-ioctl's `Cap` enum.

PR #288, opened 2024-10-09
Add kvm guest memfd related capabilities

KeyboardNerd:
### Summary of the PR

The capabilities are required to properly setup a guest_memfd to provide better host and guest memory isolation. The memory attributes capability returns an integer with each bits representing different configs. For example, the return integer & KVM_MEMORY_ATTRIBUTE_PRIVATE > 0 means the vm is capable of setting memory pages to private.

### Requirements

requirements:

  test.

### DIFF
--- a/kvm-ioctls/CHANGELOG.md
+++ b/kvm-ioctls/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - [[#288](https://github.com/rust-vmm/kvm-ioctls/pull/288)]: Introduce `Cap::GuestMemfd`, `Cap::MemoryAttributes` and 
    `Cap::UserMemory2` capabilities enum variants for use with `VmFd::check_extension`.
+- [[#288](https://github.com/rust-vmm/kvm-ioctls/pull/288)]: Introduce `VmFd::check_extension_raw` and `VmFd::check_extension_int` to allow `KVM_CHECK_EXTENSION` to return integer.
 
 ### Fixed
 
--- a/kvm-ioctls/src/ioctls/vm.rs
+++ b/kvm-ioctls/src/ioctls/vm.rs
@@ -1402,10 +1402,50 @@ impl VmFd {
     /// Wrapper over `KVM_CHECK_EXTENSION`.
     ///
     /// Returns 0 if the capability is not available and a positive integer otherwise.
-    fn check_extension_int(&self, c: Cap) -> i32 {
-        // SAFETY: Safe because we know that our file is a VM fd and that the extension is one of
-        // the ones defined by kernel.
-        unsafe { ioctl_with_val(self, KVM_CHECK_EXTENSION(), c as c_ulong) }
+    /// See the documentation for `KVM_CHECK_EXTENSION`.
+    ///
+    /// # Arguments
+    ///
+    /// * `c` - KVM capability to check.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use kvm_ioctls::Kvm;
+    /// use kvm_ioctls::Cap;
+    ///
+    /// let kvm = Kvm::new().unwrap();
+    /// let vm = kvm.create_vm().unwrap();
+    /// assert!(vm.check_extension_int(Cap::MaxVcpus) > 0);
+    /// ```
+    pub fn check_extension_int(&self, c: Cap) -> i32 {
+        self.check_extension_raw(c as c_ulong)
+    }
+
+    /// Wrapper over `KVM_CHECK_EXTENSION`.
+    ///
+    /// Returns 0 if the capability is not available and a positive integer otherwise.
+    /// See the documentation for `KVM_CHECK_EXTENSION`.
+    ///
+    /// # Arguments
+    ///
+    /// * `c` - KVM capability to check in a form of a raw integer.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use kvm_ioctls::Kvm;
+    /// # use std::os::raw::c_ulong;
+    /// use kvm_ioctls::Cap;
+    ///
+    /// let kvm = Kvm::new().unwrap();
+    /// let vm = kvm.create_vm().unwrap();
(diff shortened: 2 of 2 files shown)


KeyboardNerd:
It's a continued discussion from 
https://github.com/rust-vmm/kvm-ioctls/pull/284
I accidentally closed that one :(

src/ioctls/vm.rs
```diff
@@ -2513,6 +2554,20 @@ mod tests {
         assert!(vm.check_extension(Cap::MpState));
     }
 
+    #[test]
+    fn test_check_extension_int() {
+        let kvm = Kvm::new().unwrap();
+        let vm = kvm.create_vm().unwrap();
+        assert!(vm.check_extension_int(Cap::MpState) > 0);
+    }
```
TimePrinciple: Since `test_check_extension_raw` exists, in my view `test_check_extension_int` is not necessary, it's just a wrapper on `check_extension_raw`. This test is just duplicating existing code :)

src/ioctls/vm.rs
```diff
@@ -2513,6 +2554,20 @@ mod tests {
...
+
+    #[test]
+    fn test_check_extension_raw() {
+        let kvm = Kvm::new().unwrap();
+        let vm = kvm.create_vm().unwrap();
+        assert!(vm.check_extension_raw(Cap::MpState as c_ulong) > 0);
```
TimePrinciple: IMHO this is literally a `check_extension` test which gets expanded, if we intend to test `check_extension_raw` positively, we can do something like `assert_eq!(vm.check_extension_raw(Cap::MaxVcpus), 1024);` here to state that we want an `int` which its value matters.

Or else we just do negative test here :)

TimePrinciple (commented):
Thanks for this work! While I think you could still reopen a PR if you accidentally closed one, if you still have that branch :) Thus we could avoid throwing out PRs

And could you squash 741fb48, 2523d68, 7283cc6 and 3778efe altogether, otherwise we would break bistability along the way 🙂

src/ioctls/vm.rs
TimePrinciple: 1024 here is for RISC-V, as of the other two archs, we will need to set its value accordingly

src/ioctls/vm.rs
```diff
@@ -1390,10 +1390,51 @@ impl VmFd {
...
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use kvm_ioctls::Kvm;
+    /// # use std::os::raw::c_ulong;
+    /// use kvm_ioctls::Cap;
+    ///
+    /// let kvm = Kvm::new().unwrap();
+    /// assert!(kvm.check_extension_raw(Cap::MaxVcpus as c_ulong) > 0);
```
roypat: This is the wrong function to call here, as it does the system-wide extension check, instead of the VM-level extension check. If you fix this, you can also remove the unittest you added, because they're the same as these doctests :)

src/ioctls/vm.rs
```diff
@@ -1390,10 +1390,51 @@ impl VmFd {
...
+    /// # Arguments
+    ///
+    /// * `c` - KVM capability to check.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// extern crate kvm_bindings;
+    ///
+    /// # use kvm_bindings::{KVM_MEMORY_ATTRIBUTE_PRIVATE};
```
roypat: No need for this import, since the final version of the code is checking the MaxVcpus capability

roypat:
Could you also squash all the commits into one?

roypat:
Hey @KeyboardNerd, are you still interested in landing this?

KeyboardNerd:
> Hey @KeyboardNerd, are you still interested in landing this?

Yup, now I have cycles to finish this up. The failure comes from "Title:     Undefined behaviour in `kvm_ioctls::ioctls::vm::VmFd::create_device`" same as the one I see in the lastest PR: https://github.com/rust-vmm/kvm/pull/301. Not sure about the reason.

kvm-ioctls/src/ioctls/vm.rs
```diff
@@ -1402,10 +1402,51 @@ impl VmFd {
...
+    /// See the documentation for `KVM_CHECK_EXTENSION`.
+    ///
+    /// # Arguments
+    ///
+    /// * `c` - KVM capability to check.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use kvm_bindings::{KVM_MEMORY_ATTRIBUTE_PRIVATE};
```
roypat: nit: no need for this import either :)

kvm-ioctls/CHANGELOG.md
```diff
@@ -29,6 +29,8 @@
 - [[#273](https://github.com/rust-vmm/kvm-ioctls/pull/273)]: `DeviceFd::get_device_attr` is now
   marked as unsafe.
 - [[#277](https://github.com/rust-vmm/kvm-ioctls/pull/277)]: Updated kvm-bindings to 0.9.1.
+- [[#288](https://github.com/rust-vmm/kvm-ioctls/pull/288)]: Add kvm guest memfd related
```
roypat: The entry for this PR should be in the [unreleased] section (right now its in the 0.19.0 section, but that version was already released). It also needs to be updated to only state that `VmFd::check_extension_int` and `VmFd::check_extension_raw` were added (since the guest_memfd stuff already made it into 0.18.0)

roypat (commented):
Thanks for updating this! Let's update the commit message to reflect what's actually happening here (since the code itself isn't directly mentioning any guest_memfd stuff anymore, even though the checking of attribute support remains the motivation for these functions).

~~The CI failure I think can be fixed by just changing the version in Cargo.toml to 0.19.1. It's weird that cargo audit picks this up, but oh well lol~~ (fixed in main)

roypat (approved):

TimePrinciple (approved):
Thanks for this work, could @rbradford take a look on this if convenient 🙂

ShadowCurse (approved):

PR #292, opened 2024-10-29
monorepo: Merge `kvm-bindings` into `kvm-ioctls`

### LINKED ISSUE #291: Merging kvm-ioctls and kvm-bindings GitHub repositories
Hey all,

at a recent community sync meeting, we ended up talking about https://github.com/rust-vmm/community/issues/180 again, and how merging kvm-bindings and kvm-ioctls into a single repository (_not_ crate, e.g. the repository would hold a cargo workspace with two members, kvm-ioctls and kvm-bindings) could be one of the simplifications we can make to our repository structure. 

I found an old issue about doing exactly this that seemed to have turned stale (https://github.com/rust-vmm/community/issues/98), so this seems to have been the plan already at some point in the past at least. Are folks still interested in doing this?

cc @rbradford tagging you explicitly since it's mostly been us commiting here recently :D

TimePrinciple:
### Summary of the PR

Fixes #291 

### Requirements

requirements:

  test.


roypat:
@rbradford are you okay with this approach as well? e.g. merging kvm-bindings into the kvm-ioctls repository, and renaming this repo to just "kvm", instead of moving both ioctls and bindings to a new repository (the idea being that this way, we preserve stars and other properties of this repository)

TimePrinciple:
> @rbradford are you okay with this approach as well? e.g. merging kvm-bindings into the kvm-ioctls repository, and renaming this repo to just "kvm", instead of moving both ioctls and bindings to a new repository (the idea being that this way, we preserve stars and other properties of this repository)

I remember @rbradford  is on vacation and will be back on 4th Nov 😂 (if the date he left on his slack is correct)

rbradford:
> @rbradford are you okay with this approach as well? e.g. merging kvm-bindings into the kvm-ioctls repository, and renaming this repo to just "kvm", instead of moving both ioctls and bindings to a new repository (the idea being that this way, we preserve stars and other properties of this repository)

Sounds good to me!

.buildkite/custom-tests.json
```diff
@@ -0,0 +1,38 @@
...
+        {
+            "test_name": "build-serde-musl",
+            "command": "cargo build --release --features=serde --target {target_platform}-unknown-linux-musl",
+            "platform": [
+                "x86_64",
+                "aarch64"
+            ]
+        }
+    ]
+}
```
roypat, line 38: These don't yet show up in buildkite because the pipeline steps don't include `./rust-vmm-ci/.buildkite/autogenerate_pipeline.py -t .buildkite/custom-tests.json | buildkite-agent pipeline upload`. Let me temporarily add it, retrigger the pipeline here. If it works, I'll remove it again and only permanently add it once this is merged (as otherwise we'll break CI on all other PRs)

roypat: Seems to all work: https://buildkite.com/rust-vmm/kvm-ioctls-ci/builds/1250#_ :tada:

roypat: Commented out the steps in the pipeline definition again until after we merge this

roypat (dismissed):
I've set the DCO check to "pass" because it's failing on ages old commits from kvm-bindings that look fine to me

andreeaflorescu:
Typo on commit 1e72627c98a0197b17c2fc5a6311d0ab0002d2a6: "Prepare for `kvm-bindings` to be merged, prevent conflicks from happening.".

Here is a suggestion for rephrasing your commit message to make it more explicit:

> monorepo: Move kvm-ioctls related files into kvm-ioctls folder
>
> This change  is needed to be able to merge the `kvm-bindings` crate into this repository
> and prevent conflicts from happening.

.cargo/config
```diff
@@ -1,5 +0,0 @@
-# This workaround is needed because the linker is unable to find __addtf3,
```
andreeaflorescu: Why is this not needed anymore?

kvm-ioctls/coverage_config_aarch64.json
```diff
@@ -1,5 +0,0 @@
-{
-  "coverage_score": 77.1,
```
andreeaflorescu: [optional] I would prefer to have an independent coverage test for the kvm-ioctls. For this we would need to add a custom test that runs pytest from the directory I think. If it's too complicated, we can also add a separate issue to handle it.

Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[workspace]
+resolver = "2"
+members = [
+    "kvm-bindings",
+    "kvm-ioctls",
+]
+
+[workspace.dependencies]
```
andreeaflorescu: Why does this need to be defined as a workspace dependency?

Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[workspace]
```
andreeaflorescu, line 1: [nit] I don't quite understand what you mean with the commit message. Could you update it?

I guess what you want to say is: setup the repository as a Rust workspace?
And that the members of the workspace are kvm-bindings and kvm-ioctls?

You also have a typo in "contralize"

README.md
```diff
@@ -0,0 +1,91 @@
+[![Build Status](https://badge.buildkite.com/9e0e6c88972a3248a0908506d6946624da84e4e18c0870c4d0.svg)](https://buildkite.com/rust-vmm/kvm-ci)
+
+# `kvm` - Rust Bindings and Wrappers Linux's Kernel Virtual Machine APIs
+
+This repository contains the `kvm-bindings` and `kvm-ioctls` crates.
+
+## kvm-ioctls
```
andreeaflorescu: [blocking for the merge] If you do this, then the crates.io readme will be empty (unless you update the Cargo.toml of each crate to specify the path to the new readme).

The right approach is to have each crate with their own README.md file, and on top of that, also have a top level readme that links to the individual readmes of the individual crates.

kvm-bindings can still be used without kvm-ioctls, so it doesn't make sense to have the readmes together.

andreeaflorescu (changes requested):

Cargo.toml
TimePrinciple: IMHO keeping common dependencies here could help us better manage the monorepo, and send the message that we want these dependencies stay in sync :)

Cargo.toml
TimePrinciple: Exactly, I will update it

README.md
TimePrinciple: This was my idea of merging two READMEs together, I was thinking in a "workspace" way and didn't consider we need to publish them. Sorry for the inconvenience @roypat :(

kvm-ioctls/coverage_config_aarch64.json
TimePrinciple: Noted, I will work it out :)

.cargo/config
TimePrinciple: I will keep it then :)

kvm-ioctls/coverage_config_aarch64.json
TimePrinciple: I just recollected if we were to do coverage test per crate, we'd better update our CI to utilize codecov. I will try to do it first and come back.

How about we leave this as it is for now :)

andreeaflorescu: The current CI supports running individual coverage tests, no need to switch to codecov for this reason. But maybe codecov has some other features that make it a better choice. In any case, we can just open an issue and address this later. No need to spend time on it now.

Cargo.toml
andreeaflorescu: Looking at the documentation of workspace, it seems like you can't define the dependencies as optional: https://doc.rust-lang.org/nightly/cargo/reference/workspaces.html#the-dependencies-table. But we do need it to be optional for kvm-bindings. How does that work?


In any case, I would propose this change to be done in a separate PR because it is not strictly required for this merge, and we should discuss it separately.

README.md
andreeaflorescu: I would propose to leave the readmes as they are in this PR, and just add a top-level readme. Then you can open a separate PR in which you update the kvm-bindings readme to fix the problems you've identified.

Cargo.toml
TimePrinciple: > Looking at the documentation of workspace, it seems like you can't define the dependencies as optional: https://doc.rust-lang.org/nightly/cargo/reference/workspaces.html#the-dependencies-table. But we do need it to be optional for kvm-bindings. How does that work?
> 
> In any case, I would propose this change to be done in a separate PR because it is not strictly required for this merge, and we should discuss it separately.

Good point, I will drop it then

andreeaflorescu:
You're missing the top level README.md file. Are you planning to add it in a future PR?

TimePrinciple:
> You're missing the top level README.md file. Are you planning to add it in a future PR?

Yes, @roypat picked that up and I think it could be better to let him continue his work :)

andreeaflorescu (approved):

README.md
roypat: Argh, sorry, this was my bad. Forgot about the crates.io thing. Will post a follow up!

roypat (approved):

roypat:
Similarly to https://github.com/rust-vmm/vhost/pull/114#issuecomment-1280773845, this PR cannot be merged with a rebase. I've also temporarily allowed merge commits to merge this

roypat:
I've
- updated the buildkite steps to include the custom-tests
- Updated the repository name
- Updated the buildkite pipeline name
- Updated the buildkite pipeline to fetch from the renamed repository

roypat:
Also updated branch protection rules to point towards the new pipeline name

PR #298, opened 2024-12-09
fix: use correct ioctl wrapper for `create_device`

ShadowCurse:
### Summary of the PR

Use `ioctl_with_mut_ref` instead of `ioctl_with_ref` for `create_device` as it needs to write to the
`kvm_create_device` struct passed to it.

This incorrect usage of the `ioctl_with_ref` causes an issue if used with Rust version 1.82(and later) in `release` builds. In those cases the `kvm_create_device` struct is considered to be immutable and the following `File::from_raw_fd` creates a file from an initial value set in the `kvm_create_device` before the call.

The reason unit tests don't fail in CI is because it uses 1.80.1 version and does not run unit tests with `--release` option which does not produce the error.

### Requirements

requirements:

  test.

### DIFF
--- a/kvm-ioctls/src/ioctls/vm.rs
+++ b/kvm-ioctls/src/ioctls/vm.rs
@@ -22,11 +22,11 @@ use crate::ioctls::{KvmRunWrapper, Result};
 use crate::kvm_ioctls::*;
 use vmm_sys_util::errno;
 use vmm_sys_util::eventfd::EventFd;
+#[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
+use vmm_sys_util::ioctl::ioctl;
 #[cfg(target_arch = "x86_64")]
 use vmm_sys_util::ioctl::ioctl_with_mut_ptr;
-#[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
-use vmm_sys_util::ioctl::{ioctl, ioctl_with_mut_ref};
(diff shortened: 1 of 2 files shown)


kvm-ioctls/src/ioctls/vm.rs
```diff
@@ -1300,7 +1300,7 @@ impl VmFd {
     /// ```
     pub fn create_device(&self, device: &mut kvm_create_device) -> Result<DeviceFd> {
         // SAFETY: Safe because we are calling this with the VM fd and we trust the kernel.
-        let ret = unsafe { ioctl_with_ref(self, KVM_CREATE_DEVICE(), device) };
+        let ret = unsafe { ioctl_with_mut_ref(self, KVM_CREATE_DEVICE(), device) };
```
andreeaflorescu, line 1303: Can you add a test that would trigger the error? As we discussed offline this would not be triggering in the CI because we're not using the latest Rust version. It's enough if we can trigger it locally because eventually we'll be using the latest version in the CI as well.

ShadowCurse: We already have tests which can trigger this behavior. Tests are: `test_register_unregister_irqfd` and `test_set_irq_line`.

roypat (approved):
Probably worth it to check all other uses of `ioctl_with_ref` for their correctness, but let's not block this fix on that.

rbradford (approved):

PR #309, opened 2025-03-02
Align folder name to kernel arch

RuoqingHe:
### Summary of the PR

- x86_64: Unify the way of calling serde_impl: We are using curly brackets for aarch64 and riscv64 serialize.rs, let's
stick to the same style.
- riscv: Rename riscv64 to riscv: `riscv64` is the arch name used for Rust, let change it to arch name
used in kernel source to align with other two architectures.

Signed-off-by: Ruoqing He <heruoqing@iscas.ac.cn>

### Requirements

requirements:

  test.


RuoqingHe:
I'm proposing this change because I'm working on a automatic binding generation tool, from downloading the kernel source code of a specific version, extracting, installing headers for architectures we supports, generate bindings and add custom attributes with a single command:

```console
rust-vmm-ci/scripts/rustvmm_gen.py --version 6.12.8 generate_kvm_bindings --attribute "#[cfg_attr(feature = \"serde\", derive(zerocopy::AsBytes, zerocopy::FromBytes, zerocopy::FromZeroes))]"
```

image

roypat:
Mh, riscv support is fairly new, so I think we can probably get away with a breaking change like this, but maybe we should still just add a public re-export of the module under the old name? This is probably a question for the riscv maintainers/users to answer though, no strong opinions from my side :)

RuoqingHe:
> Mh, riscv support is fairly new, so I think we can probably get away with a breaking change like this, but maybe we should still just add a public re-export of the module under the old name? This is probably a question for the riscv maintainers/users to answer though, no strong opinions from my side :)

You are right, I can leave those constants unchanged, only renaming `riscv64` folder and module name instead. In that case, this will not effect any of our downstream, WDYT @roypat 

And this is my fault I didn't realize we are using `kernel arch` here (not `rust arch`) 😔

stefano-garzarella:
@RuoqingHe are you still interested in getting this PR merged? If yes, could you please re-open it in https://github.com/rust-vmm/rust-vmm/. We moved `kvm` crates there.

PR #317, opened 2025-05-05
build: Bump `zerocopy` to 0.8.23

RuoqingHe:
### Summary of the PR

- build: Bump zerocopy to 0.8.23
- Update config.toml to build with 0.8 zerocopy

### Requirements

requirements:

  test.

### DIFF
--- a/kvm-bindings/src/x86_64/bindings.rs
+++ b/kvm-bindings/src/x86_64/bindings.rs
@@ -4,7 +4,7 @@
 #[derive(Copy, Clone, Debug, Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
 #[cfg_attr(
     feature = "serde",
-    derive(zerocopy::AsBytes, zerocopy::FromBytes, zerocopy::FromZeroes)
+    derive(zerocopy::IntoBytes, zerocopy::Immutable, zerocopy::FromBytes)
 )]
 pub struct __BindgenBitfieldUnit<Storage> {
     storage: Storage,
@@ -88,7 +88,7 @@ where
 #[derive(Default)]
 #[cfg_attr(
     feature = "serde",
-    derive(zerocopy::AsBytes, zerocopy::FromBytes, zerocopy::FromZeroes)
+    derive(zerocopy::IntoBytes, zerocopy::Immutable, zerocopy::FromBytes)
 )]
 pub struct __IncompleteArrayField<T>(::std::marker::PhantomData<T>, [T; 0]);
 impl<T> __IncompleteArrayField<T> {
@@ -789,7 +789,7 @@ pub type __poll_t = ::std::os::raw::c_uint;
 #[derive(Debug, Default, Copy, Clone, PartialEq)]
 #[cfg_attr(
     feature = "serde",
-    derive(zerocopy::AsBytes, zerocopy::FromBytes, zerocopy::FromZeroes)
+    derive(zerocopy::IntoBytes, zerocopy::Immutable, zerocopy::FromBytes)
 )]
 pub struct kvm_pic_state {
     pub last_irr: __u8,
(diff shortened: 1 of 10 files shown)


.cargo/config.toml
```diff
@@ -1,5 +1,8 @@
 # This workaround is needed because the linker is unable to find __addtf3,
 # __multf3 and __subtf3.
 # Related issue: https://github.com/rust-lang/compiler-builtins/issues/201
+[build]
+rustflags = "--cfg zerocopy_derive_union_into_bytes"
```
roypat: Does this mean that downstream users of kvm-bindings will not need to set this cfg? iirc, the `--cfg` escape hatch was added because libraries themselves couldn't force it on, unlike feature (which, I personally really dislike - I don't want to publish a library where then every end user has to pass some --cfg to their cargo invocations to get it to work). If not, I'd prefer if we just write out the unsafe trait impls, like we already do for `kvm_irqchip__bindgen_ty_1` in x86_64/serialize.rs. Or we just stay on zerocopy 0.7 until upstream figure out union support again :/

roypat (commented):
thanks for tackling this! Just a question about the cfg thingy below

.cargo/config.toml
RuoqingHe: You are right, I tried this with cloud-hypervisor and we indeed need to pass the same `--cfg` to get it compile 😔

As of the "write out the unsafe trait impls", I didn't succeed in getting around passing the `--cfg`, could you elaborate 🤔

roypat: Zerocopy's derives essentially only implement unsafe marker traits, with the derives statically ensuring the unsafe invariants of the unsafe traits are upheld. Without the `--cfg` flag passed, these derives will simply refuse to implement the marker traits for unions, but it doesn't actually stop those impls from existing in the first place. So we can just manually write out the unsafe trait impl that the derive _would_ have generated, e.g. for all unions in kvm-bindings that we need serde support for, we just have a `unsafe impl FromBytes for <union type> {...}` inside `serde.rs`

RuoqingHe: I'm not sure this is the correct way to do it, please take a look 👀

roypat: Yes, that looks good to me!! Just the comment should be updated to reflect that its now _all_ unions that zerocopy doesnt support anymore

kvm-bindings/src/x86_64/bindings.rs
```diff
@@ -857,10 +857,7 @@ pub struct kvm_ioapic_state {
 }
 #[repr(C)]
 #[derive(Copy, Clone)]
-#[cfg_attr(
-    feature = "serde",
-    derive(zerocopy::AsBytes, zerocopy::FromBytes, zerocopy::FromZeroes)
-)]
+#[cfg_attr(feature = "serde", derive(zerocopy::Immutable, zerocopy::FromBytes))]
```
roypat, line 860: mhh, why no `IntoBytes` for this one? (I guess it compiles so we don't we need it, but I wonder why haha)

kvm-bindings/src/x86_64/bindings.rs
```diff
@@ -3260,6 +3257,7 @@ pub struct kvm_irqchip {
 }
 #[repr(C)]
 #[derive(Copy, Clone)]
+#[cfg_attr(feature = "serde", derive(zerocopy::TryFromBytes))]
```
roypat, line 3260: why `TryFromBytes` for this one? :o

kvm-bindings/src/x86_64/bindings.rs
RuoqingHe: IntoBytes is moved to serialize.rs as a `unsafe impl` now, otherwise we need that `--cfg` to get it compile

kvm-bindings/src/x86_64/bindings.rs
RuoqingHe: Both `FromBytes` and `FromZeros` are demanding this trait to be implemented for `kvm_irqchip__bindgen_ty_1`, `unsafe impl TryFromBytes` seems not working for this because it would require more functions to be implemented manually

roypat (approved):
I am admittedly quite confused by zerocopy allowing us to use _some_ derives on unions but not others, but firecracker compiles against this, and basic snapshot tests pass, so lets ship it!

Thanks for taking up this one :heart:

ShadowCurse (approved):

PR #321, opened 2025-05-19
kvm-ioctls: Remove tests related to invalid FD

jinankjain:
### Summary of the PR

The kvm-ioctl interface does not perform any special handling for invalid
file descriptors; such cases are transparently managed by the kernel.
Therefore, adding a negative test for this behavior does not provide
value when validating a kernel feature.

### Requirements

requirements:

  test.


roypat (commented):
oh wow, I thought we'd caught all of these, rust's been complaining about it since 1.80.0. Wonder how this one slipped through the cracks (especially because the tests actually all pass locally for me, and I'm on 1.85.0). Thanks for the report!

That being said, I'd vote to instead just remove this entire test, because it inherently doesnt really make sense. `File::from_raw_fd(-2)` is undefined behavior, because `from_raw_fd` requires the given fd to be owned and open, which -2 is neither of. We could replace it with something like a `dup()` of stdout, but really, what _are_ we testing here? kvm's ioctls dont work on non-kvm fds? That's something the kernel has to uphold, kvm-ioctls doesnt have any logic related to that

jinankjain:
> oh wow, I thought we'd caught all of these, rust's been complaining about it since 1.80.0. Wonder how this one slipped through the cracks (especially because the tests actually all pass locally for me, and I'm on 1.85.0). Thanks for the report!

Oh, I was using nightly compiler and that's why I started seeing this issue.

jinankjain:
> That being said, I'd vote to instead just remove this entire test, because it inherently doesnt really make sense. `File::from_raw_fd(-2)` is undefined behavior, because `from_raw_fd` requires the given fd to be owned and open, which -2 is neither of. We could replace it with something like a `dup()` of stdout, but really, what _are_ we testing here? kvm's ioctls dont work on non-kvm fds? That's something the kernel has to uphold, kvm-ioctls doesnt have any logic related to that

Do you think we should remove all of these? @roypat 

```
kvm-ioctls/src/ioctls/vcpu.rs
2749:    fn test_faulty_vcpu_fd() {
2796:    fn test_faulty_vcpu_fd_x86_64() {
2941:    fn test_faulty_vcpu_fd_aarch64() {
3022:    fn test_faulty_vcpu_fd_riscv64() {
```

roypat:
> > That being said, I'd vote to instead just remove this entire test, because it inherently doesnt really make sense. `File::from_raw_fd(-2)` is undefined behavior, because `from_raw_fd` requires the given fd to be owned and open, which -2 is neither of. We could replace it with something like a `dup()` of stdout, but really, what _are_ we testing here? kvm's ioctls dont work on non-kvm fds? That's something the kernel has to uphold, kvm-ioctls doesnt have any logic related to that
> 
> Do you think we should remove all of these? @roypat
> 
> ```
> kvm-ioctls/src/ioctls/vcpu.rs
> 2749:    fn test_faulty_vcpu_fd() {
> 2796:    fn test_faulty_vcpu_fd_x86_64() {
> 2941:    fn test_faulty_vcpu_fd_aarch64() {
> 3022:    fn test_faulty_vcpu_fd_riscv64() {
> ```

Yes, I think so, unless anyone else has any objections?

jinankjain:
@andreeaflorescu Any thoughts since you introduced the first test regarding invalid fds?

jinankjain:
@roypat @andreeaflorescu  I have updated the PR with the removal of that test.

roypat (dismissed):

RuoqingHe (commented):
You will need to manually update the coverage in `coverage_config_x86_64.json` as CI suggested to get the tests passing 🙂

roypat (approved):

RuoqingHe:
I don't have write access, we need @rbradford @ShadowCurse to take a look here 🙂

rbradford (approved):

ShadowCurse (approved):

PR #336, opened 2025-08-12
dependabot: drop open-pull-request-limit

roypat:
Currently, dependabot is for some reason stuck saying  "Dependabot cannot open any more pull requests" (the error message one gets if there are many than open-pull-request-limit dependabot PRs open), despite there being exactly 0 open dependabot PRs. We also cannot manually trigger it, as it fails with the same reason.

Try to resolve this by removing the limit on pull requests. In practice, this should not matter, as we have exactly 1 submodule, and the cargo dependencies are all grouped, meaning simply based on how dependabot functions, we never get more than one concurrent PR of each type anyway.

### Summary of the PR

*Please summarize here why the changes in this PR are needed.*

### Requirements

requirements:

  test.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,6 @@ updates:
   directory: "/"
   schedule:
     interval: monthly
-  open-pull-requests-limit: 1
 
 # A monthly update to rust dependencies. These will be grouped,
 # e.g. one PR will contains updates for all dependencies.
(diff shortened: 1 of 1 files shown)


roypat:
I'll also update the dependabot templates over in rust-vmm-ci to not set the open-pull-request-limit once I verify that this works.

stefano-garzarella (approved):

ShadowCurse (approved):

PR #344, opened 2025-09-03
KVM Dirty log ring interface

davidkleymann:
### Summary of the PR

Add support for dirty page tracking via the Dirty ring interface. Adds KvmDirtyLogRing structure for keeping track of the indices and the base pointer to the shared memory buffer. Implements iterating over dirty pages, thereby harvesting them. Implements reset_dirty_rings on VmFd to trigger recycling of dirty ring buffer elements by the kernel after processing. Adds the dirty_log_ring field to VcpuFd.

This is a draft that needs some review and improvements, I'm hereby asking for suggestions for improving the following remaining weaknesses:

1. If any vCPU is running while harvesting dirty pages, the user of KvmDirtyLogRing needs to call reset_dirty_rings *before* reading contents of dirty pages. This is currently the users responsibility, which allows for calling reset_dirty_rings after all dirty pages have been read by the VMM for cases where all vCPUs are halted, but this may be a confusing interface for other cases.

More info on the interface:
https://www.kernel.org/doc/html/latest/virt/kvm/api.html#kvm-cap-dirty-log-ring-kvm-cap-dirty-log-ring-acq-rel

### Requirements

requirements:

  test.

### DIFF
--- a/kvm-ioctls/CHANGELOG.md
+++ b/kvm-ioctls/CHANGELOG.md
@@ -5,6 +5,27 @@
 - Plumb through KVM_CAP_DIRTY_LOG_RING as DirtyLogRing cap.
 - [[#359]](https://github.com/rust-vmm/kvm/pull/359) Add support for `KVM_SET_MSR_FILTER` vm ioctl on x86_64.
 
+### Fixed
+
+- Fixed `VmFd::enable_cap` available for all architectures
+
+### Added
+
+- Added `KvmDirtyLogRing` structure to mmap the dirty log ring.
+- Added `KVM_DIRTY_GFN_F_DIRTY` and `KVM_DIRTY_GFN_F_RESET` bitflags.
+- Added `KvmDirtyLogRing` iterator type for accessing dirty log entries.
+- Added `dirty_log_ring` field to `VcpuFd` to access per-vCpu dirty rings.
+- Inserted fences in KvmDirtyLogRing iterator `next` for architectures with weak memory consistency that require Acquire/Release
+- Added `DirtyLogRingInfo` struct and `dirty_log_ring_info` field to `VmFd` to
+  track dirty ring configuration.
+- Added `check_maximum_dirty_log_ring_size` that checks the maximum dirty log ring size
+  supported by the kernel
+- Added `enable_dirty_log_ring` function on `VmFd` to check corresponding 
+  capabilities and enable KVM's dirty log ring.
+- Added `VcpuFd::dirty_log_ring_iter()` to iterate over dirty guest frame numbers.
+- Added `VmFd::reset_dirty_rings()` to reset all dirty rings for the VM.
+- Added `VcpuExit::DirtyRingFull` for `KVM_EXIT_DIRTY_RING_FULL`.
+
 ## v0.24.0
 
 ### Added
--- a/kvm-ioctls/CHANGELOG.md
+++ b/kvm-ioctls/CHANGELOG.md
@@ -2,6 +2,28 @@
 
 ## Upcoming Release
 
+### Added
+
+- Added `KvmDirtyLogRing` structure to mmap the dirty log ring.
+- Added `KVM_DIRTY_GFN_F_DIRTY` and `KVM_DIRTY_GFN_F_RESET` bitflags.
+- Added `KvmDirtyLogRing` iterator type for accessing dirty log entries.
+- Added `dirty_log_ring` field to `VcpuFd` to access per-vCpu dirty rings.
+- Inserted fences in KvmDirtyLogRing iterator `next` for architectures with weak
+  memory consistency that require Acquire/Release.
+- Added `DirtyLogRingInfo` struct and `dirty_log_ring_info` field to `VmFd` to
+  track dirty ring configuration.
+- Added `check_maximum_dirty_log_ring_size` that checks the maximum dirty log
+  ring size supported by the kernel.
+- Added `enable_dirty_log_ring` function on `VmFd` to check corresponding
+  capabilities and enable KVM's dirty log ring.
+- Added `VcpuFd::dirty_log_ring_iter()` to iterate over dirty guest frame numbers.
+- Added `VmFd::reset_dirty_rings()` to reset all dirty rings for the VM.
+- Added `VcpuExit::DirtyRingFull` for `KVM_EXIT_DIRTY_RING_FULL`.
+
+### Fixed
+
+- Fixed `VmFd::enable_cap` availability for all architectures.
+
 ## v0.25.0
 
 ### Added
--- a/kvm-ioctls/src/ioctls/mod.rs
+++ b/kvm-ioctls/src/ioctls/mod.rs
@@ -8,9 +8,11 @@
 use std::mem::size_of;
 use std::os::unix::io::AsRawFd;
 use std::ptr::{NonNull, null_mut};
+use std::sync::atomic::{Ordering, fence};
 
 use kvm_bindings::{
-    KVM_COALESCED_MMIO_PAGE_OFFSET, kvm_coalesced_mmio, kvm_coalesced_mmio_ring, kvm_run,
+    KVM_COALESCED_MMIO_PAGE_OFFSET, KVM_DIRTY_GFN_F_DIRTY, KVM_DIRTY_GFN_F_RESET,
+    KVM_DIRTY_LOG_PAGE_OFFSET, kvm_coalesced_mmio, kvm_coalesced_mmio_ring, kvm_dirty_gfn, kvm_run,
 };
 use vmm_sys_util::errno;
 
@@ -28,6 +30,110 @@ pub mod vm;
 /// This typedef is generally used to avoid writing out errno::Error directly and
 /// is otherwise a direct mapping to Result.
 pub type Result<T> = std::result::Result<T, errno::Error>;
+
+/// A wrapper around the KVM dirty log ring page.
+#[derive(Debug)]
+pub(crate) struct KvmDirtyLogRing {
+    /// Next potentially dirty guest frame number slot index
+    next_dirty: u64,
+    /// Memory-mapped array of dirty guest frame number entries
+    gfns: NonNull<kvm_dirty_gfn>,
+    /// Ring size mask (size-1) for efficient modulo operations
+    mask: u64,
+}
+
+impl KvmDirtyLogRing {
+    /// Maps the KVM dirty log ring from the vCPU file descriptor.
+    ///
+    /// # Arguments
+    /// * `fd` - vCPU file descriptor to mmap from.
+    /// * `size` - Size of memory region in bytes.
+    pub(crate) fn mmap_from_fd<F: AsRawFd>(fd: &F, bytes: usize) -> Result<Self> {
+        // SAFETY: We trust the sysconf libc function and we're calling it
+        // with a correct parameter.
+        let page_size = match unsafe { libc::sysconf(libc::_SC_PAGESIZE) } {
+            -1 => return Err(errno::Error::last()),
+            ps => ps as usize,
+        };
+
+        let offset = page_size * KVM_DIRTY_LOG_PAGE_OFFSET as usize;
+
+        if bytes % std::mem::size_of::<kvm_dirty_gfn>() != 0 {
+            // Size of dirty ring in bytes must be multiples of slot size
+            return Err(errno::Error::new(libc::EINVAL));
+        }
+        let slots = bytes / std::mem::size_of::<kvm_dirty_gfn>();
+        if !slots.is_power_of_two() {
+            // Number of slots must be power of two
+            return Err(errno::Error::new(libc::EINVAL));
+        }
+
+        // SAFETY: KVM guarantees that there is a page at offset
+        // KVM_DIRTY_LOG_PAGE_OFFSET * PAGE_SIZE if the appropriate
+        // capability is available. If it is not, the call will simply
+        // fail.
+        let gfns = unsafe {
+            NonNull::<kvm_dirty_gfn>::new(libc::mmap(
+                null_mut(),
+                bytes,
+                libc::PROT_READ | libc::PROT_WRITE,
+                libc::MAP_SHARED,
+                fd.as_raw_fd(),
+                offset as i64,
+            ) as *mut kvm_dirty_gfn)
+            .filter(|addr| addr.as_ptr() != libc::MAP_FAILED as *mut kvm_dirty_gfn)
+            .ok_or_else(errno::Error::last)?
+        };
+        Ok(Self {
+            next_dirty: 0,
+            gfns,
+            mask: (slots - 1) as u64,
+        })
+    }
+}
+
+impl Drop for KvmDirtyLogRing {
+    fn drop(&mut self) {
+        // SAFETY: This is safe because we mmap the page ourselves, and nobody
+        // else is holding a reference to it.
+        unsafe {
+            libc::munmap(
+                self.gfns.as_ptr().cast(),
+                (self.mask + 1) as usize * std::mem::size_of::<kvm_dirty_gfn>(),
+            );
+        }
+    }
+}
+
+impl Iterator for KvmDirtyLogRing {
+    type Item = (u32, u64);
+    fn next(&mut self) -> Option<Self::Item> {
+        let i = self.next_dirty & self.mask;
+        // SAFETY: i is not larger than mask, thus is a valid offset into self.gfns,
+        // therefore this operation produces a valid pointer to a kvm_dirty_gfn
+        let gfn_ptr = unsafe { self.gfns.add(i as usize).as_ptr() };
+
+        fence(Ordering::Acquire);
+
+        // SAFETY: Can read a valid pointer to a kvm_dirty_gfn
+        let gfn = unsafe { gfn_ptr.read_volatile() };
+
+        if gfn.flags & KVM_DIRTY_GFN_F_DIRTY == 0 {
+            // next_dirty stays the same, it will become the next dirty element
+            None
+        } else {
+            self.next_dirty += 1;
+            let mut updated_gfn = gfn;
+            updated_gfn.flags ^= KVM_DIRTY_GFN_F_RESET;
+            // SAFETY: Can write to a valid pointer to a kvm_dirty_gfn
+            unsafe {
+                gfn_ptr.write_volatile(updated_gfn);
+            };
+            fence(Ordering::Release);
+            Some((gfn.slot, gfn.offset))
+        }
+    }
+}
 
 /// A wrapper around the coalesced MMIO ring page.
 #[derive(Debug)]
--- a/kvm-ioctls/src/ioctls/vcpu.rs
+++ b/kvm-ioctls/src/ioctls/vcpu.rs
@@ -16,7 +16,7 @@ use libc::EINVAL;
 use std::fs::File;
 use std::os::unix::io::{AsRawFd, RawFd};
 
-use crate::ioctls::{KvmCoalescedIoRing, KvmRunWrapper, Result};
+use crate::ioctls::{KvmCoalescedIoRing, KvmDirtyLogRing, KvmRunWrapper, Result};
 use crate::kvm_ioctls::*;
 use vmm_sys_util::errno;
 use vmm_sys_util::ioctl::{ioctl, ioctl_with_mut_ref, ioctl_with_ref};
@@ -184,6 +184,8 @@ pub enum VcpuExit<'a> {
         /// size
         size: u64,
     },
+    /// Corresponds to KVM_EXIT_DIRTY_RING_FULL.
+    DirtyRingFull,
     /// Corresponds to an exit reason that is unknown from the current version
     /// of the kvm-ioctls crate. Let the consumer decide about what to do with
     /// it.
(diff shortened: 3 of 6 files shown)


kvm-ioctls/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Upcoming Release
 
+- Plumb through KVM_CAP_DIRTY_LOG_RING as DirtyLogRing cap.
```
roypat, line 5: This should go into an `### Added` section

kvm-ioctls/src/ioctls/mod.rs
```diff
@@ -29,6 +29,78 @@ pub mod vm;
...
+            if (*gfn).flags & KVM_DIRTY_GFN_F_DIRTY == 0 {
+                // next_dirty stays the same, it will become the next dirty element
+                return None;
+            } else {
+                self.next_dirty += 1;
+                (*gfn).flags ^= KVM_DIRTY_GFN_F_RESET;
+                return Some(((*gfn).slot, (*gfn).offset));
+            }
+        }
+    }
```
roypat: Here we should probably do just a single `read_volatile` of the `struct kvm_dirty_gfn`, to avoid data races, and write the updated kvm_dirty_gfn flags via `write_volatile` (or, on weakly ordered architectures such as arm64, an atomic read with acquire ordering, and write with release ordering. Which means we'll probably also want to support checking `KVM_CAP_DIRTY_LOG_RING_ACQ_REL`)

kvm-ioctls/src/ioctls/vcpu.rs
```diff
@@ -2104,6 +2106,65 @@ impl VcpuFd {
...
+    /// if kvm.check_extension(Cap::DirtyLogRing) {
+    ///     vcpu.map_dirty_log_ring(4096).unwrap();
+    ///     if let Some(mut iter) = vcpu.dirty_log_ring_iter() {
+    ///         for (slot, offset) in iter {
+    ///             println!("Dirty page in slot {} at offset {}", slot, offset);
+    ///         }
+    ///     }
+    /// }
+    /// ```
+    pub fn dirty_log_ring_iter(&mut self) -> Option<&mut KvmDirtyLogRing> {
```
roypat: we can hide the actual iterator type here by just returning `impl Iterator<Item = blablabla>`. Then maybe the entire struct doesnt need to be exported?

kvm-ioctls/src/ioctls/vm.rs
```diff
@@ -1930,14 +1930,14 @@ impl VmFd {
     /// }
     /// ```
     ///
-    pub fn reset_dirty_rings(&self) -> Result<()> {
+    pub fn reset_dirty_rings(&self) -> Result<c_int> {
```
roypat, line 2204: this should go into the previous commit

kvm-ioctls/src/ioctls/vm.rs
```diff
@@ -1915,8 +1915,8 @@ impl VmFd {
         Ok(())
     }
 
-    /// Resets all vCPU's dirty log rings.
-    ///
+    /// Resets all vCPU's dirty log rings. This notifies the kernel that pages have been harvested
+    /// from the dirty ring and the corresponding pages can be reprotected.
```
roypat, line 2187: this should also just go into the commit that introduced the function

kvm-ioctls/src/ioctls/vcpu.rs
```diff
@@ -2106,8 +2106,7 @@ impl VcpuFd {
         }
     }
 
-    /// Maps the coalesced MMIO ring page. This allows reading entries from
-    /// the ring via [`coalesced_mmio_read()`](VcpuFd::coalesced_mmio_read).
+    /// Maps the KVM dirty log ring.
```
roypat: this should go into the commit that introduced the function

kvm-ioctls/src/ioctls/mod.rs
roypat, line 1: and all the changes to this file should go into the commit that introduced struct KvmDirtyLogRing

kvm-ioctls/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Upcoming Release
 
 - Plumb through KVM_CAP_DIRTY_LOG_RING as DirtyLogRing cap.
+- Added support for dirty log ring interface introducing `VcpuFd::reset_dirty_rings`, `KvmDirtyLogRing`
```
roypat: Just add the changelog entries in the commit that introduces the respective functions.

kvm-ioctls/src/ioctls/mod.rs
```diff
@@ -54,17 +51,29 @@ impl KvmDirtyLogRing {
...
+        };
+
+        let offset  = page_size * KVM_DIRTY_LOG_PAGE_OFFSET as usize;
+
         if bytes % std::mem::size_of::<kvm_dirty_gfn>() != 0 {
             // Size of dirty ring in bytes must be multiples of slot size
             return Err(errno::Error::new(libc::EINVAL));
         }
         let slots = bytes / std::mem::size_of::<kvm_dirty_gfn>();
         if slots & (slots - 1) != 0 {
```
roypat: `slots.is_power_of_two()`

kvm-ioctls/src/ioctls/vcpu.rs
```diff
@@ -2104,6 +2106,65 @@ impl VcpuFd {
...
+    ///     vcpu.coalesced_mmio_ring().unwrap();
+    /// }
+    /// ```
+    pub fn map_dirty_log_ring(&mut self, bytes: usize) -> Result<()> {
+        if self.dirty_log_ring.is_none() {
+            let ring = KvmDirtyLogRing::mmap_from_fd(&self.vcpu, bytes)?;
+            self.dirty_log_ring = Some(ring);
+        }
+        Ok(())
+    }
```
roypat: the dirty log ring interface is enabled via `KVM_ENABLE_CAP` (which we expose as `enable_cap()` in this crate), which is where the size of the ring buffer is specified as well. So I would propose to make the API here a wrapper on top of `enable_cap()`, say `enable_dirty_log_ring(bytes: usize)`, which does the VmFd::enable_cap() call and then stores the actual size of the ring buffer directly in the VmFd struct. Then on VcpuFd creation, we check if the dirty ring capability was ever enabled on the VmFd, and if so, just mmap the ring with the size stored in the VmFd at vcpu creation time (maybe disallow calling enable_dirty_log_ring() if vcpus were already created previously, although KVM also already checks this). That way the issue of "what is the correct size to mmap" goes away.

roypat (changes requested):
please squash your commits so that we do not have later commits be fixups for previous commits.

roypat:
Also, do we need to do anything about KVM_CAP_DIRTY_LOG_RING_WITH_BITMAP?

roypat:
> Send and Sync are probably not safe on KvmDirtyLogRing because accesses are not stateless, due to the state we need to keep track of in user space (next_dirty)

As for this, `next_dirty` doesnt imply that it cannot be Send/Sync, because the traits dont mean that multiple threads can access the struct at the same time (thats always unsound in Rust's memory model), just that you can transfer ownership between threads, and that read-only references can be shared between threads. A problem would only arrive if we can also `Clone` this structure, because then the iterator can cause data races inside the mmap'd area from different Rust threads. But as long as for each vcpu we only allow creating a single instance of the iterator (using safe code that is), there are no problems.

Now, whether Send and Sync will actually be useful is a different matter, because I'm assuming you want them to be able to harvest the dirty ring while the vcpu is still running, but with the current API in this PR, you can only get a &mut reference to the ring buffer structure, which you cannot send to another thread anyway. So in this case, you'd need some API that lets you take ownership of the KvmDirtyRingLog structure (maybe store the one that gets created at vcpu creation time in an option, and then have a function that just wraps .take() on that option, keeping in mind that we must never allow safe code to get two owned versions of it for the same vcpu).

roypat:
Also sorry for the late response, we were all preparing for / traveling to KVM Forum last week!

kvm-ioctls/src/ioctls/mod.rs
davidkleymann: Is supporting arm64 mandatory, or can we gate the dirty ring functionality based on the arch? On arm64, apparently dirty rings need not only KVM_CAP_DIRTY_LOG_RING_ACQ_REL but also KVM_CAP_DIRTY_LOG_RING_WITH_BITMAP

kvm-ioctls/src/ioctls/vcpu.rs
davidkleymann: That would be a more elegant solution. So you would also opt for returning `Option<impl Iterator<Item = (u32,u64)>>` instead of `Option<impl Iterator<Item = <KvmDirtyLogRing as Iterator>::Item>>`?

kvm-ioctls/src/ioctls/mod.rs
roypat: I think even if we don't support arm64 initially (although, is it really that much more complicated to also do the KVM_CAP_DIRTY_LOG_RING_WITH_BITMAP thing? Sorry, it's been a while since I looked at all this), I think we should make the code forward compatible for when we _do_ add arm64 support.

kvm-ioctls/src/ioctls/vcpu.rs
roypat: Yeah, the compiler will yell if we mess it up and the types don't end up matching for whatever reason (as I don't think we _can_ do the latter with the struct itself not exported, right?) :)

davidkleymann:
> Also, do we need to do anything about KVM_CAP_DIRTY_LOG_RING_WITH_BITMAP?

Turns out we do, and if it is supported, we also need to enable KVM_CAP_DIRTY_LOG_RING_ACQ_REL prior to enabling _WITH_BITMAP.  Correct me if I'm wrong.

davidkleymann:
Why is enable_cap not available in aarch64?
`#[cfg(any(target_arch = "x86_64", target_arch = "s390x", target_arch = "powerpc"))]`

roypat:
> Why is enable_cap not available in aarch64? `#[cfg(any(target_arch = "x86_64", target_arch = "s390x", target_arch = "powerpc"))]`

~~according to the docs, it doesn't exist on arm: https://docs.kernel.org/virt/kvm/api.html#kvm-enable-cap . But it seems the docs are wrong there~~ (they're not, there's two ioctls, one for vcpus and one for vms, with the vcpu one being arch-specific, and the VM one documented as "architectures: all"), because there is a generic implementation in the kernel for some capabilities (including the dirty log related ones) which is available on all arches. So kvm-ioctls is doing it wrong (probably whoever wrote it also misread the docs). Can you change `Vm::enable_cap` to be available always (e.g. without any cfgs) and add a changelog entry under "fixed"?

kvm-ioctls/src/ioctls/mod.rs
davidkleymann: No I actually added support for KVM_CAP_DIRTY_LOG_RING_WITH_BITMAP, But still need to work on the atomic acq/rel accessing

davidkleymann:
The enable_cap doctest uses KVM_CAP_SPLIT_IRQCHIP as an example, which is x86 only. This breaks the test and I disabled the enable_cap call in the test on other arches.

davidkleymann:
> > Send and Sync are probably not safe on KvmDirtyLogRing because accesses are not stateless, due to the state we need to keep track of in user space (next_dirty)
> 
> As for this, `next_dirty` doesnt imply that it cannot be Send/Sync, because the traits dont mean that multiple threads can access the struct at the same time (thats always unsound in Rust's memory model), just that you can transfer ownership between threads, and that read-only references can be shared between threads. A problem would only arrive if we can also `Clone` this structure, because then the iterator can cause data races inside the mmap'd area from different Rust threads. But as long as for each vcpu we only allow creating a single instance of the iterator (using safe code that is), there are no problems.
> 
> Now, whether Send and Sync will actually be useful is a different matter, because I'm assuming you want them to be able to harvest the dirty ring while the vcpu is still running, but with the current API in this PR, you can only get a &mut reference to the ring buffer structure, which you cannot send to another thread anyway. So in this case, you'd need some API that lets you take ownership of the KvmDirtyRingLog structure (maybe store the one that gets created at vcpu creation time in an option, and then have a function that just wraps .take() on that option, keeping in mind that we must never allow safe code to get two owned versions of it for the same vcpu).

I needed Send and Sync to move the Vcpus to separate threads, but I only implemented what the rust compiler suggested and am still looking for a better solution for this unrelated problem.

davidkleymann:
Everything should be ready

roypat:
> Everything should be ready

Please still squash your commits (and there's also conflicts, although GitHub isn't telling me _what_ they are for some reason :thinking:)

davidkleymann:
So you want everything squashed into 1 commit?

roypat:
> So you want everything squashed into 1 commit?

I will tentatively say "yes" (sorry, been a while since I looked at the actual changes here). But please have a look and see if there's anything that can be logically separated into their own commits (e.g. preliminary refactors, or changes to tests).

davidkleymann:
I did change quite a few things so I already rebased it to seperate the commits logically, but if you require a single large commit I can change it as well. I can't really see the conflicts either, I suspect it is just because the commit introducing the DirtyLogRing cap has already been merged to main, while this PR contains my unmerged one.

davidkleymann:
@roypat I squashed everything as requested.

zolutal:
Hi @davidkleymann , thanks for your work on this feature! I was testing it out and noticed that no VcpuExit is currently defined for `KVM_EXIT_DIRTY_RING_FULL`. Currently you just get a `VcpuExit::Unsupported(31)` when the ring buffer fills up. If it doesn't make it in this PR I can make one myself to add it, but I figured I'd mention it here.

roypat:
> Hi @davidkleymann , thanks for your work on this feature! I was testing it out and noticed that no VcpuExit is currently defined for `KVM_EXIT_DIRTY_RING_FULL`. Currently you just get a `VcpuExit::Unsupported(31)` when the ring buffer fills up. If it doesn't make it in this PR I can make one myself to add it, but I figured I'd mention it here.

that is a good catch, thanks! This should definitely be added in this PR :)

davidkleymann:
I can't re-request review, but I added the `VcpuExit` @zolutal  @roypat

roypat:
thanks! could you rebased on main again? it seems like you accidentally picked up some commits from somewhere that shouldn't be on this branch.

@ShadowCurse could you also have a look at this PR please?

davidkleymann:
Thanks, rebased.

kvm-ioctls/src/ioctls/mod.rs
```diff
@@ -29,6 +31,121 @@ pub mod vm;
...
+#[derive(Debug)]
+pub(crate) struct KvmDirtyLogRing {
+    /// Next potentially dirty guest frame number slot index
+    next_dirty: u64,
+    /// Memory-mapped array of dirty guest frame number entries
+    gfns: NonNull<kvm_dirty_gfn>,
+    /// Ring size mask (size-1) for efficient modulo operations
+    mask: u64,
+    /// `true` if we need to use Acquire/Release memory ordering
+    use_acq_rel: bool,
```
ShadowCurse: Why not always use `acq_rel`? On x86_64 `fence(...)` will most likely be a `nop`, while on aarch64/riscv it will have a correct behavior. There is no reason to waste 64 bits for a single boolean and then some cycles for unnecessary branches.

kvm-ioctls/src/ioctls/vm.rs
```diff
@@ -2050,6 +2078,131 @@ impl VmFd {
...
+    /// # use kvm_bindings::{ KVM_CAP_DIRTY_LOG_RING, };
+    ///
+    /// let kvm = Kvm::new().unwrap();
+    /// let mut vm = kvm.create_vm().unwrap();
+    /// let max_supported_size = vm.check_extension_raw(KVM_CAP_DIRTY_LOG_RING.into());
+    /// vm.enable_dirty_log_ring(Some(max_supported_size));
+    /// // Create one vCPU with the ID=0 to cause a dirty log ring to be mapped.
+    /// let vcpu = vm.create_vcpu(0);
+    /// ```
+    pub fn enable_dirty_log_ring(&mut self, bytes: Option<i32>) -> Result<bool> {
```
ShadowCurse: How about instead of providing `bytes` as an argument, we provide a number of entries in the log we want? So `entries: Option<NonZeroU32>`? This will make it harder to make a mistake + less validation checks inside.
This can also be done for the `KvmDirtyLogRing::mmap_from_fd`

kvm-ioctls/src/ioctls/mod.rs
davidkleymann: It does seem not to generate any fence instruction on x86_64, but it does check if fence() was called with Ordering::Relaxed in which case it would panic. With `-C opt-level=3` it is optimized out though.

I think it's acceptable to always call fence() and I will add this change

kvm-ioctls/src/ioctls/vm.rs
davidkleymann: This decision was mainly out of convenience since the kernel also returns the size in bytes when checking the cap. The extra checks are just for better error handling.

davidkleymann: I can change it to providing number of entries but in that case it would also be helpful to implement something that wraps `vm.check_extension_raw(KVM_CAP_DIRTY_LOG_RING.into())` and returns the max number of entries.

kvm-ioctls/src/ioctls/vm.rs
```diff
@@ -2049,6 +2067,160 @@ impl VmFd {
...
+    /// if kvm.check_extension(Cap::DirtyLogRing) {
+    ///     let max_supported_size = vm
+    ///         .check_maximum_dirty_log_ring_size()
+    ///         .expect("Should be supported");
+    ///     vm.enable_dirty_log_ring(Some(max_supported_size));
+    ///     // Create one vCPU with the ID=0 to cause a dirty log ring to be mapped.
+    ///     let vcpu = vm.create_vcpu(0);
+    /// }
+    /// ```
+    pub fn enable_dirty_log_ring(&mut self, entries: Option<usize>) -> Result<bool> {
```
ShadowCurse: for `entries` use `Option<NonZeroUsize>`, it will remove the need for 0 check.

kvm-ioctls/src/ioctls/vm.rs
```diff
@@ -2049,6 +2067,160 @@ impl VmFd {
...
+                } else {
+                    (0, 0, false)
+                }
+            }
+        };
+
+        if dirty_ring_cap == 0 {
+            // Neither KVM_CAP_DIRTY_LOG_RING nor KVM_CAP_DIRTY_LOG_RING_ACQ_REL are available
+            return Err(errno::Error::new(libc::EOPNOTSUPP));
+        }
```
ShadowCurse: Just move the error path into `else` branch

kvm-ioctls/src/ioctls/vm.rs
```diff
@@ -2049,6 +2067,160 @@ impl VmFd {
...
+
+        // Return EINVAL if multiplication overflows
+        let cap_ring_size = entries
+            .map(|e| <usize as TryInto<i32>>::try_into(e * std::mem::size_of::<kvm_dirty_gfn>()))
+            .unwrap_or(Ok(max_bytes))
+            .map_err(|_| errno::Error::new(libc::EINVAL))?;
```
ShadowCurse: Maybe replace this with simple `let cap_ring_size: u64 = e as u64 * std::mem::size_of::<kvm_dirty_gfn>() as u64;`? If you really want for overflow to not happen just make `entries` value be u32. Because `cap_ring_size` is u64(usize), the multiplication of u32 and some small number will not overflow.
I see no point in casting all of this into `i32`. If `check_extension_raw` returns positive number, you can cast it to u64 for simpler math down the line.

.vscode/settings.json
```diff
@@ -0,0 +1,44 @@
+{
```
ShadowCurse: pretty sure this file should not be here

davidkleymann: thanks. fixed

ShadowCurse (dismissed):

roypat (dismissed):

davidkleymann:
When will this be merged?

roypat:
I've just trigged dependabot, which should give us a PR for fixing the CI failure, and then we can merge this immediately after! Sorry for the long wait :(

davidkleymann:
Thanks. No worries, just making sure it gets merged eventually :)

kvm-ioctls/src/ioctls/vcpu.rs
```diff
@@ -197,6 +199,9 @@ pub struct VcpuFd {
     kvm_run_ptr: KvmRunWrapper,
     /// A pointer to the coalesced MMIO page
     coalesced_mmio_ring: Option<KvmCoalescedIoRing>,
+    /// A pointer to the dirty log ring
+    #[allow(unused)]
```
ludfjig: nit: I think this `#[allow(unused)]` potentially can be removed

ludfjig (approved):
lgtm (the latest main branch merge commit should maybe have been a rebase instead?)

kvm-ioctls/src/ioctls/vcpu.rs
davidkleymann: fixed

davidkleymann:
Rebased and removed an unneccessary #[allow(unused)].

davidkleymann:
CI is green, @roypat mind giving it a look?

stefano-garzarella:
@davidkleymann are you still interested in getting this PR merged? If yes, could you please re-open it in https://github.com/rust-vmm/rust-vmm/. We moved `kvm` crates there.

PR #345, opened 2025-09-09
Add riscv64 sbi bindings

BillXiang:
### Summary of the PR

- Generate and add bindings for riscv64 sbi ecall interface from qemu-10.1.0 target\riscv\sbi_ecall_interface.h
using bindgen-cli 0.72.1.
- Handle console sbi call, which implement early console io while apply 'earlycon=sbi' into kernel parameters.

### Requirements

requirements:

  test.


roypat:
@RuoqingHe could you have a look at this one? :o

RuoqingHe (commented):
6e53a2a7 should be squashed into other commits

kvm-bindings/src/riscv64/mod.rs
```diff
@@ -5,12 +5,14 @@
 #[allow(clippy::all)]
 #[allow(clippy::undocumented_unsafe_blocks)]
 pub mod bindings;
+pub mod sbi_bindings;
```
RuoqingHe: Please align the bindgen version with the version we used to generate kvm-bindings

kvm-ioctls/src/ioctls/vcpu.rs
```diff
@@ -1688,6 +1692,20 @@ impl VcpuFd {
...
+                    match unsafe {run.__bindgen_anon_1.riscv_sbi.extension_id} as u32 {
+                        SBI_EXT_0_1_CONSOLE_PUTCHAR => {
+                            let ch = unsafe { run.__bindgen_anon_1.riscv_sbi.args[0] };
+                            Ok(VcpuExit::SbiExt0_1ConsolePutchar(ch))
+                        }
+                        SBI_EXT_0_1_CONSOLE_GETCHAR => {
+                            let ch = unsafe { &mut run.__bindgen_anon_1.riscv_sbi.ret[..1] };
+                            Ok(VcpuExit::SbiExt0_1ConsoleGetchar(ch))
+                        }
+                        r => Ok(VcpuExit::Unsupported(r)),
```
RuoqingHe, line 1712: Is this match arm correct, maybe `_`?

kvm-ioctls/src/ioctls/vcpu.rs
```diff
@@ -1688,6 +1692,20 @@ impl VcpuFd {
...
+                    }
+                }
```
RuoqingHe: IMHO we should just return `__bindgen_anon_1.riscv_sbi` here, leave the vmm to decide what to do next

RuoqingHe (commented):
IMO we should only introduce `KVM_EXIT_RISCV_SBI` here and return `__bindgen_anon_1.riscv_sbi` for vmm to further decide

And those sbi bindings are not from kernel, this would not be the right place for them to live in, perhaps add them to the VMM you are working on 🙂

BillXiang:
Maybe it's better to add those sbi bindings to the VMM.

kvm-ioctls/src/ioctls/vcpu.rs
```diff
@@ -1688,6 +1692,26 @@ impl VcpuFd {
                     Ok(VcpuExit::IoapicEoi(eoi.vector))
                 }
                 KVM_EXIT_HYPERV => Ok(VcpuExit::Hyperv),
+                #[cfg(target_arch = "riscv64")]
+                KVM_EXIT_RISCV_SBI => {
+                    // SAFETY: Safe because the exit_reason (which comes from the kernel) told us
+                    // which union field to use and the type of extension_id is 'enum sbi_ext_id'.
+                    match unsafe { run.__bindgen_anon_1.riscv_sbi.extension_id } as u32 {
+                        SBI_EXT_0_1_CONSOLE_PUTCHAR => {
```
rbradford, line 1700: I don't think these should be handled at the KVM crate level - they should be done by the VMM.

rbradford (changes requested):

BillXiang:
I noticed that [sbi_spec ](https://docs.rs/sbi-spec/latest/sbi_spec/) has implemented SBI. And I have submitted a [new pr ](https://github.com/rust-vmm/kvm/pull/352)that return riscv_sbi to VMM for further process. @RuoqingHe  @rbradford

PR #352, opened 2025-09-26
Return riscv_sbi to VMM for further process

BillXiang:
### Summary of the PR

- Return riscv_sbi to VMM for further process.
- Add test for KVM_EXIT_RISCV_SBI exit.

### Requirements

Before submitting your PR, please make sure you addressed the following
requirements:

- [x] All commits in this PR have Signed-Off-By trailers (with
  `git commit -s`), and the commit message has max 60 characters for the
  summary and max 75 characters for each description line.
- [x] All added/changed functionality has a corresponding unit/integration
  test.
- [x] All added/changed public-facing functionality has entries in the "Upcoming 
  Release" section of CHANGELOG.md (if no such section exists, please create one).
- [x] Any newly added `unsafe` code is properly documented.

PR #354, opened 2025-10-13
build: Bump `rust-vmm-ci` to `c41370c`

RuoqingHe:
### Summary of the PR

Test cases introduced by SBI related changes require SBI_V01 to be enabled during kernel compilation. `c41370c` is capable of covering this case.

Unblocks #352.

### Requirements

requirements:

  test.


ShadowCurse (dismissed):

roypat:
seems like no_run tests are still compiled. I guess this one needs some imports behind `#` and the size of `sizeof` -> `size_of` then

RuoqingHe:
> seems like no_run tests are still compiled. I guess this one needs some imports behind `#` and the size of `sizeof` -> `size_of` then

I'm in the middle of addressing this, it looks like this snippet is outdated, it looks like `KVM_CHECK_EXTENSION(KVM_CAP_XSAVE2)` will not work, I trying to use `ioctls_with_val` with `KVM_CAP_XSAVE2` to get the value here

roypat:
> > seems like no_run tests are still compiled. I guess this one needs some imports behind `#` and the size of `sizeof` -> `size_of` then
> 
> I'm in the middle of addressing this, it looks like this snippet is outdated, it looks like `KVM_CHECK_EXTENSION(KVM_CAP_XSAVE2)` will not work, I trying to use `ioctls_with_val` with `KVM_CAP_XSAVE2` to get the value here

admittedly, this snippet was never meant to actually compile or anything, it was just some pseudo code to illuminate the calculation

kvm-bindings/src/x86_64/fam_wrappers.rs
```diff
@@ -135,8 +135,27 @@ pub struct kvm_xsave2 {
...
+    ///
+    /// ((unsafe {
+    ///     ioctl_with_val(
+    ///         &File::from_raw_fd(kvm_fd),
+    ///         KVM_CHECK_EXTENSION(),
+    ///         KVM_CAP_XSAVE2.into(),
+    ///     )
+    /// }) as usize
+    ///     - size_of::<kvm_xsave>())
+    /// .div_ceil(size_of::<__u32>());
```
roypat: I really do think this is clearer as just the pseudo code it was before :/

RuoqingHe: Should I just remove this code snippet (the surrounding ```), and perhaps use `(KVM_CHECK_EXTENSION(KVM_CAP_XSAVE2) - sizeof::<kvm_xsave>()).div_ceil(sizeof::<__u32>())` instead

roypat: yeah, that'd be my preference

stefano-garzarella (approved):

roypat (approved):

PR #359, opened 2025-10-29
kvm-ioctls: Add KVM_X86_SET_MSR_FILTER vm ioctl

### LINKED ISSUE #358: x86: KVM_X86_SET_MSR_FILTER vm ioctl
Hi I am interested in [KVM_X86_SET_MSR_FILTER](https://docs.kernel.org/virt/kvm/api.html#kvm-x86-set-msr-filter) which does not yet have rust bindings. Are there any plans on adding this, or can I submit PR for it?

ludfjig:
### Summary of the PR

Adds KVM_X86_SET_MSR_FILTER vm ioctl. This is my first contribution so I might be missing something, feedback greatly appreciated. I'm not sure whether there needs to be an actual test running vcpu that tries to read/write some MSR

Closes #358 

### Requirements

requirements:

  test.

### DIFF
--- a/kvm-ioctls/src/ioctls/vm.rs
+++ b/kvm-ioctls/src/ioctls/vm.rs
@@ -668,7 +668,7 @@ impl VmFd {
             dst.flags = src.flags.bits();
             dst.nmsrs = src.msr_count;
             dst.base = src.base;
-            dst.bitmap = src.bitmap.as_ptr() as *mut u8; // TODO: is this cast ok? kvm_msr_filter_range.bitmap is __*mut__ u8. Ideally I don't want to require input parameters to be mutable unless it's necessary
+            dst.bitmap = src.bitmap.as_ptr() as *mut u8; // The ioctl doesn't modify the bitmap, so this cast is safe.
         }
 
         // SAFETY: We checked the length of all bitmaps above.
--- a/kvm-ioctls/src/ioctls/vm.rs
+++ b/kvm-ioctls/src/ioctls/vm.rs
@@ -533,6 +533,44 @@ impl VmFd {
         }
     }
 
+    /// Sets the MSR filter as per the `KVM_X86_SET_MSR_FILTER` ioctl.
+    ///
+    /// See the documentation for `KVM_X86_SET_MSR_FILTER` in the
+    /// [KVM API doc](https://www.kernel.org/doc/Documentation/virtual/kvm/api.txt).
+    ///
+    /// # Arguments
+    ///
+    /// * `filter` - MSR filter configuration to be set.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the given `kvm_msr_filter` is valid. Specifically any `bitmap` pointers in the `kvm_msr_filter_range`
+    /// structures within `filter.ranges` must point to valid memory of sufficient size.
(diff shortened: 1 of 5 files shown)


ludfjig:
CI is red, but looks unrelated to this change I believe

RuoqingHe:
> CI is red, but looks unrelated to this change I believe

Yes, leave it to me

roypat (commented):
I wonder if we could have a slightly higher level, but safe API here. E.g. kvm_msr_filer::bitmap is a pointer to array of u8s that has nmsrs many bits. We could just have an function that takes a Vec<u8>, range checks against nmsrs, and then convert its arguments to the `kvm_msr_filter` structure and does the ioctl maybe? The other args could even be enums then I think

ludfjig:
> I wonder if we could have a slightly higher level, but safe API here. E.g. kvm_msr_filer::bitmap is a pointer to array of u8s that has nmsrs many bits. We could just have an function that takes a Vec, range checks against nmsrs, and then convert its arguments to the `kvm_msr_filter` structure and does the ioctl maybe? The other args could even be enums then I think

Thanks for reviewing. That sounds reasonable to me. Do you prefer to have 2 versions of it around (keep the old function around as an unsafe `_unchecked()`), or do you only want the safe one?

kvm-ioctls/src/ioctls/vm.rs
```diff
@@ -533,6 +577,103 @@ impl VmFd {
...
+            // Validate that the provided bitmap is large enough to hold the specified number of MSRs.
+            let required_bytes = src.msr_count.div_ceil(8) as usize;
+            if src.bitmap.len() < required_bytes {
+                return Err(errno::Error::new(libc::EINVAL));
+            }
+
+            dst.flags = src.flags.bits();
+            dst.nmsrs = src.msr_count;
+            dst.base = src.base;
+            dst.bitmap = src.bitmap.as_ptr() as *mut u8; // TODO: is this cast ok? kvm_msr_filter_range.bitmap is __*mut__ u8. Ideally I don't want to require input parameters to be mutable unless it's necessary
```
ludfjig: Note this comment

roypat:
Sorry for taking a while to get back to this, life's been busy :sob: 

I think having both higher and lower level functions is good, so that if the ioctl ever gets updated, people can use the low level function without needing to wait for the crate to update the high level apis. 

as for the cast, yea, this should be fine as the ioctl really does not actually write to `bitmap`

roypat (dismissed):

ludfjig:
> Sorry for taking a while to get back to this, life's been busy 😭
> 
> I think having both higher and lower level functions is good, so that if the ioctl ever gets updated, people can use the low level function without needing to wait for the crate to update the high level apis.

No worries. Thanks for the feedback. 

> as for the cast, yea, this should be fine as the ioctl really does not actually write to `bitmap`

Thanks, removed the TODO comment.

kvm-ioctls/src/ioctls/vm.rs
ludfjig: Determined to be fine given that ioctl does not modify the bitmap

roypat (approved):

stefano-garzarella (approved):

PR #360, opened 2025-10-30
fam_wrappers: Remove code snippet of kvm_save2.len

RuoqingHe:
### Summary of the PR

Code snippet for len field of kvm_xsave2 is outdate for a long while, which also fails to compile. Change the code snippet to pseudo code.

### Requirements

requirements:

  test.

### DIFF
--- a/kvm-bindings/src/x86_64/fam_wrappers.rs
+++ b/kvm-bindings/src/x86_64/fam_wrappers.rs
@@ -132,12 +132,10 @@ pub type MsrList = FamStructWrapper<kvm_msr_list>;
 pub struct kvm_xsave2 {
     /// The length, in units of sizeof::<__u32>(), of the FAM in [`kvm_xsave`].
     ///
-    /// Note that `KVM_CHECK_EXTENSION(KVM_CAP_XSAVE2)` returns the size of the entire
-    /// `kvm_xsave` structure, e.g. the sum of header and FAM. Thus, this `len` field
-    /// is equal to
-    /// ```norun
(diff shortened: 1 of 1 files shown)


kvm-bindings/src/x86_64/fam_wrappers.rs
```diff
@@ -134,10 +134,8 @@ pub struct kvm_xsave2 {
     ///
     /// Note that `KVM_CHECK_EXTENSION(KVM_CAP_XSAVE2)` returns the size of the entire
     /// `kvm_xsave` structure, e.g. the sum of header and FAM. Thus, this `len` field
-    /// is equal to
-    /// ```norun
-    /// (KVM_CHECK_EXTENSION(KVM_CAP_XSAVE2) - sizeof::<kvm_xsave>()).div_ceil(sizeof::<__u32>())
-    /// ```
+    /// is equal to `KVM_CHECK_EXTENSION(KVM_CAP_XSAVE2)`, which should also be equal
+    /// to `sizeof::<kvm_xsave>()).div_ceil(sizeof::<__u32>())`.
```
roypat: been a while since I looked at this, but IIRC the len field describes the number of bytes that the struct contains in addition to the header, so the `size_of::kvm_xsave>()` is needed (its the len of only the FAM part, not the size of the entire thing)

RuoqingHe: Fixed, I was too careless, sorry about that

roypat (approved):

stefano-garzarella (approved):

PR #375, opened 2026-06-03
x86: Add TDX support

### LINKED ISSUE #374: No support for KVM_EXIT_TDX
In linux 6.16, there is a new KVM_EXIT type defined in include/uapi/linux/kvm.h:

`#define KVM_EXIT_TDX              40`

We need this exit type to implement a TDX VM, so is it possible to add the implementation in VcpuExit?

00xc:
### Summary of the PR

Add all the required interfaces to be able to create TDX VMs. This includes:

* Updating kvm-bindings to include new structures and definitions.
* Add FAM wrappers for the TDX structs that need it.
* Enable using `KVM_CAP_VM_TYPES` to detect if KVM supports TDX.
* Enable handling `KVM_EXIT_TDX` and accessing the right fields in `kvm_run`.
* Add the required TDX-specific `KVM_MEMORY_ENCRYPT_OP` variants (i.e. the TDX side of `VmFd::encrypt_op_sev()`.

This closes #374.

### Requirements

requirements:

  test.


00xc:
The unit test failure on arm64 is unrelated. I opened #376 to fix it.

00xc:
> The unit test failure on arm64 is unrelated. I opened #376 to fix it.

...and the x86 failures should be fixed in #377

RainaYL:
Thanks a lot for the amazing work! Our team is developing a lightweight hypervisor that supports TDX VM, and it would be very helpful if this PR could be merged.

RainaYL:
@00xc Could you add pub use for TdxExit enum? Or users would not be able to access its variants from outside the crate.

00xc:
> @00xc Could you add pub use for TdxExit enum? Or users would not be able to access its variants from outside the crate.

Done!

kvm-ioctls/src/ioctls/vcpu.rs
```diff
@@ -1246,6 +1246,70 @@ impl VcpuFd {
...
+    ///     id: kvm_tdx_cmd_id_KVM_TDX_INIT_VCPU,
+    ///     data: 0,
+    ///     ..Default::default()
+    /// };
+    /// vcpu.encrypt_op_tdx(&mut cmd).unwrap();
+    /// ```
+    #[cfg(target_arch = "x86_64")]
+    pub fn encrypt_op_tdx(&self, op: &mut kvm_tdx_cmd) -> Result<()> {
+        // SAFETY: we trust the kernel and verified parameters
+        let ret = unsafe { ioctl_with_mut_ptr(self, KVM_MEMORY_ENCRYPT_OP(), op) };
```
olivereanderson: Please mark this method as unsafe: It is easy for the caller to create UB without invoking unsafe the way it is now.

Here is a simple example:
```
let cpuid = kvm_cpuid2::default();
let cpuid_ref = &cpuid;
let cmd = kvm_tdx_cmd {id: 5, flags: 0, data: std::ptr::from_ref(cpuid_ref) as u64, hw_error: 0};  
vcpu.encrypt_op_tdx(&mut cmd); // UB: This alters cpuid while a borrow is alive
dbg!(cpuid_ref); 
```

It should not be possible to invoke UB from safe Rust even if the caller provides wrong/bad parameters.

You may also consider creating safe methods for each individual command where it makes sense: e.g.
```
Vcpu::tdx_get_cpuid(&self, cpuid: &mut cpuid2, hw_error: &mut u64) -> Result<()>;
```

kvm-ioctls/src/ioctls/vm.rs
```diff
@@ -1836,6 +1836,53 @@ impl VmFd {
...
+    ///     data: caps.as_mut_fam_struct_ptr() as u64,
+    ///     ..Default::default()
+    /// };
+    /// vm.encrypt_op_tdx(&mut cmd).unwrap();
+    /// ```
+    #[cfg(target_arch = "x86_64")]
+    pub fn encrypt_op_tdx(&self, op: &mut kvm_tdx_cmd) -> Result<()> {
+        // SAFETY: Safe because we know that kernel will only read the correct amount of memory
+        // from our pointer and we know where it will write it (op.error).
+        unsafe { self.encrypt_op(op) }
```
olivereanderson: Please mark this as unsafe to call: It suffers from the same problem as explained in the case of the vCPU ioctl.

It should not be possible to invoke UB from safe Rust even if the caller provides wrong/bad parameters.

You may also consider adding safe methods for individual TDX commands where it makes sense.

olivereanderson:
It would be really nice with some safe high level methods to issue TDX commands as well like:

```
impl VcpuFd {
  pub fn tdx_get_cpuid(&self, cpuid: &mut cpuid2, hw_error: &mut u64) -> Result<()> {

    let cmd = kvm_tdx_cmd {id: 5, flags: 0, data: std::ptr::from_mut(cpuid) as u64, hw_error: 0}; 

    // SAFETY: Supported command with valid parameters
    let ret = unsafe {vcpu.encrypt_op_tdx(&mut cmd)};
    *hw_error = cmd.hw_error;
    ret
  }
}
```

kvm-ioctls/src/ioctls/vcpu.rs
```diff
@@ -94,6 +94,52 @@ bitflags::bitflags! {
...
+    /// The guest has requested to set up a notification interrupt for the given
+    /// vector. See documentation for `TDVMCALL_SETUP_EVENT_NOTIFY_INTERRUPT`.
+    SetupEventNotify {
+        /// Return value for the request
+        ret: &'a mut u64,
+        /// Vector for which the notification should be set up
+        vector: u64,
+    },
+    /// Unknown TDVMCALL.
+    Unknown,
```
olivereanderson, line 140: From the  [KVM docs](https://docs.kernel.org/6.16/virt/kvm/api.html#the-kvm-run-structure) it seems that the Unknown union member has a `ret` and a `data` field as well.

Perhaps it would make sense to expose those too for logging and/or possible future compatibility purposes?

kvm-ioctls/src/ioctls/vcpu.rs
```diff
@@ -184,6 +230,13 @@ pub enum VcpuExit<'a> {
         /// size
         size: u64,
     },
+    /// Corresponds to KVM_EXIT_TDX
+    Tdx {
+        /// Exit flags
+        flags: u64,
+        /// Enum describing the exit and its parameters
+        exit: TdxExit<'a>,
+    },
```
olivereanderson, line 239: I am wondering if we perhaps should also add `nr` here. It could potentially be useful for debugging purposes.

olivereanderson:
> It would be really nice with some safe high level methods to issue TDX commands as well like:
> 
> ```
> impl VcpuFd {
>   pub fn tdx_get_cpuid(&self, cpuid: &mut cpuid2, hw_error: &mut u64) -> Result<()> {
> 
>     let cmd = kvm_tdx_cmd {id: 5, flags: 0, data: std::ptr::from_mut(cpuid) as u64, hw_error: 0}; 
> 
>     // SAFETY: Supported command with valid parameters
>     let ret = unsafe {vcpu.encrypt_op_tdx(&mut cmd)};
>     *hw_error = cmd.hw_error;
>     ret
>   }
> }
> ```

but that can be done in a follow up PR, at least from my point of view.

stefano-garzarella:
@00xc are you still interested in getting this PR merged? If yes, could you please re-open it in https://github.com/rust-vmm/rust-vmm/. We moved `kvm` crates there.

PR #376, opened 2026-06-03
CI: Fix lint warnings & arm64 test failures

00xc:
### Summary of the PR

Fix two issues with CI:
* Ever since commit 0aa1b76fe142 ("KVM: arm64: Prevent access to vCPU events before init") in the Linux kernel, an arm64 vCPU needs to be initialized before events can be accessed. `vcpu_events_test()`, as well as the doctests for `VcpuFd::{g,s}et_vcpu_events()` currently fail because they do not have this initialization, so just add it.
* Disable the `confusable_idents` lint for kvm-bindings, since most code is generated by bindgen and derive macros.

### Requirements

requirements:

  test.

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,0 +1,3 @@
+[advisories]
+  # bincode is only used a dev dependency
+  ignore = [ "RUSTSEC-2025-0141" ]
--- a/kvm-bindings/src/lib.rs
+++ b/kvm-bindings/src/lib.rs
@@ -3,6 +3,7 @@
 
 //! Rust FFI bindings to KVM, generated using [bindgen](https://crates.io/crates/bindgen).
 
+#![allow(confusable_idents)]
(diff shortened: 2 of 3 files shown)


00xc:
~The x86 CI failures should be fixed with #377~

rbradford:
Thanks! Would you also mind updating the rust-vmm-ci submodule and including that - I think that will address the cargo-audit failure

00xc:
The CI failure now is due to bincode being no longer maintained. It's "only" a test dependency for kvm-bindings. We could switch to something else like [wincode](https://github.com/anza-xyz/wincode) if needed.

rbradford:
> The CI failure now is due to bincode being no longer maintained. It's "only" a test dependency for kvm-bindings. We could switch to something else like [wincode](https://github.com/anza-xyz/wincode) if needed.

For now I just marked it as ignored.

rbradford (approved):

PR #382, opened 2026-08-15
Expose VmFd::enable_cap on all architectures

rogersnm:
## Summary of the PR

The `KVM_ENABLE_CAP` vm ioctl is architecture independent, but both the ioctl definition and `VmFd::enable_cap` were compiled only for x86_64, s390x and powerpc. This PR removes the cfg gates so they are available on every architecture. The doc example is unchanged as rendered; two hidden doctest lines gate the x86 specific call so the example builds and runs cleanly on every architecture.

Motivation: aarch64 now needs the vm level `enable_cap`. Linux 6.15 gates writes to the implementation ID registers (`MIDR_EL1`, `REVIDR_EL1`, `AIDR_EL1`) behind `KVM_CAP_ARM_WRITABLE_IMP_ID_REGS`, which a VMM must enable on the VM before creating any vCPU. Rewriting these registers is how a VMM masks the host CPU identity for guests (the reason the capability exists in KVM), and without `enable_cap` a VMM has to issue the raw ioctl itself, as Firecracker currently does in firecracker-microvm/firecracker#6116.

Verified: builds and passes clippy and rustfmt on aarch64. The doc example compiles for aarch64 and skips the x86 specific call at runtime; on hosts without `/dev/kvm` it fails the same way as every other doctest in the crate, so CI coverage is unchanged. The underlying ioctl path (enable the capability, then successfully write `MIDR_EL1` via `KVM_SET_ONE_REG`) was exercised end to end on an EC2 c7g.metal running a 7.0 kernel as part of the Firecracker work linked above.

## Requirements

Before submitting your PR, please make sure you addressed the following requirements:

- [x] All commits in this PR have Signed-Off-By trailers (with `git commit -s`), and the commit message has max 60 characters for the summary and max 75 characters for each description line.
- [x] All added/changed functionality has a corresponding unit/integration test. (Covered by the existing doc example, which now compiles and runs on every architecture.)
- [x] All added/changed public-facing functionality has entries in the "Upcoming Release" section of CHANGELOG.md (if no such section exists, please create one).
- [x] Any newly added `unsafe` code is properly documented. (No new unsafe.)

### DIFF
--- a/kvm-ioctls/src/ioctls/vm.rs
+++ b/kvm-ioctls/src/ioctls/vm.rs
@@ -1521,9 +1521,9 @@ impl VmFd {
     /// // Because an IOAPIC supports 24 pins, that's the reason why this test
     /// // picked this number as reference.
     /// cap.args[0] = 24;
+    /// #[cfg(target_arch = "x86_64")]
     /// vm.enable_cap(&cap).unwrap();
     /// ```
-    #[cfg(any(target_arch = "x86_64", target_arch = "s390x", target_arch = "powerpc"))]
     pub fn enable_cap(&self, cap: &kvm_enable_cap) -> Result<()> {
         // SAFETY: The ioctl is safe because we allocated the struct and we know the
         // kernel will write exactly the size of the struct.
--- a/kvm-ioctls/CHANGELOG.md
+++ b/kvm-ioctls/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Upcoming Release
 
+### Changed
+
+- [[#382]](https://github.com/rust-vmm/kvm/pull/382) `VmFd::enable_cap` and the
+  `KVM_ENABLE_CAP` ioctl definition are now available on all architectures
+  instead of only x86_64, s390x and powerpc. aarch64 needs them to enable
+  `KVM_CAP_ARM_WRITABLE_IMP_ID_REGS` (Linux 6.15), without which writes to the
+  implementation ID registers (for example `MIDR_EL1` through a VMM CPU
+  template) fail with EINVAL.
+
 ## v0.25.0
 
 ### Added
--- a/kvm-ioctls/src/kvm_ioctls.rs
+++ b/kvm-ioctls/src/kvm_ioctls.rs
@@ -226,7 +226,6 @@ ioctl_io_nr!(KVM_SET_TSC_KHZ, KVMIO, 0xa2);
 ioctl_io_nr!(KVM_GET_TSC_KHZ, KVMIO, 0xa3);
 
 /* Available with KVM_CAP_ENABLE_CAP */
-#[cfg(not(any(target_arch = "aarch64", target_arch = "riscv64")))]
 ioctl_iow_nr!(KVM_ENABLE_CAP, KVMIO, 0xa3, kvm_enable_cap);
 /* Available with KVM_CAP_SIGNAL_MSI */
 #[cfg(any(


kvm-ioctls/src/ioctls/vm.rs
```diff
@@ -1521,9 +1521,11 @@ impl VmFd {
     /// // Because an IOAPIC supports 24 pins, that's the reason why this test
     /// // picked this number as reference.
     /// cap.args[0] = 24;
+    /// # #[cfg(target_arch = "x86_64")]
     /// vm.enable_cap(&cap).unwrap();
+    /// # #[cfg(not(target_arch = "x86_64"))]
+    /// # let _ = (&vm, &cap);
```
00xc: Why is this needed?

I initially thought you may have added this to avoid a unused variable warning, but doctests have a few extra allow attributes (see https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#pre-processing-examples), so removing this does not produce any warnings. I tested it out with:

```
cargo clippy --tests --target aarch64-unknown-linux-gnu
```

rogersnm: Thanks for reviewing the PR and catching this, @00xc. You’re right: I added the dummy use to avoid unused-variable warnings on non-x86 targets, but I’d overlooked that rustdoc automatically allows unused variables in doctests. I verified the removal on aarch64 by compiling the relevant doctest with `RUSTDOCFLAGS="-D warnings" cargo test --doc -p kvm-ioctls enable_cap`, which produced no unused-variable warnings.

I’ve kept the `x86_64` guard around the actual call because the example uses the x86-specific `KVM_CAP_SPLIT_IRQCHIP`, which is unsupported on aarch64 and causes `KVM_ENABLE_CAP` to return `EINVAL`. I’ve removed only the redundant non-x86 dummy use.

RuoqingHe (commented):
Codewise looking good, while those changes are already in #344

kvm-ioctls/src/ioctls/vm.rs
andreeaflorescu, line 1511: This is the part that is architecture specific I guess.

Is there a capability that we can enable on all platforms? Alternatively, can we define a per platform capability?

kvm-ioctls/src/ioctls/vm.rs
```diff
@@ -1521,9 +1521,9 @@ impl VmFd {
     /// // Because an IOAPIC supports 24 pins, that's the reason why this test
     /// // picked this number as reference.
     /// cap.args[0] = 24;
+    /// # #[cfg(target_arch = "x86_64")]
```
ShadowCurse: The first `#` is not needed

ShadowCurse (dismissed):
Overall lgtm

kvm-ioctls/src/ioctls/vm.rs
rbradford: Indeed, please fix.

rbradford (changes requested):

rbradford (approved):
Thanks!

kvm-ioctls/src/ioctls/vm.rs
rogersnm: Thanks both, @ShadowCurse and @rbradford. Fixed now by removing the unnecessary leading `#`, and I’ve rebased the PR onto the latest `main`.

ShadowCurse (approved):

kvm-ioctls/src/ioctls/vm.rs
rogersnm: Thanks for raising this, @andreeaflorescu. I looked into it further and `KVM_CAP_HALT_POLL` is documented as supporting all architectures and can be enabled through `VmFd`, making it suitable for this example. I also verified that `KVM_CHECK_EXTENSION` reports it as supported and that `KVM_ENABLE_CAP` succeeds on both aarch64 and x86_64.

To remain compatible with older kernels, the example could check for support first:

```rust
use kvm_bindings::{kvm_enable_cap, KVM_CAP_HALT_POLL};
use kvm_ioctls::Kvm;

if kvm.check_extension_raw(KVM_CAP_HALT_POLL.into()) > 0 {
    let cap = kvm_enable_cap {
        cap: KVM_CAP_HALT_POLL,
        args: [100_000, 0, 0, 0],
        ..Default::default()
    };

    vm.enable_cap(&cap).unwrap();
}
```

Since this PR has now merged, I’m happy to submit this as a small follow-up if you think it would improve the example.

andreeaflorescu: Sounds good. Thanks!

PR #385, opened 2026-09-10
fix: preserve header-only nested state

windsornguyen:
### Summary of the PR

`KVM_GET_NESTED_STATE` can return only the 128-byte header when VMXON is active but no VMCS12 payload is loaded. That header still carries `vmxon_pa`. SVM can likewise carry GIF state in a header-only response.

`VcpuFd::nested_state()` currently maps every header-sized response to `None`, which lets callers discard valid nested state. This change returns the non-zero state length after every successful ioctl and leaves the populated buffer authoritative, matching the [KVM API contract](https://docs.kernel.org/virt/kvm/api.html#kvm-get-nested-state).

### API impact

The return type changes from `Result<Option<NonZeroUsize>>` to `Result<NonZeroUsize>`. Callers should serialize or inspect the populated buffer after success instead of treating a header-sized result as absent.

### Verification

- Red on Intel KVM: the regression test reported `left: None`, `right: Some(128)` after setting header-only VMXON state.
- Green on Intel KVM: `test_get_and_set_header_only_nested_state` preserves the 128-byte response and `vmxon_pa`.
- `cargo test --all-features --workspace`: 8 kvm-bindings tests, 68 passing kvm-ioctls tests with 2 ignored, and 102 doc tests.
- `cargo clippy --all-features --workspace --all-targets -- -D warnings -D clippy::undocumented_unsafe_blocks`.
- `cargo fmt --all -- --check --config format_code_in_doc_comments=true`.
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --all-features --no-deps`.
- Official rust-vmm coverage harness with `--profile=devel`; x86_64 coverage remains 88.24%.
- [Buildkite #1619](https://buildkite.com/rust-vmm/kvm-ci/builds/1619) passed the complete upstream matrix.

### Requirements

- [x] All commits have a matching `Signed-off-by` trailer.
- [x] The changed behavior has an x86 KVM regression test.
- [x] The public API change is recorded under `Upcoming Release`.
- [x] The added union read documents its safety invariant.


stefano-garzarella:
@windsornguyen are you still interested in getting this PR merged? If yes, could you please re-open it in https://github.com/rust-vmm/rust-vmm/. We moved `kvm` crates there.

PR #387, opened 2026-09-18
kvm-ioctls: Expose size information for IoIn/IoOut

### LINKED ISSUE #371: kvm-ioctls: Need size for IoIn/IoOut
`VcpuExit::IoIn(port, data)` and `VcpuExit::IoOut(port, data)` flatten `kvm_run.io.size * kvm_run.io.count` into a single `&[u8]` of length `size * count`. Once the variant is constructed, the per-iteration  `io.size` is no longer recoverable from the variant — only the product is. This makes it impossible for downstream VMMs to dispatch repeated  PIO accesses (`rep ins/outs`, string IO) the way QEMU does, where each  iteration is a single bus transaction at the same port with `io.size` bytes. 

We need to break the API and change `VcpuExit:IoIn/Out` to embed the size

scholzp:
### Summary of the PR

In the current state, information about the size is hidden. This makes it impossible for consuming code to make decisions based on it, e.g. if a certain access width is allowed or not. For example, I was working on `fw_cfg` for Cloud Hypervisor for whose data register only one byte port reads are allowed, which I currently cannot ensure given this information is hidden.

This fixes https://github.com/rust-vmm/kvm/issues/371. 

### Requirements

requirements:

  test.


stefano-garzarella:
@scholzp are you still interested in getting this PR merged? If yes, could you please re-open it in https://github.com/rust-vmm/rust-vmm/. We moved `kvm` crates there.

